PR #33, opened 2022-11-19
Add an API Token Auth Mechanism

JSv4:
Prior to this PR, the only way to authenticate against OpenContracts is to use the JWT token, the password auth (if you've turned off auth0 authentication) or Django's session authentication. None of these are well-suited to server-side connections. It would be nice to be able to connect other backend tools and infrastructure to the platform. 

This PR lets you use Authentication Tokens linked to a particular user's account. It uses DRF's token authentication app (so DRF is a dependency, but we're using this already, so no new dependencies). It provides a Graphene Middleware and an Authentication Backend, the first base on the Graphql JWT Middleware and the latter based on DRF's token auth backend.

### DIFF
--- a/config/graphql/base.py
+++ b/config/graphql/base.py
@@ -135,7 +135,7 @@ class DRFMutation(graphene.Mutation):
         try:
             logger.info("Test if context has user")
             if info.context.user:
-                logger.info("User id: ", info.context.user.id)
+                logger.info(f"User id: {info.context.user.id}")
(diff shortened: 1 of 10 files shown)


JSv4:
Coverage went down for a stupid reason in `permissioning_utils.py`... the comments were "using" certain vars so that counted as coverage. Lame. Those aren't proper tests and I'm not writing them now.

PR #92, opened 2023-03-10
Add Funsd Export

JSv4:
Major change here is added logic to convert a target annotation to FUNSD format which can then be used to easily train LayoutLMv2, v3, etc. Also fixed an issue with annotation loading and download url generation for exports. 

Coverage is going to take a small hit. Will add export tests in the future.


JSv4:
Coverage hit is way too high. Need to write actual tests.

JSv4:
Actually, it's probably better to have this in here. Open a task to write test.

PR #116, opened 2024-05-26
Add Structural Annotations and Vector Embeddings

JSv4:
Added a couple major features in preparation for more advanced analytics:

1. A vector embeddings column using pgvector to store vector embeddings for annotations and documents.
2. A vector embeddings microservice to generate vector embeddings
3. A modified nlm-ingestor microservice to create PAWLs token data layers and layout 
    1. The old grobid-based one I forked from AllenAI has some funky output sometimes anyway, so planning to totally deprecate that.


JSv4:
Will improve coverage in another PR.

PR #156, opened 2024-07-15
Install Embeddings Model @ /models in Production Container + Fix Extract Where Search Text is None

### LINKED ISSUE #155: [BUG] - Fix Production Embedder Deploy
- Forgot to update prod Dockerfile to properly setup embeddings model in /models
- Fixed a bug in extract task where `search_text` is None causes an error.,

JSv4:
See description. 

Fixes #155

### DIFF
--- a/opencontractserver/tasks/data_extract_tasks.py
+++ b/opencontractserver/tasks/data_extract_tasks.py
@@ -64,7 +64,7 @@ def oc_llama_index_doc_query(cell_id, similarity_top_k=15, max_token_length: int
         query = datacell.column.query
 
         # Special character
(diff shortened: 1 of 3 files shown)


opencontractserver/tasks/data_extract_tasks.py
```diff
@@ -64,7 +64,7 @@ def oc_llama_index_doc_query(cell_id, similarity_top_k=15, max_token_length: int
         query = datacell.column.query
 
         # Special character
-        if "|||" in search_text:
```
JSv4, line 67: This failed where `search_text` was None, which it could be.

PR #172, opened 2024-07-29
Upgrade Django App Dependencies to work with Django LTS

JSv4:
Django-stubs and django extensions both have management commands that don't work with post Django 4.1 syntax. This was not caught by the tests so v2.0 branch fails on startup. I've upgraded the dependency versions to work with Django 4.2

**There is another issue here... for some reason, upgrading to Django 4.2 creates an issue with opencv and/or vcr.py. At the moment, I can fix one or the other but not both.** Extract thumbnail fails when the upgrades mentioned above are applied and opencv throws some kind of numpy version error. If you upgrade opencv to latest, it works fine, but then vcr.py throws a super cryptic error about a setuptools issue. If you do not install vcr.py (which is only needed for tests), then the application (plus thumbnail extraction) appears to work. It's a bit of a quandary, though, as dropping vcr.py really drops test coverage and is not really a "fix". I'm opening an issue in the vcr.py repo and the issue may be due to how they're using pyproject.toml and setuptools.  

Given the choices here, I'm going to merge this without a resolution for the opencv/vcr,py problem so the v2.post1 release works and tests continue to function, but thumbnails are broken. I don't love that, but I consider a minor cosmetic function to be the least important of possible issues here (1. boot failure, 2. huge drop in test coverage, 3. missing thumbnail).

### DIFF
--- a/compose/local/django/Dockerfile
+++ b/compose/local/django/Dockerfile
@@ -36,6 +36,8 @@ RUN apt-get update && apt-get install --no-install-recommends -y \
 # Requirements are installed here to ensure they will be cached.
 COPY ./requirements .
 
+RUN pip install --upgrade pip
+
 # Create Python Dependency and Sub-Dependency Wheels.
 RUN pip wheel --wheel-dir /usr/src/app/wheels  \
   -r ${BUILD_ENVIRONMENT}.txt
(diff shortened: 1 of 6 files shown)


JSv4:
Weirdly these changes caused an issue installing vcr.py. Issue opened [here](https://github.com/kevin1024/vcrpy/issues/854) by someone else minutes before I did.

JSv4:
Setuptools v 72 broke, like, [all the things](https://github.com/pypa/setuptools/issues/4519). This was causing an issue with vcr.py not being able to install as their test suite relied on deprecated (now removed) code. Currently using a specific commit on their main as they've not released an official fix for this on pypi yet. I believe the reason *this* popped up was upgrading to latest opencv (for numpy compatibility) was forcing setuptools upgrade.

PR #293, opened 2025-01-07
Issue85 port redirect to 3000

ThreatLentes:
- Made changes to config/urls.py that allows the routing to port 3000


JSv4 (approved):
Thanks for the contribution! LGTM.

PR #311, opened 2025-01-24
Implement ASGI, Doc Chat and Doc Summary View

JSv4:
- Added ASGI application and django channels to start adding in Delphic-style streaming again. 
- Added new models for conversation and message. 
- Added document summary view


JSv4:
Merge approve. Lingering issue with async connections dropping in our extract task tests means coverage score dropped but it seems to be related to GitHub action env as tests pass locally. Huge amount of time wasted trying to resolve. Revisit when time permits.

PR #410, opened 2025-07-12
Move CI to k8s / Kind

JSv4:
Probably should split the repo... but this is not really a step back in speed / complexity from the compose-based CI atm, and it will support much more powerful deployment options.


JSv4:
Abandoning for now. Will get much more value restructuring container builds for performance and pushing release containers to ghcr.io

PR #808, opened 2026-01-19
Audit semantic-ui usage across frontend

JSv4:


JSv4:
This was originally just an analysis and then later a much too large PR that would be a massive undertaking. Doing this bit by bit now following style guide merged in #922

Issue #841, opened 2026-02-01
Agent Tools Async Issue


JSv4:
<pre><b>Summary</b>                                                                       
                                                                                
  The <font color="#B1B9F9">update_document_description</font> tool fails when called by the PydanticAI agent
   during corpus actions, preventing document descriptions from being saved.    
                                                                                
  <b>Error</b>                                                                         
                                                                                
  Error in tool update_document_description: You cannot call this from an async 
  context - use a thread or sync_to_async.                                      
                                                                                
  <b>Observed Behavior</b>                                                             
                                                                                
  - The &quot;Create Description&quot; agent corpus action runs successfully              
  - The agent generates a description for the document                          
  - When the agent calls <font color="#B1B9F9">update_document_description</font> to save the result, the    
  tool fails                                                                    
  - The task is still marked as <font color="#B1B9F9">succeeded</font> with <font color="#B1B9F9">response_length=82</font>, masking the  
  failure                                                                       
  - <b>Descriptions are not persisted to the database</b>                              
                                                                                
  <b>Affected Workflow</b>                                                             
                                                                                
  1. Documents are added to a corpus with an active &quot;Create Description&quot; corpus 
  action                                                                        
  2. <font color="#B1B9F9">process_corpus_action</font> triggers <font color="#B1B9F9">run_agent_corpus_action</font> for each document   
  3. PydanticAI agent reviews the document and generates a description          
  4. Agent calls <font color="#B1B9F9">update_document_description</font> tool → <b>fails silently</b>              
                                                                                
  <b>Example Log Trace</b>                                                             
                                                                                
  [2026-02-01 07:09:41,979: INFO/ForkPoolWorker-2] [AgentCorpusAction] Executing
   &apos;Create Description&apos; on document &apos;Chapter 1, Subchapter X - Sale Of Assets,  
  Dissolution And Winding Up&apos; (id=385)                                          
  [2026-02-01 07:09:42,374: INFO/ForkPoolWorker-1] Caller tool                  
  &apos;update_document_description&apos; overrides default - using caller&apos;s configuration
  [2026-02-01 07:09:42,374: INFO/ForkPoolWorker-1] Created pydantic ai agent    
  with context You are a legal librarian...                                     
  [2026-02-01 07:09:43,440: INFO/ForkPoolWorker-2] [PydanticAI sync chat]       
  Starting chat with message: &apos;You must review this document and generate a     
  description...&apos;                                                               
[...]

PR #842, opened 2026-02-01
Fix sync-to-async boundary for Django ORM in PydanticAI tools

JSv4:
## Summary
Fixes Issue #841 where sync functions performing Django ORM operations would fail with "You cannot call this from an async context" when invoked by PydanticAI agents. The fix properly wraps sync tool functions with `database_sync_to_async` to ensure thread isolation for database operations.

## Changes

- **Added sync-to-async database wrapper**: Introduced `_db_sync_to_async` helper that wraps sync functions with Django's `database_sync_to_async` (or `asgiref.sync_to_async` as fallback) with `thread_sensitive=False` to prevent stale database connections between tool invocations

- **Updated PydanticAIToolWrapper**: Modified the `sync_to_async_wrapper` method to use `_db_sync_to_async` when calling sync tool functions, ensuring Django ORM operations run in a proper thread pool context rather than directly in the async event loop

- **Added comprehensive tests**: Created `TestSyncToAsyncDatabaseWrapping` test class with three test cases covering:
  - Sync tools with database read operations
  - Sync tools with database write operations  
  - Sync tools wrapped through both CoreTool and PydanticAIToolWrapper

## Implementation Details

The fix uses `functools.partial` to create a reusable wrapper that:
1. Attempts to import `database_sync_to_async` from `channels.db` (preferred for Django Channels projects)
2. Falls back to `asgiref.sync_to_async` if Channels is not installed
3. Sets `thread_sensitive=False` to ensure each tool invocation gets a fresh worker thread, preventing connection pool issues

This allows tools like `update_document_description` to work correctly when invoked by PydanticAI agents during corpus actions, while maintaining backward compatibility with projects that may not have Channels installed.

https://claude.ai/code/session_018hsEGD6TF1KuUoaMpTvUKh

### DIFF
--- a/opencontractserver/llms/tools/pydantic_ai_tools.py
+++ b/opencontractserver/llms/tools/pydantic_ai_tools.py
@@ -3,6 +3,7 @@
 import inspect
 import logging
 from collections.abc import Awaitable
+from functools import partial
 from typing import Any, Callable, Optional, get_type_hints
 
 from pydantic import BaseModel, ConfigDict, Field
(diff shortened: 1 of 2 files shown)


JSv4:
@claude create follow-up issue noted here in earlier comment

PR #979, opened 2026-02-27
Fix code review issues from text block deep linking (#958)

JSv4:
## Summary

Addresses code review feedback on the (now merged) text block deep linking feature from PR #958.

- **Fix double-action on cross-document source click** (CorpusChat.tsx): Cross-document sources now navigate immediately and skip the local `setChatSourceState`/`onMessageSelect` calls, preventing a flash of stale local selection state before navigation replaces the view
- **Deduplicate and hoist text block clearing logic**: Extracted `useClearTextBlockOnInteraction` shared hook used by both PDF and text annotators. Moved the clearing effect out of per-page `PDFPage` (which fired O(visible-pages) redundant navigate calls) into the single-instance parent `PDF` component
- **Remove dead code in `buildQueryParams`**: Removed the no-op `else if (params.textBlock === null) { searchParams.delete("tb") }` branch — since `buildQueryParams` creates a fresh `URLSearchParams`, passing `null` simply omits the key (no stale param to delete). Added clarifying comment.
- **Add `LocationLike`/`NavigateFn` type aliases**: Applied to all 10 `update*Param` functions in `navigationUtils.ts` for consistent, explicit typing

### PR description corrections (from review items #6 and #7)
- The original PR description mentioned `corpus_id` on `ChatMessageSource`, but only `document_id` was added
- The original PR description listed `clearTextBlockParam()` as a helper, but only `updateTextBlockParam(location, navigate, null)` exists

## Test plan
- [x] TypeScript compiles cleanly (`tsc --noEmit`)
- [x] All 736 unit tests pass (`vitest run`)
- [x] All 27 text block encoding tests pass
- [x] Prettier formatting verified
- [ ] Verify cross-document source click navigates without flash
- [ ] Verify ?tb= param survives Relay-ID canonical redirect
- [ ] Verify selecting an annotation clears ?tb= highlight


JSv4:
### Responses to review items

**1. `onSourceNavigate` catching same-document sources**

By design. The prop JSDoc (lines 101–110 of `CorpusChat.tsx`) explicitly states: *"Only pass this prop in contexts where no document is currently displayed (e.g. corpus-level chat), so that every source is effectively a cross-document navigation."*

In practice, `onSourceNavigate` is only provided in two places — both corpus-level views in `Corpuses.tsx` (line 2278 for `CorpusHome`, line 2504 for the Chat History tab). The document-level `ChatTray` in the annotator does NOT pass this prop, so sources select locally there. The caller-responsibility contract is correct and documented.

**2. Missing `buildQueryParams({ textBlock: null })` test**

Already exists. `navigationUtils.test.ts` lines 90–103 cover all three cases:
- `{ textBlock: "s100-200" }` → `?tb=s100-200`
- `{ textBlock: null }` → `""`
- `{ textBlock: undefined }` → `""`

**3. No unit test for `useClearTextBlockOnInteraction`**

Acknowledged as nice-to-have. The hook is small, extracted from existing working code, and its constraint is documented in the JSDoc.

**4. `location` in dependency array**

Noted. Not a regression — same pattern as pre-existing code. The `highlightedTextBlock()` guard prevents redundant navigate calls.

---

Also merged latest main (resolved CHANGELOG.md and queries.py conflicts — took main's modularized queries.py and reapplied the corpus-membership fix to `slug_queries.py`).

JSv4:
Addressed review items selectively:

**Implemented:**
- **Superuser guardian permission test** — Added `GuardianModelSuperuserPermissionsTestCase` covering the full 7-permission set and individual permission type checks via `user_has_permission_for_obj`. Both tests pass.
- **TxtAnnotator read-only comment** — Added clarifying comment in `handleMouseUp` documenting the intentional behavior: menu shown in read-only mode for Copy Text/Copy Link, with Apply Label gated on `allowInput && !read_only`.
- **PR description correction** — Updated the `buildQueryParams` bullet to accurately describe it as dead code removal (the `else if` branch was a no-op since `URLSearchParams` is always fresh).

**Evaluated and skipped:**
- **Double-encoding risk on `annotation_id`** — False positive. Traced the full pipeline: `SourceNode.annotation_id` is typed `int` (Django PK), serialized as a plain integer via `json.dumps()`, received as `number` in the `WebSocketSources` TypeScript interface, and passed through without transformation. `toGlobalId("AnnotationType", rawInt)` is the correct and only encoding.
- **`.distinct()` on slug query** — Not needed. `Document` has a unique constraint on `(creator, slug)`, and `DocumentPath` has a unique constraint on `(corpus, path)` when `is_current=True, is_deleted=False`. The data model prevents duplicate results.
- **Hardcoded colour tokens** — Acknowledged as not blocking; out of scope for this PR.

PR #1135, opened 2026-03-22
Add bbox_annotations for bounding-box to token resolution on import

JSv4:
## Summary

- Adds a new optional `bbox_annotations` field to document import data that lets external tools provide PDF bounding-box coordinates instead of PAWLs token indices
- A standalone pure function resolves bounding boxes to standard `TOKEN_LABEL` annotations by matching token center points against bounding rectangles
- Integrated into all four annotation-bearing import pathways: annotated document import, corpus export/import, bulk ZIP sidecars, and worker uploads
- 19 unit tests cover core resolution, multi-page annotations, edge cases, image tokens, overlapping annotations, malformed input, and the merge helper

## Motivation

External tools (pdfplumber, PyMuPDF, etc.) can identify regions in PDFs using bounding boxes, but can't produce `TOKEN_LABEL` annotations because PAWLs token indices don't exist until after parsing. This feature bridges that gap — tools provide coordinates, OpenContracts resolves them to proper token-anchored annotations at import time.

## Key Design Decisions

- **No new annotation types or label types** — `bbox_annotations` is resolved to standard `TOKEN_LABEL` before entering the database
- **Pure function** — resolution logic has no side effects, no DB access, trivially testable
- **Import-only** — `bbox_annotations` is never persisted or exported; resolved annotations are indistinguishable from any other TOKEN_LABEL
- **Coordinates in PDF points** — same coordinate system as PAWLs tokens, no scaling needed

## Test plan

- [x] 19 unit tests for resolution function (multi-page, edge cases, image tokens, overlapping, malformed keys, merge helper)
- [x] 184 import-related regression tests pass (annotated doc, corpus v2, sidecar, worker uploads)
- [ ] Manual test: import a document with `bbox_annotations` via annotated document import mutation
- [ ] Manual test: bulk ZIP with sidecar containing `bbox_annotations` + `skip_pipeline: true`


JSv4:
Over complicated. Not needed

PR #1195, opened 2026-04-04
Track document moves via DocumentPath history nodes

JSv4:
DocumentFolderService.move_document_to_folder() and
move_documents_to_folder() previously did in-place updates on
DocumentPath.folder, silently losing move history. Now they create new
DocumentPath records linked to the previous state via parent, making
moves traversable through get_path_history().

Changes:
- folder_service: move_document_to_folder() creates a new DocumentPath
  node with parent=current instead of in-place .update(folder=folder)
- folder_service: move_documents_to_folder() does the same per-document
- folder_service: Add _compute_moved_path() helper to derive path string
  from target folder, with conflict resolution for occupied paths
- folder_service: Skip no-op moves (document already in target folder)
- versioning: get_path_history() now detects folder-only moves and
  includes folder_id in history entries
- tests: Updated move tests to verify new history nodes are created,
  parent chain is linked, version number is preserved, and same-folder
  moves are no-ops


JSv4:
this must be addressed:

2) Concern: delete_folder partial-success state - when some documents relocate and others fail, already-relocated docs stay at root while stuck ones stay in the folder. The API prevents deletion but leaves a mixed state. The caller can safely retry (already-relocated docs won't be found by the folder filter), but this retry-safety should be documented in the docstring. Also verify the test suite covers retry-after-partial-failure.

PR #1237, opened 2026-04-12
Fix O(N) query regression in bulk folder operations (#1199)

### LINKED ISSUE #1199: DocumentPath: O(N) queries in bulk move/delete operations
## Context

Flagged during review of PR #1195 (DocumentPath history tracking).

## Problem

`move_documents_to_folder()` and `delete_folder()` in `opencontractserver/corpuses/folder_service.py` previously used a single `.update()` call (O(1)). They now issue **3 DB queries per document**: one `EXISTS` conflict check, one `UPDATE` (deactivate old path via `save(update_fields=...)`), and one `INSERT` (`DocumentPath.objects.create(...)`).

For a 100-document bulk move, that's ~300 DB round-trips vs the original 1. This is a significant performance regression for large batch operations.

## Suggested Fix

1. Batch-deactivate old paths with `DocumentPath.objects.filter(pk__in=ids_to_retire).update(is_current=False)`
2. Collect all new path objects, then use `bulk_create`
3. Pre-load all `(corpus, path)` active paths into a set for conflict checking instead of per-document `EXISTS` queries

Note: disambiguation still needs per-path calls so it can't be fully batched, but the save/create round-trips can be.

## Files

- `opencontractserver/corpuses/folder_service.py` — `move_documents_to_folder()`, `delete_folder()`

JSv4:
## Summary

Optimize `move_documents_to_folder()` and `delete_folder()` in `DocumentFolderService` to reduce database queries from ~3 per document to 2 total queries per batch operation. This fixes a performance regression where bulk operations issued hundreds of queries instead of leveraging batch operations.

## Key Changes

- **Batch path disambiguation**: Replace per-document `_disambiguate_path()` queries with a single pre-fetch of occupied paths in the target directory, shared across all documents in the batch. Within-batch conflicts are resolved in-memory by mutating the shared set as each path is claimed.

- **Batch database writes**: Replace individual `save(update_fields=["is_current"])` and `DocumentPath.objects.create()` calls with two bulk operations:
  - Single `DocumentPath.objects.filter().update(is_current=False)` for all superseded paths
  - Single `DocumentPath.objects.bulk_create()` for all new successor paths

- **Signal dispatch for bulk_create**: Since `bulk_create()` bypasses per-row `post_save` signals, manually dispatch them via new `_dispatch_document_path_created_signals()` to preserve the document-text embedding side effect.

- **Query optimization helpers**:
  - `_fetch_occupied_paths_in_directory()`: Single-query fetch of active paths in a directory (with regex filtering to avoid pulling entire subtrees)
  - `_target_directory_string()`: Normalize folder paths to the format expected by the fetch helper
  - Updated `_disambiguate_path()` to accept `occupied_override` parameter for pre-fetched sets

- **Row-level locking refinement**: Add `of=("self",)` to `select_for_update()` calls to scope locks to `DocumentPath` rows only, avoiding accidental locks on `Document` rows for the transaction duration.

## Implementation Details

- The batch path planning phase pre-computes all target paths and detects within-batch conflicts before any database writes occur
- Occupied paths are tracked in a mutable set that grows as each document claims its disambiguated path, ensuring siblings with identical filenames receive unique suffixes
- All changes occur within existing `transaction.atomic()` blocks, preserving rollback semantics
- Test updates reflect the shift from sequential to batch processing (e.g., `extra_occupied` → `occupied_override` parameter)

https://claude.ai/code/session_01Wp8voAR22aHiENmipX9jjL


JSv4:
## Review feedback addressed

Merged `main` into the PR branch and assessed the four review items:

### 1. Slashless paths raising from wrong location (FIXED)

**Valid concern, addressed.** Added an early `startswith("/")` guard at the top of `_disambiguate_path` so callers that pass slashless paths get a clear, localized error message rather than a confusing `ValueError` from `_fetch_occupied_paths_in_directory`. Also removed the now-unreachable `directory = ""` dead code branch.

In practice, all current callers pass paths from `_compute_moved_path`, which always returns paths starting with `/`, so this is defense-in-depth for future callers.

### 2. Two redundant static methods (FIXED)

**Valid concern, addressed.** Consolidated `_target_directory_string` to delegate to `_target_directory_string_from_path` instead of duplicating the path-normalization logic. The `_from_path` variant is the canonical implementation; the folder-accepting variant is now a thin wrapper that calls `get_path()` then delegates.

### 3. `move_single_document_to_folder` not updated to `of=("self",)` (ALREADY ADDRESSED)

**Not a real issue.** The single-document method `move_document_to_folder` at line 998 already uses `select_for_update(of=("self",))`. This was present before the review.

### 4. Missing `update_fields` in `post_save.send()` (ALREADY ADDRESSED)

**Not a real issue.** The signal dispatch in `_dispatch_document_path_created_signals` already passes `update_fields=None`, which is correct for `created=True` signals — this matches Django's own `Model.save()` behavior for newly created instances.

PR #1322, opened 2026-04-20
CI: remove tippy debug step, harden run selection, note coverage change

### LINKED ISSUE #1319: PR 1318 Follow-up
### Minor issues

**1. Debug step left in `frontend.yml` (lines 38-45)**

The `Debug - Check for tippy references` step looks like a leftover debugging artifact. It runs on every PR and adds noise to CI logs. Should be removed before merging.

**2. `matching[0]` relies on undocumented API sort order**

In `codecov-notify.yml` line 69, the script takes `matching[0]` as the latest run, relying on the GitHub API returning most-recent-first. A more defensive approach would sort explicitly before taking element 0. Low-risk since the assumption generally holds, but worth hardening.

**3. `frontend-unit` flag will show a lower percentage after `all: true`**

This is correct behavior, not a regression — the denominator now includes all source files, not just files imported by tests. A one-liner in the CHANGELOG noting that `frontend-unit` coverage will appear to drop would prevent future contributors from treating it as a regression.

### No concerns on

- Security: no secrets beyond `CODECOV_TOKEN`, correctly scoped
- Artifact retention (7 days appropriate for cross-workflow temp data)
- `npx --yes lcov-result-merger@5`: pinned major version, `--yes` correct for CI
- `if: success() || failure()` on upload steps: ensures coverage uploads even on partial test failures
- `yarn.lock` additions: expected transitive deps for `lcov-result-merger`

**Overall**: Clean up the debug step (issue 1) before merging. Issue 2 is a low-risk hardening suggestion. The rest looks correct and well thought out.

JSv4:
## Summary
Addresses the three follow-up items flagged in issue #1319 (review of PR #1318):

- **Remove debug CI step** (`.github/workflows/frontend.yml`): Drop the `Debug - Check for tippy references` step that was a leftover debugging artifact running on every PR and adding noise to CI logs.
- **Harden `matching[0]` run selection** (`.github/workflows/codecov-notify.yml`): Sort matching workflow runs explicitly by `run_number` desc before taking element 0, instead of relying on the undocumented "most-recent-first" ordering of `listWorkflowRunsForRepo`. Re-runs bump `run_attempt` while keeping `run_number`, so the highest `run_number` reliably selects the latest attempt.
- **CHANGELOG note on `frontend-unit` percentage drop** (`CHANGELOG.md`): Added a one-liner explaining that `frontend-unit` will display a lower percentage after enabling `all: true` in the Vitest coverage block (denominator now includes every file under `frontend/src/`), so future contributors don't treat the drop as a regression.

## Test plan
- [x] `python -c "import yaml; yaml.safe_load(open('.github/workflows/frontend.yml')); yaml.safe_load(open('.github/workflows/codecov-notify.yml'))"` passes
- [ ] Frontend CI runs cleanly on this PR without the debug step
- [ ] Codecov Notify picks up the latest attempt per workflow after the explicit sort

Closes #1319


JSv4:
Closing — main (PR #1320) already landed 2 of the 3 fixes (tippy-debug removal + explicit codecov-notify sort, via created_at desc). Only the CHANGELOG note on the frontend-unit % drop was unique, not worth its own PR.

PR #1440, opened 2026-05-02
Optimize document list loading with lazy relationship queries

JSv4:
## Summary

This PR eliminates N+1 query performance issues when loading document lists on large corpora by deferring relationship metadata fetching until user interaction and optimizing backend relationship counting.

## Key Changes

**Backend (Django/GraphQL)**
- Added `DocumentRelationshipQueryOptimizer.get_relationship_counts_by_document()` method that computes relationship counts for all visible documents in two aggregated SQL queries instead of per-document `.count()` calls
- Implemented request-level caching on the GraphQL context to ensure counts are computed once per request, even when resolving the field for hundreds of documents
- Updated `resolve_doc_relationship_count` to use the new batch counting method, reducing O(N) queries to O(1) per document after the initial aggregation
- Simplified `in_folder` filter in `config/graphql/filters.py` to use lazy QuerySet subqueries instead of materializing document IDs into Python sets

**Frontend (React/GraphQL)**
- Removed eager fetching of `allDocRelationships` from `GET_DOCUMENTS` query — this was pulling source/target/label data for every linked document on the page (5,000+ extra fields on a 100-document corpus)
- Created new `GET_DOC_RELATIONSHIPS_FOR_DOC` query that lazily fetches relationships for a single document through the existing `bulkDocRelationships` field
- Updated `ModernDocumentItem` to use `useLazyQuery` for relationships, triggering the fetch on first hover/focus of the relationship badge instead of on initial page load
- Added `handleRelationshipHover` callback to defer relationship data loading until user interaction

## Implementation Details

- The backend optimizer uses `defaultdict` to aggregate counts from both `source_document_id` and `target_document_id` aggregations, ensuring each relationship contributes to both documents' counts
- Relationship queries respect existing permission filters (visible document/corpus IDs) and superuser bypass logic
- Frontend uses Apollo's `cache-first` fetch policy to avoid redundant queries if the user hovers multiple times
- The lazy query approach keeps the initial document list query lightweight while maintaining full functionality for users who need relationship details


JSv4:
Thanks for the careful review. Triaged and addressed in 69bbaa4d4. Notes on each item:

### Permissioning regression — **invalid concern**

`DocumentRelationship` does **not** carry its own guardian permissions. The `documents_documentrelationshipuserobjectpermission` / `…groupobjectpermission` tables were intentionally dropped in [`documents/migrations/0029_alter_documentrelationshipuserobjectpermission_unique_together_and_more.py`](https://github.com/Open-Source-Legal/OpenContracts/blob/claude/fix-document-loading-speed-RWVzZ/opencontractserver/documents/migrations/0029_alter_documentrelationshipuserobjectpermission_unique_together_and_more.py), and the model itself documents this at [`documents/models.py:520-522`](https://github.com/Open-Source-Legal/OpenContracts/blob/claude/fix-document-loading-speed-RWVzZ/opencontractserver/documents/models.py#L520-L522):

> Note: DocumentRelationship inherits permissions from source_document, target_document, and corpus - no individual guardian permissions needed.

Visibility is the AND of source-doc, target-doc, and corpus visibility — exactly the filter the new `get_relationship_counts_by_document` applies. The pre-existing `get_relationships_for_document` / `get_visible_relationships` use the same rule (and the test suite's "one-sided visibility" case verifies it).

The reviewer's confusion likely came from a stale docstring on `resolve_all_doc_relationships` that still said "DocumentRelationship has its own guardian permissions". I've refreshed that docstring in this commit so it doesn't mislead the next reader.

### `allDocRelationships` dead code — **invalid concern**

The field is removed from `GET_DOCUMENTS` (the document-list query that was the bottleneck) but is still actively consumed by the `DocumentKnowledgeBase` query (`frontend/src/graphql/queries.ts:2867`) and rendered in `frontend/src/components/knowledge_base/document/DocumentKnowledgeBase.tsx:1473`. Removing it from the schema would break the knowledge-base view. Not dead.

### Missing error state in lazy query — **valid, fixed**

Before this commit:
- A failed fetch left the popup stuck on the "Loading relationships..." fallback (the popup branches on `allDocRelationships.length > 0`; on error `relationshipsData` stays `undefined` and falls into the loading text).
- Every subsequent hover re-fired the failing query because the guard only checked `data` and `loading`.

Fix in `ModernDocumentItem.tsx`:
- Destructure `error` from `useLazyQuery`.
- Add `relationshipsError` to the `handleRelationshipHover` guard.
- Render a distinct `"Couldn't load relationships."` fallback in both the card-view and list-view popups.
- Added a network-error test (`network error short-circuits subsequent hovers`) that fails without the guard because it consumes a second mock that isn't provided.

### `onFocus` not tested — **valid, fixed**

Added `fireRelationshipFocus` helper following the existing hover/click pattern, plus a `focus trigger fires fetch (keyboard navigation)` test.

### Test comment references PR #1440 — **valid, fixed**

Replaced the "PR #1440" header and the misleading "early-return guard" comment in `does not fetch when docRelationshipCount is 0` (the badge isn't rendered at all in that case — the gate is render-time, not handler-time).

### Cache scoping with anonymous users

Non-issue as you noted — context is per-request, so `_none_all` only ever contains data the resolver was already willing to surface to that anonymous request.

All 6 lazy-relationship CT tests and all 6 backend optimizer tests pass locally.

Issue #1447, opened 2026-05-02
Continue draining the mypy baseline established in #1331


JSv4:
`mypy.ini` currently silences ~335 modules from the original 357-file / 7,208-error baseline established in #1331. New modules already pass — the gap is the legacy surface.

Recent issues (#1335 and follow-ups) graduated several packages to ≥70% return-annotation coverage. Suggest continuing that cadence by picking the next 5–6 packages most often touched by feature work, on the theory that drained packages stop accumulating new errors faster.

### Candidate next packages

Selected for high churn in CHANGELOG and high blast radius in feature work:

- `opencontractserver/extracts/`
- `opencontractserver/annotations/`
- `opencontractserver/documents/`
- `config/graphql/` (or a focused subset — e.g. `*_queries.py` first)
- `opencontractserver/llms/tools/`

### Per-package workflow

1. Enumerate current errors for the package.
2. Fix or annotate (preferring real annotations over `# type: ignore`; where ignores are needed, scope them with an error code).
3. Remove the package's `[mypy-...]` block(s) from `mypy.ini`.
4. Document delta in CHANGELOG following the existing convention (e.g. `34.4% → 71.9%`).

### Acceptance

- Per-package PR with ≥70% return-annotation coverage and the corresponding `mypy.ini` block(s) removed.
- Pre-commit `mypy` run on the full project surface continues to pass.

JSv4:
Tracking the remaining work as child issues — each is a per-package graduation PR scoped to fit the workflow in this issue's description:

**Already in flight / merged-or-open against this issue (PRs):**
- #1468 — \`llms.tools.*\` (re-do of closed #1460 after the core_tools refactor in #1461)
- #1469 — eight small single-file packages (config + worker_uploads + feedback + …)
- #1470 — \`shared.{QuerySets,fields,mixins}\`
- #1472 — \`agents.* + analyzer.* + discovery.views\` (also fixes real bug #1471)
- #1473 — \`corpuses.*\` (5 files, -166 baseline lines)
- #1474 — \`annotations.*\` (7 files)
- #1475 — \`documents.*\` (8 files)

**Remaining child issues (per-package graduation):**
- #1477 — \`shared.{Managers,decorators}\` (deferred from #1470 — needs Protocol scaffolding)
- #1478 — \`conversations.models\` (deferred from #1469 — 24 errors)
[...]

PR #1474, opened 2026-05-03
typing: graduate opencontractserver.annotations.* (refs #1447)

JSv4:
Continues draining the mypy baseline. Graduates all seven \`opencontractserver.annotations.*\` modules.

## Removed from \`mypy.ini\`

- \`opencontractserver.annotations.{admin,compact_json,models,query_optimizer,utils}\`
- \`opencontractserver.annotations.management.commands.{migrate_structural_annotations,populate_content_modalities}\`

## Added to \`mypy.ini\`

\`\`\`
[mypy-opencontractserver.annotations.models]
disable_error_code = django-manager-missing
\`\`\`

Mirrors the existing \`users.models\` exception — django-stubs / django-tree-queries cannot resolve the \`Corpus.label_set → LabelSet.used_by_corpuses\` reverse relation.

## Pruned from \`docs/typing/mypy_baseline.txt\`

27 lines (7124 → 7097).

## Per-file fixes

| File | Fix |
|------|-----|
| \`compact_json.py\` | Collapsed \`isinstance(bounds, dict)\` re-binding into a single \`bounds_raw → bounds\` assignment; renamed \`token_indices: list[int]\` to drop the redundant re-annotation. |
| \`query_optimizer.py\` | Narrowed \`path_query.first()\` into a separate \`first_path: DocumentPath \| None\` binding before the is-None guard so the \`DocumentPath \| None\` value doesn't flow into the previously \`DocumentPath\`-typed \`document_path\` variable. |
| \`models.py\` | Added \`author_obj: AbstractBaseUser\` annotation in \`Note.update_content_with_revision\`; scoped \`# type: ignore[misc]\` on three manager-override \`objects = ...Manager()\` declarations (\`Embedding\`, \`Annotation\`, \`Note\`); scoped \`# type: ignore[assignment]\` on the \`StructuralAnnotationSet.creator = ForeignKey(..., null=True)\` redeclaration (intentional override of \`BaseOCModel.creator\`); scoped \`# type: ignore[misc]\` on the lambda inside \`transaction.on_commit(...)\` and on \`author=author_obj\` passed to \`NoteRevision.objects.create(...)\`; dropped the \`: str\` re-annotation on \`embedder_path = CharField(...)\` (CharField field declarations don't accept Python type annotations on the LHS). |
| \`utils.py\` | Typed \`all_tokens: list[Any] = []\` accumulator. |
| \`populate_content_modalities.py\` | Typed \`all_token_refs: list[Any] = []\`. |
| \`migrate_structural_annotations.py\` | Widened the heterogeneous \`stats\` dict to \`dict[str, Any]\` so the mixed counter / error-list values type-check. |

## Acceptance (per #1447)

- [x] \`mypy.ini\` block(s) removed
- [x] Pre-commit \`mypy\` runs clean on full project surface
- [x] No runtime behavior changes intended

\`\`\`
\$ docker compose -f test.yml run --rm django python -m mypy --config-file mypy.ini opencontractserver config
Success: no issues found in 1019 source files

\$ pre-commit run --all-files
mypy.....................................................................Passed
\`\`\`

## Test plan
- [ ] CI mypy job passes
- [ ] CI backend tests pass


JSv4:
Thanks for the review. Addressing the three items:

**1. How `admin.py` went clean without a code change in this PR** — `opencontractserver/annotations/admin.py` was already made type-clean in PR #1355 ("Add return-type annotations to core models and import/export pipeline", commit `36921acbc`). That PR added `from __future__ import annotations`, return-type annotations on `lookups` / `queryset` / `dimension` / `total_embeddings` / `modality_display` / `get_queryset`, and `# type: ignore[attr-defined]` on the `short_description` / `admin_order_field` callable attributes — exactly the pattern you described. Per #1355's own commit message, mypy graduation was deliberately deferred to a follow-up. This PR is that follow-up: the file already passes mypy on its own, so removing the `[mypy-opencontractserver.annotations.admin]` block from `mypy.ini` and the nine baseline lines is purely a cleanup.

**2. `list[Any]` → `list[dict[str, int]]`** — addressed in commit `a847bcb43`. The accumulator shape is fixed at `{"pageIndex": int, "tokenIndex": int}`, so the tighter annotation carries real information. `compute_content_modalities` accepts `list[dict[str, Any]]`, which `Any`'s bidirectional consistency lets `dict[str, int]` flow into without complaint. Also dropped a now-unused `Any` import in `populate_content_modalities.py`.

**3. Comment on `author=author_obj  # type: ignore[misc]`** — addressed in the same commit. Added a comment explaining that django-stubs' `create()` infers `AUTH_USER_MODEL` exactly (not the abstract `AbstractBaseUser` base), even though `author_obj` is a runtime-correct subtype.

PR #1489, opened 2026-05-03
Raise zip import sidecar size limit to 50MB and make it env-configurable

JSv4:
## Summary

This PR increases the per-sidecar JSON size limit during zip imports from 10 MB to 50 MB and makes it configurable via environment variable, addressing a silent data loss issue where oversized annotation sidecars were being dropped without proper error signaling.

## Changes

- **Increased default sidecar size limit**: Raised `ZIP_MAX_SIDECAR_SIZE_BYTES` from 10 MB to 50 MB to accommodate larger annotation corpora
- **Added environment variable configuration**: The limit is now read from the `ZIP_MAX_SIDECAR_SIZE_BYTES` environment variable via `django-environ` in `config/settings/base.py`, following the same pattern as `MAX_WORKER_UPLOAD_SIZE_BYTES`
- **Updated constants module**: Bumped the fallback default in `opencontractserver/constants/zip_import.py` to 50 MB to ensure non-Django importers use the same default
- **Improved documentation**: Added comments explaining the configuration override mechanism and the rationale for the limit

## Implementation Details

The issue was that when sidecars exceeded the 10 MB limit, `_read_sidecar()` would raise a `ValueError` that was caught and recorded in the `errors` field, but the importer would still return `success: True, completed: True`. This meant callers had no signal that annotations were silently lost.

The fix mirrors the existing pattern used for other configurable size limits in the codebase:
1. Define the setting in `config/settings/base.py` with `django-environ` for env var support
2. Reference it in the constants module via `getattr(settings, ...)`
3. Provide a sensible default that works for both Django and non-Django contexts

Existing tests that patch the constant directly remain unaffected.

https://claude.ai/code/session_01EyZxjL4oMERhxmpDwUZ5Yt


JSv4:
Addressed both flagged items in 53ff70b27:

**1. Silent `success: True` with sidecar errors (the real bug):**
- `import_zip_with_folder_structure` now requires `annotation_sidecars_errored == 0` in addition to the existing `files_errored == 0` and user-cap check. Sidecar size-limit failures, malformed JSON, and missing-labels-for-sidecar-annotations all now correctly drop `success` to `False`.
- `process_documents_zip` had the same shape of bug (only checked `user_cap_reached_mid_processing`, ignored `error_files`); fixed to require `error_files == 0` too.
- `relationship_errors` are intentionally not folded in — those documents *are* imported and the partial-failure case is already surfaced via `relationships_skipped` + `relationship_errors`. Wider scope can be a follow-up if desired.
- New regression test `test_sidecar_error_drops_overall_success_flag` covers both the failure path (oversized sidecar → `success=False`) and the happy path (clean sidecar at default limit → `success=True`).
- `test_skip_pipeline_without_labels_json` previously asserted the broken contract (`success=True` with `annotation_sidecars_errored == 1`); updated to assert `success=False` with a comment explaining why per project conventions.

**2. Other zip constants not env-configurable:**
Added `int(env(...))` entries in `config/settings/base.py` for `ZIP_MAX_FILE_COUNT`, `ZIP_MAX_TOTAL_SIZE_BYTES`, `ZIP_MAX_SINGLE_FILE_SIZE_BYTES`, `ZIP_MAX_COMPRESSION_RATIO`, `ZIP_MAX_FOLDER_DEPTH`, `ZIP_MAX_FOLDER_COUNT`, `ZIP_MAX_PATH_COMPONENT_LENGTH`, `ZIP_MAX_PATH_LENGTH`, and `ZIP_DOCUMENT_BATCH_SIZE`. Defaults match what `constants/zip_import.py` already used so there is no behavior change at the defaults. Updated the constants-module docstring to point readers at the env-var path.

All 235 sidecar / zip-import / corpus-export tests pass.

PR #1637, opened 2026-05-13
Centralize authorization logic in Manager.user_can() API

JSv4:
## Summary

Consolidates permission checking logic into a single source of truth by introducing a new `Manager.user_can(user, instance, permission)` API and corresponding instance method `obj.user_can(user, permission)`. This replaces scattered permission checks throughout `DocumentFolderService` with centralized authorization that guarantees filter (`visible_to_user`) and check operations are provably aligned.

## Key Changes

- **New centralized authorization function** (`_default_user_can` in `utils/permissioning.py`):
  - Single source of truth for standard permission rules (superuser bypass, creator ownership, `is_public` read-only, guardian permissions)
  - Handles `None`, `AnonymousUser`, and authenticated users uniformly
  - Supports all `PermissionTypes` (READ, CREATE, UPDATE, DELETE, COMMENT, PUBLISH, PERMISSION, CRUD, ALL)
  - Preserves critical read/write asymmetry: `is_public=True` grants READ only, never writes

- **Manager and QuerySet integration**:
  - `BaseVisibilityManager.user_can()` delegates to `_default_user_can`
  - `PermissionedTreeQuerySet.user_can()` provides same API for tree-based models
  - `BaseOCModel.user_can()` instance method for ergonomic `obj.user_can(user, perm)` syntax

- **Removed duplicate permission logic** from `DocumentFolderService`:
  - Deleted `check_corpus_read_permission()`, `check_corpus_write_permission()`, `check_corpus_delete_permission()` (146 lines)
  - Replaced 30+ call sites with `corpus.user_can(user, PermissionTypes.*)` throughout the service

- **Comprehensive regression test suite** (`test_authorization_invariants.py`):
  - Pins filter/check equivalence: `user_can(READ)` ≡ `visible_to_user().exists()`
  - Validates manager and instance surfaces agree
  - Tests superuser bypass, creator ownership, public corpus read-only, anonymous access, explicit guardian grants
  - Covers all (user, corpus, permission) combinations to prevent silent permission widening

## Implementation Details

- `_default_user_can` is the single decision point for all standard-rule models; per-model overrides (e.g., `AnnotationManager.user_can`) add structural/privacy rules before delegating here
- Maintains backward compatibility: existing `visible_to_user` queryset filter unchanged
- All permission checks now route through the same code path, eliminating drift between filter and check operations
- Scope: Phase 0 covers Corpus only; subsequent phases extend to Document, Annotation, Relationship, etc.

https://claude.ai/code/session_01C3ZJNaePzxHGQEpXzb5XhH


JSv4:
Addressed three of the five items from the latest review in 1fdc8d81d:

- **Item 1** (misleading `is_public` fold-in comment): rewrote the comment in `_default_user_can`'s CRUD/ALL branch to be honest about the fact that `get_users_permissions_for_obj` already injects `read_<model>` when `is_public=True`, so today the fold-in is a no-op — kept defensively in case that helper is ever refactored.
- **Item 3** (`UserCanMixin` QuerySet semantics): added a docstring note that on a QuerySet, `user_can()` is a single-object check that ignores the queryset's current WHERE clause.
- **Item 5** (`CorpusFolder.user_can` scope): added a docstring note that `folder.user_can(...)` checks folder-level perms, NOT the parent corpus — write-op call sites should keep using `folder.corpus.user_can(...)`.

- **Item 2** (per-call DB cost) — filed as #1640 for a Phase-1 follow-up. The Phase 0 blast radius is bounded (one check per `folder_service.py` entry point, no per-row loops), and request-scoped caching mirroring `ConversationQueryOptimizer` is the right shape but premature here. Issue covers two-tier mitigation (per-instance memoization + request-scoped `PermissionQueryOptimizer`) and acceptance criteria.
- **Item 4** (`_default_user_can` underscore naming): leaving as-is for this PR. Cosmetic, and the docstring already explains the convention.

Issue #1921, opened 2026-06-04
Optimize per-agent-construction DB credential read in `build_agent_model`


JSv4:
Tracking issue split out from the PR #1897 review (#1917, item 8).

## Context

`opencontractserver/llms/model_factory.py::_get_db_credentials` calls `PipelineSettings.get_instance()` on **every** `build_agent_model` call — i.e. every chat message, every structured-output call, every memory-curation task.

`PipelineSettings.get_instance()` is already backed by Django's cache with a 5-minute TTL, so this is **not** a raw DB round-trip every time. But each call still incurs:

- a cache read, and
- Fernet decryption of the stored secrets (via `get_full_component_settings`).

This is **not a blocking concern** — it was explicitly deemed acceptable in the review — but is worth tracking as a future optimization.

## Possible directions (not yet decided)

- A short-lived in-process memoization of the resolved per-provider credentials, keyed on the `PipelineSettings` row's `modified`/version so it invalidates the moment a superuser rotates a key (the live-rotation guarantee of #1897 must be preserved — credentials may already be up to 5 min stale via the existing cache TTL, but must not become *permanently* stale).
- Resolving credentials once per agent/run rather than once per `make_pydantic_ai_agent` call, if the call graph allows it.

## Constraints

- Must not regress the **live, DB-configurable** behavior from #1897: a rotated/cleared key must take effect without a redeploy.
- Must keep the **DB-wins / env-fallback** precedence intact.

No action required now; filed so the optimization is not forgotten.

Issue #1920, opened 2026-06-04
Lift the `aiohttp<3.14` cap once `vcrpy` supports the 3.14 stream API


JSv4:
Tracking issue split out from the PR #1897 review (#1917, item 10).

## Context

`requirements/base.txt` pins:

```
aiohttp>=3.13,<3.14  # transitive via llama-index-core; cap below 3.14 because aiohttp 3.14.0 removed aiohttp.streams.AsyncStreamReaderMixin, which the pinned vcrpy==8.1.1 aiohttp stub imports at module load (AttributeError on import → whole pytest suite fails). Lift this cap together with a vcrpy bump that supports aiohttp 3.14+.
```

The cap was added in #1914 (see `changelog.d/1914-aiohttp.fixed.md`). `aiohttp` is otherwise an unpinned transitive dep (via `llama-index-core`), so a fresh CI dependency resolution will pick up the newest release; 3.14.0 removed `aiohttp.streams.AsyncStreamReaderMixin`, which the pinned `vcrpy==8.1.1` aiohttp stub imports at module load — aborting the entire pytest collection with `AttributeError`.

`vcrpy==8.1.1` is pinned in `requirements/local.txt`.

## Action

When `vcrpy` is bumped to a release that supports the aiohttp 3.14 stream API:

- [ ] Remove (or widen) the `aiohttp<3.14` cap in `requirements/base.txt`.
- [ ] Bump `vcrpy` in `requirements/local.txt` to the compatible version.
- [ ] Confirm the pytest suite collects and runs cleanly with `aiohttp>=3.14`.

This is purely a dependency-hygiene follow-up so the temporary cap is not silently carried forever.

PR #1922, opened 2026-06-04
LLM credential review follow-ups: clear_cache() API + test hardening (#1917)

### LINKED ISSUE #1917: LLM Singleton Follow Up
## Code Review: PR #1897 — LLM provider DB-configurable credentials & endpoints

**Overview**

This PR closes a clear gap in the pipeline-component architecture: parsers and embedders already had live-configurable credentials via PipelineSettings; LLM providers did not. The approach is sound — add a Settings dataclass to each provider, build a model_factory.py that resolves DB-wins/env-fallback, wire it into all call sites. The backward-compatibility guarantee (env-fallback when DB is empty) is especially important.

---

**Bugs / Correctness**

**1. parse_model_spec treats bare specs as OpenAI — misleading test comment**

opencontractserver/llms/llm_registry.py:34 makes bare model names like gpt-4o silently resolve to (openai, gpt-4o) rather than raising ValueError. The test test_colonless_spec_returned_unchanged therefore does *not* exercise the ValueError catch path — it exercises the no DB creds -> return spec string path. The test still passes and the behavior is correct, but the comment A spec without a provider:model colon cannot be parsed is inaccurate. Worth fixing so future readers understand what is actually being tested.

**2. Silent failure when base_url is set but api_key is left blank**

_get_db_credentials only includes keys with non-empty values. If a user sets a custom base_url (pointing at a gateway) but leaves api_key blank, creds = {base_url: http://...} and _construct_model is called with api_key = None. For OpenAI/Anthropic, pydantic-ai then falls back to the env var — documented intentional behavior. But if the gateway does not use the standard env var, the request will fail at runtime with no warning at construction time. A single logger.warning when base_url is not None and api_key is None would make this misconfiguration much easier to diagnose.

**3. All 5 call sites are covered** ✅

pydantic_ai_agents.py has exactly 3 make_pydantic_ai_agent call sites on main (lines 1766, 3004, 3468), all updated in this diff. memory_tasks.py has 2 calls sharing one abuild_agent_model resolution. The all five claim checks out.

---

**Code Quality**

**4. except Exception is too broad — twice**

model_factory.py uses bare except Exception: in both _get_db_credentials (~line 194) and build_agent_model (~line 311) with a log-and-fallback strategy. The intent (never let a misconfiguration crash the chat path) is correct, but bare except Exception also catches MemoryError, RecursionError, and programming errors like TypeError from a wrong function signature. Consider tightening to the expected exception types (OSError, ImportError, AttributeError, ValueError, RuntimeError) so genuine bugs surface rather than silently degrading.

**5. Tests use PipelineSettings._invalidate_cache() private API**

test_llm_model_factory.py calls PipelineSettings._invalidate_cache() directly in several setUp methods. This makes the tests fragile to internal cache-implementation changes. If the class can expose a documented clear_cache(), prefer that. If not, a comment noting this is test-only plumbing would make the intent explicit.

**6. Fragile model-type assertion**

test_db_creds_build_real_pydantic_ai_model asserts type(result).__name__.endswith(Model). The pydantic-ai class-naming convention is not part of its public API. If a future release renames OpenAIChatModel, this assertion fails silently. The assertNotIsInstance(result, str) line above is already the right check; the endswith(Model) line could be dropped or replaced with assertIsInstance against a pydantic-ai base class if one is exported.

**7. reset_registry() without symmetric teardown**

setUp calls reset_registry() but no tearDown restores it. Adding self.addCleanup(reset_registry) would make cleanup explicit and symmetric.

---

**Performance**

**8. DB credential reads on every agent construction**

_get_db_credentials calls PipelineSettings.get_instance() on every make_pydantic_ai_agent call (every chat message, every structured call). PipelineSettings uses Django cache internally so this is not a raw DB round-trip every time, but it is a cache read and Fernet decryption per call. Not a blocking concern now, but worth tracking as a future optimization.

---

**Security**

**9. base_url is not validated before being passed to the provider**

A misconfigured base_url is passed directly to the pydantic-ai Provider constructor. Since only superusers can write this setting, the threat model is low, but a simple URL-format check (urlparse(base_url).scheme in http/https) before _construct_model would be a good defensive addition.

---

**Packaging**

**10. aiohttp cap is an unrelated fix bundled into a feature PR**

requirements/base.txt adds aiohttp>=3.13,<3.14 to prevent a vcrpy==8.1.1 / aiohttp 3.14+ compatibility break. The fix is correct and well-documented inline. However, it must be lifted alongside a future vcrpy bump. Consider opening a follow-up issue to track the cap so it does not get forgotten.

---

**Summary**

The architecture is clean and the feature fills a real operational gap. Three items before merge:

1. **Tighten the except Exception blocks** in both _get_db_credentials and build_agent_model — a wrong function signature or ImportError from a pydantic-ai version bump should not silently fall back to env credentials.
2. **Fix the test comment** for test_colonless_spec_returned_unchanged — the actual path exercised is no DB creds for openai provider -> env fallback, not a ValueError from an unparseable spec.
3. **Track the aiohttp cap** as a follow-up issue alongside a vcrpy upgrade.

The rest are minor quality items suitable for a follow-up PR.

_Originally posted by @claude[bot] in https://github.com/Open-Source-Legal/cite/issues/1897#issuecomment-4618841966_

JSv4:
Closes #1917.

Resolves the outstanding follow-up items from the PR #1897 code review (issue #1917 — *LLM Singleton Follow Up*).

## Status of all 10 review items

| # | Item | Resolution |
|---|------|-----------|
| 1 | Misleading `test_colonless_spec_returned_unchanged` comment | ✅ Already corrected before #1897 merged (`test_llm_model_factory.py`) |
| 2 | Warning when `base_url` set but `api_key` blank | ✅ Already in merged code (`model_factory.py` — `logger.warning` for `api_key is None and provider_key != "ollama"`) |
| 3 | All 5 call sites covered | ✅ Confirmation only |
| 4 | `except Exception` too broad (×2) | ✅ Already in merged code — narrowed to `_DB_READ_RECOVERABLE_ERRORS` / `_MODEL_BUILD_RECOVERABLE_ERRORS` tuples |
| **5** | Tests use private `_invalidate_cache()` | **✅ This PR** |
| **6** | Fragile `endswith("Model")` assertion | **✅ This PR** |
| **7** | `reset_registry()` without symmetric teardown | **✅ This PR** |
| 8 | Per-call DB credential read | 📋 Tracked in #1921 |
| 9 | `base_url` not validated | ✅ Already in merged code — `urlparse(base_url).scheme in ("http","https")` check |
| 10 | aiohttp `<3.14` cap tracking | 📋 Tracked in #1920 |

Items 1, 2, 4, and 9 were already addressed during #1897's review cycle (before merge), so this PR is scoped to the remaining "minor quality items suitable for a follow-up PR" (5, 6, 7) plus the two tracking issues the review asked for (8 → #1921, 10 → #1920).

## Changes

**Item 5 — drop the private-API reach.** Added a public, documented `PipelineSettings.clear_cache()` classmethod (`opencontractserver/documents/models.py`) as the canonical way to drop the cached singleton. The existing `_invalidate_cache()` becomes a thin backwards-compatible alias that delegates to it — `save()` and the other call sites across the test suite keep working unchanged (no churn). `save()` itself now calls `clear_cache()` directly (registered on `transaction.on_commit` without a `self`-capturing lambda). `test_llm_model_factory.py` now calls `clear_cache()`, which also let me delete a redundant `cache.delete(CACHE_KEY)` + `_invalidate_cache()` double-call in each `setUp` and remove the now-unused `cache` import.

**Item 6 — robust model-type assertion.** `test_db_creds_build_real_pydantic_ai_model` now asserts `assertIsInstance(result, pydantic_ai.models.Model)` (the exported abstract base) instead of `type(result).__name__.endswith("Model")`, which would have silently broken on an upstream rename of `OpenAIChatModel`.

**Item 7 — symmetric registry teardown.** Every `setUp` now registers `self.addCleanup(reset_registry)` (and, where it clears the singleton cache, `self.addCleanup(PipelineSettings.clear_cache)`) so the pipeline-component registry and cache are restored after each test, not just reset before it.

## Follow-up refinements (review)

Subsequent commits applied review nits: route `save()` invalidation through `clear_cache()` directly; drop the now-redundant post-save `clear_cache()` in the test helper; register the `on_commit` callback without a closure; move the `pydantic_ai.models.Model` import to module level; add symmetric cache teardown; drop a redundant `assertNotIsInstance(result, str)` (implied by the `Model` assertion); and small docstring/comment accuracy tweaks. Migrating the remaining `_invalidate_cache()` callers and deleting the alias is tracked in #1928.

## Verification

- `python -m py_compile` ✅, `black --check` ✅, `flake8` ✅ on both changed files.
- `python scripts/collate_changelog.py --check` ✅ (changelog fragment valid).
- Verified the item-6 assertion against the pinned pydantic-ai (`>=1.105.0,<2`): `from pydantic_ai.models import Model` imports, `OpenAIChatModel` is a subclass, and a real credentialed model built exactly as `_construct_model` does passes `isinstance(model, Model)`.

A `changelog.d/1917-llm-singleton-followup.changed.md` fragment is included.

### DIFF
--- a/changelog.d/1917-llm-singleton-followup.changed.md
+++ b/changelog.d/1917-llm-singleton-followup.changed.md
@@ -1,0 +1,24 @@
+- **LLM provider credential follow-ups (PR #1897 review, issue #1917).** Code-quality
+  hardening for the DB-configurable LLM credential feature:
+  - Added a public, documented `PipelineSettings.clear_cache()` classmethod
+    (`opencontractserver/documents/models.py`) as the canonical way to drop the
+    cached singleton; the existing `_invalidate_cache()` is now a thin
+    backwards-compatible alias that delegates to it. Tests and admin/CLI tooling
+    that write the singleton out-of-band no longer have to reach for the private
+    underscore method.
+  - `opencontractserver/tests/test_llm_model_factory.py`: assert the credentialed
+    model against pydantic-ai's exported abstract base `pydantic_ai.models.Model`
+    (`assertIsInstance`) instead of the fragile `type(result).__name__.endswith("Model")`
+    class-name string, which would have silently broken on an upstream rename of
+    `OpenAIChatModel`; switched cache resets to the new `clear_cache()` (dropping a
+    redundant `cache.delete()` + `_invalidate_cache()` double call); and added
+    `self.addCleanup(reset_registry)` to every `setUp` so the pipeline-component
+    registry is restored symmetrically after each test rather than only reset
+    before it.
+  - The correctness/security items from the same review — narrowing the
+    `except Exception` fallbacks to explicit recoverable-error tuples, the
+    `base_url`-without-`api_key` warning, `urlparse` scheme validation of a
+    DB-configured `base_url`, and the `test_colonless_spec_returned_unchanged`
+    comment — were already addressed in `opencontractserver/llms/model_factory.py`
+    / the test before #1897 merged. The aiohttp `<3.14` cap (#1920) and the
+    per-call credential-read cost (#1921) are tracked as separate follow-up issues.
--- a/opencontractserver/documents/models.py
+++ b/opencontractserver/documents/models.py
@@ -1298,10 +1298,10 @@ class PipelineSettings(django.db.models.Model):
         super().save(*args, **kwargs)
         # Eagerly invalidate cache after save for immediate consistency
         # (required in autocommit mode and Django TestCase which never commits).
-        self._invalidate_cache()
+        self.clear_cache()
         # Also invalidate on commit in case save() runs inside a larger
         # transaction that might roll back and be retried.
-        transaction.on_commit(lambda: self._invalidate_cache())
+        transaction.on_commit(lambda: self.clear_cache())
 
     def delete(self, *args: Any, **kwargs: Any) -> NoReturn:
         """Prevent deletion of the singleton instance."""
@@ -1323,7 +1323,11 @@ class PipelineSettings(django.db.models.Model):
 
     @classmethod
     def _invalidate_cache(cls) -> None:
-        """Backwards-compatible internal alias for :meth:`clear_cache`."""
+        """Backwards-compatible alias for :meth:`clear_cache`.
+
+        Retained for the existing test call sites that predate
+        :meth:`clear_cache`; new code should use the public method.
+        """
         cls.clear_cache()
 
     @classmethod
--- a/opencontractserver/documents/models.py
+++ b/opencontractserver/documents/models.py
@@ -1308,11 +1308,23 @@ class PipelineSettings(django.db.models.Model):
         raise ValidationError("PipelineSettings singleton cannot be deleted.")
 
     @classmethod
-    def _invalidate_cache(cls) -> None:
-        """Invalidate the cached instance."""
+    def clear_cache(cls) -> None:
+        """Drop the cached singleton so the next ``get_instance()`` re-reads the DB.
+
+        The public, documented entry point for forcing cache invalidation.
+        ``save()`` already invalidates automatically, so application code
+        rarely needs this; it exists for tests and admin/CLI tooling that
+        write the singleton row out-of-band (e.g. a direct ``QuerySet.update``
+        or a fixture load) and then need the change visible immediately.
+        """
         from django.core.cache import cache
 
         cache.delete(cls.CACHE_KEY)
+
+    @classmethod
+    def _invalidate_cache(cls) -> None:
+        """Backwards-compatible internal alias for :meth:`clear_cache`."""
+        cls.clear_cache()
 
     @classmethod
     def get_instance(cls, use_cache: bool = True) -> PipelineSettings:
--- a/opencontractserver/documents/models.py
+++ b/opencontractserver/documents/models.py
@@ -1327,8 +1327,9 @@ class PipelineSettings(django.db.models.Model):
     def _invalidate_cache(cls) -> None:
         """Backwards-compatible alias for :meth:`clear_cache`.
 
-        Retained for the existing test call sites that predate
-        :meth:`clear_cache`; new code should use the public method.
+        Retained for existing callers (the test suite and ``conftest.py``)
+        that predate :meth:`clear_cache`; new code should use the public
+        method.
         """
         cls.clear_cache()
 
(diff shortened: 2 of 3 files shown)


JSv4:
Thanks for the review!

**Nit 2 (save() → clear_cache())** — applied in b056707. `save()` now calls `self.clear_cache()` directly; `_invalidate_cache()` stays as a backwards-compatible alias for the existing test call sites.

**Nit 1 (module-level `Model` import)** — keeping it local, intentionally. The premise that "pydantic-ai is already a transitive import via the other `opencontractserver.llms` imports" isn't the case here: `model_factory.py` does *all* its pydantic-ai imports lazily/in-function (none at module level — it's a deliberate design point so the pipeline registry stays importable during early startup/migrations), and the provider modules the test imports don't import `pydantic_ai` either. So the test module currently pulls in `pydantic_ai` only inside the one test that builds a real model. Keeping the import local mirrors `model_factory`'s lazy philosophy and keeps the `TestBuildAgentModelEnvFallback` class — which exercises the bare-string passthrough that never touches pydantic-ai — free of a module-load pydantic-ai import.

---

JSv4:
Addressed the latest review round. This is the final intended state of the PR — all CI green on `5d5321b`, `mergeable_state: clean`.

- **`on_commit` lambda captures `self`** — ✅ applied in `5d5321b`. `save()` now registers `transaction.on_commit(PipelineSettings.clear_cache)` directly (no closure), since `clear_cache` is a classmethod.
- **Redundant post-save `clear_cache()` in `_configure_openai_creds`** — ✅ already removed in `0ac73bb` (save() invalidates eagerly).
- **`DeprecationWarning` on `_invalidate_cache()`** — holding. `_invalidate_cache()` is an intentionally **kept** backwards-compatible alias, not a deprecation: the PR deliberately leaves the ~11 existing callers untouched (which the first review correctly praised as "the right call — zero churn"). A `DeprecationWarning` signals impending removal and would push that exact churn onto those files while emitting warnings on every run. The docstring nudge ("new code should use the public method") is the right level for an alias we plan to keep.
- **Full-suite `addCleanup(reset_registry)` migration** — out of scope by design (as the review notes); this PR only hardens the test file it touches. Migrating the broader suite is a separate follow-up to avoid unrelated churn.
- **Module-level `Model` import** — holding, per my [earlier note](https://github.com/Open-Source-Legal/cite/pull/1922#issuecomment-4620304123): the local import matches this codebase's deliberate pydantic-ai lazy-import discipline and keeps the env-fallback test class free of a module-load pydantic-ai import.

---

PR #1976, opened 2026-06-10
Corpus reference enrichment: cross-corpus law linking, authority corpora, corpus-scoped analyzers

JSv4:
## What this is

The reference-web substrate for the corpus intelligence initiative: drop a collection of SEC filings into a corpus and the system detects, normalizes, and **resolves every explicit reference** — exhibit cross-references become in-app document links, statutory citations become cross-corpus links to the *actual section text* of the governing law, ingested as first-class corpus documents.

The Governance Graph

*Generated from real data: 5 filing corpora (SpaceX, Fervo Energy, Cerebras, Quantinuum, and 87 Anthropic-linked Form D SPVs — 254 documents), 1,943 references resolved, 1,501/1,726 statutory citations linked to in-system law across 6 authority corpora (DGCL, Securities Act, SEC Rules/17 C.F.R., Exchange Act, ICA, IRC). The amber cascade on the left is 957 Reg D citations from the SPV swarm resolving into Rule 504/506.*

## What's inside

**Engine** (`opencontractserver/enrichment/`)
- High-precision citation grammars: suffix ("Section 145 of the DGCL"), prefix ("Securities Act Section 4(a)(5)"), statute-internal relative ("§ 251 of this title", keyed by the document's own authority), bare SEC rules ("Rule 506(b)" → `sec-rule:506(b)`); plus exhibit refs, internal section refs, and opt-in defined-term definition sites (599 distinct terms found, 46 shared across companies).
- Idempotent writer: mention annotations (deduped by document/label/span-start so alias growth can't duplicate), `OC_REFERENCES` relationships, `DocumentRelationship` rollups (these feed #1969's `DocumentGraphGlimpse` directly), and `CorpusReference` rows.

**Cross-corpus law linking**
- New `CorpusReference` model (migration `annotations/0078`) — cross-document / cross-corpus / external-law links with an indexed `canonical_key` join key.
- `AuthorityCorpusBootstrapper`: statute sections as keyed text documents (`custom_meta.canonical_key`), idempotent (skip / version-up on amendment / self-healing restamp), with subsection→section resolution fallback (`dgcl:122(17)` → `dgcl:122`).
- Data-driven authority alias registry (`custom_meta.authority_aliases`): **adding a body of law is a bootstrap call, zero code** — pinned by an NYBCL end-to-end test.
- `link_external_references`: late-binding upgrade of EXTERNAL citations to RESOLVED cross-corpus links; runs inside `apply()` and is re-runnable as authorities appear.

**Analyzer framework**
- New `@corpus_analyzer_task` decorator — corpus-scoped sibling of `@doc_analyzer_task` (one run per Analysis, task owns its writes, wrapper manages the Analysis lifecycle). Dispatch, startup auto-sync, and `analyzer.W001` cover both flavours.
- Enrichment registered as a real task-based Analyzer via a thin adapter — dispatchable through `run_task_name_analyzer` and `CorpusAction`.

**Surfacing**: `scan_corpus_references` / `apply_corpus_reference_enrichment` agent tools (approval + write gated), read-only `corpusReferences` GraphQL query, `demo/` pipeline (seed JSONs with real statutory text, bootstrap/import/export scripts, standalone D3 visualization).

**Tests**: 110 across extractor/resolver/writer/tools/authorities/linking/analyzer-integration; pre-commit + mypy + system checks green.

## ⚠️ Callout: integration next steps (post-#1969)

This PR deliberately does **not** touch #1969. The cross-corpus graph above is a standalone D3 artifact today; the data layer it reads (`CorpusReference` + `DocumentRelationship`) is all in this PR. Once #1969 lands, the path in-app is:

1. ✅ **`governanceGraph` GraphQL query** — **shipped in this PR** (`bd0b01d16`): corpus-scoped node-link query over `DocumentRelationship` + resolved `CorpusReference` rows, mirroring `demo/export_governance_graph.py`. Assembly in `GovernanceGraphService` (visibility-enforced: invisible targets degrade to ghost nodes, invisible sources drop their edges), relay-id encoding in the resolver, degree-capped at `GOVERNANCE_GRAPH_MAX_NODES`. Global (cross-corpus-home) variant remains future work.
2. **Governance Graph panel** on the Corpus Intelligence home — stacked PR *after* #1969, adding a new panel alongside `DocumentGraphGlimpse` (not modifying it); the demo's deterministic D3 composition ports directly.
3. **CorpusAction auto-bootstrap** — wire a corpus action to the `corpus_reference_enrichment` analyzer so dropping documents into a corpus grows the web automatically (machinery in this PR; needs a UI affordance + default-action decision).
4. **Citation → statute reading experience** — mention `link_url`s already route in-app; a reference side-panel (inbound/outbound citations per document, powered by `corpusReferences`) closes the loop.
5. **Known gaps tracked for follow-up**: USC→act-section mapping for federal statute self-citation ("section 78p of this title"), usage→definition term linking (volume controls needed), DGCL citation frontier depth-3+ (§253, §278…), defined-term cross-corpus UI.

## Test plan

- `pytest opencontractserver/tests/test_enrichment_*.py opencontractserver/tests/test_analyzer_app_coverage.py -n 0` — 110 passing
- Live-proven end-to-end: import → enrich → bootstrap → relink → graph export on the 5-corpus dataset above (see `docs/test_scripts/corpus_reference_enrichment.md` and `demo/`)


JSv4:
## Review response (addressed in 17469c06e, after merging main in 00e23a046)

**🔴 GraphQL traversal visibility** — Verified: nested FK traversal is framework-enforced. graphene-django 3.2.3's FK converter routes related-object resolution through `_type.get_node(info, pk)` → `get_queryset`, and both `DocumentType.get_queryset` and `AnnotationType.get_queryset` filter via `BaseService.filter_visible_qs`. Added `CorpusReferenceTraversalVisibilityTests` pinning the invariant: a reader of a public corpus sees the reference rows (corpus-as-gate) but gets `null` for a `targetDocument` they lack document-level READ on, while the owner resolves it.

**🔴 Raw `Corpus.objects.get` in `link_external_references` / `bootstrap`** — Fixed. `EnrichmentService._load`, `link_external_references`, and `AuthorityCorpusBootstrapper.bootstrap` now fetch via `Corpus.objects.visible_to_user(user).get(pk=...)`, so invisible and nonexistent corpora raise the same `Corpus.DoesNotExist` (no existence oracle). Error-path tests added for both entry points.

**🟡 `_terms()` starvation** — Fixed, it was a real bug: the intent is a *total* per-document cap, but the per-regex loop could exhaust it on parentheticals before the means regex ever ran. Matches from both grammars are now merged in document order before capping (first N definition sites win regardless of form), documented in the docstring, and pinned by two boundary tests (means-form before the cap survives; after the cap is dropped).

**🟡 Service location** — Moved to `opencontractserver/enrichment/services/` (`corpus_reference_service.py` + `enrichment_service.py`, re-exported from `__init__`), matching the repo-wide convention; all importers and mock-patch targets updated. `CorpusReferenceService` stays in `enrichment/` deliberately: it is the read surface of the enrichment feature, and the `annotations` app has no knowledge of enrichment semantics — same pattern as other cross-app feature services. Bonus dedup: the local `OC_SECTION_LABEL` redeclaration now imports the canonical constant from `constants/annotations.py`.

**🟡 Choices duplication** — Fixed. `REFERENCE_TYPE_CHOICES` / `RESOLUTION_STATUS_CHOICES` discriminator values are now built from `enrichment/constants.py` (`REF_*`, `STATUS_*`); only the human-readable labels live in `models.py`. `enrichment/constants.py` is import-pure so no cycle, and `makemigrations --check` confirms the change is migration-neutral.

**🟢 `visible_to_user` subquery scale** — No change. The `corpus__in=visible_to_user(...)` shape matches the platform's current pattern; flagged for revisit if profiling shows it hot at many-corpora scale.

**🟢 `corpus_analyzer_task` corpus check** — Tightened. The existence check is now `BaseService.filter_visible(Corpus, analysis.creator).filter(id=corpus_id).exists()` — a task can no longer operate on a corpus the analysis owner couldn't see, with the same error message either way.

**🟢 Choices defined after the model class** — No change: the premise doesn't match the code. Both choice lists are defined immediately *before* `class CorpusReference` (`annotations/models.py:1884-1894`).

**Test coverage gaps** — All three covered: traversal-leak test, `DoesNotExist` error-path tests for `link_external_references` and `bootstrap(corpus_id=...)`, and the `_terms` cap-boundary pair.

Verification: 95 tests green across the enrichment + analyzer suites (`-n 0`), pre-commit (incl. mypy) green, `manage.py check` green (E001 clean; the two remaining warnings are environment-state: unsynced analyzers in the fresh test container and a missing LlamaParse key).

Note for follow-up: `makemigrations --check` flags pending `corpuses` index-rename churn for `CorpusVote` (`corpuses_co_session_idx` → hashed name) that arrived with the merge from main — unrelated to this PR.

JSv4:
## Review response (296423e3a)

**1. Retry swallowed as permanent failure** — Confirmed and fixed. `Retry` extends `Exception`, so the wrapper stamped `Analysis.status=FAILED` before the retry ran. `corpus_analyzer_task` now catches `Retry` first and re-raises untouched, mirroring `doc_analyzer_task`. Added the regression test you suggested: `test_celery_retry_does_not_mark_analysis_failed` pins `RUNNING` status and a clean `error_message` surviving a `Retry`.

**6. `link_url` raw integer PKs** — Verified, and it's worse than "worth confirming": **no frontend route serves `/corpus/{id}/document/{id}`** (`App.tsx` only routes slug shapes; that path falls into the `*` catch-all → 404). Both writers (`EnrichmentWriter._get_or_create_mention` and `_link_external`) now emit the canonical `/d/{corpus.creator.slug}/{corpus.slug}/{document.slug}` path via a new shared helper (`opencontractserver/utils/frontend_paths.py::document_in_corpus_path`, mirroring the frontend's `buildCanonicalPath`); cross-corpus law links point into the authority corpus, and missing slugs skip the link rather than writing a broken one. Both link-shape tests updated TDD-first (watched them fail on the old format).

**2. `_link_external` O(N) saves** — Fixed: refs and mentions accumulate and flush via two `bulk_update` calls, with `modified` stamped explicitly since `bulk_update` bypasses `auto_now`.

**3. `_sections_for` N+1** — Fixed: replaced by `_sections_by_doc`, one grouped query per corpus.

**4. `WriteResult` `__post_init__` anti-pattern** — Fixed with `field(default_factory=list)`; the `type: ignore` suppressions are gone.

**7. `CorpusReferenceService` doesn't extend `BaseService`** — Fixed.

**8. `_get_analysis` COMPLETED-from-the-start** — Comment added documenting the two provenance paths (framework-via-Celery wrapper owns the lifecycle; the synchronous agent-tool path has no observable in-between state).

**5. `auto_create_doc_analyzers` rename** — Declining for now: three historical data migrations (`analyzer/0009`, `0012`, `0013`) import the function lazily *at migrate time*, so a rename breaks `migrate` on fresh databases unless we edit applied migrations or keep an alias (dead code per project rules). The name is stale but the churn/risk isn't worth it inside this PR; happy to do it as a follow-up that also touches the migrations deliberately.

**9. Import-direction comment** — Already present since `17469c06e`: `annotations/models.py:1881` carries "`enrichment/constants.py` is pure — no model imports — so this cannot cycle."

Verification: 188 tests green across the enrichment/analyzer/governance-graph/architecture suites (`-n 0`), pre-commit incl. mypy green.

Issue #1682, opened 2026-05-17
Reconcile CorpusObjsService corpus-as-gate vs GraphQL MIN-permission semantic


JSv4:
`CorpusObjsService.get_corpus_documents` implements a **corpus-as-gate** semantic: if the user has corpus READ, all documents in the corpus are returned. This matches the canonical `DocumentFolderService.get_corpus_documents` behaviour pre-split.

However, `config/graphql/corpus_types.py:218` (`resolve_caml_articles`) deliberately implements **MIN(corpus_perm, doc_perm)** (per its docstring and the documented model in `CLAUDE.md`'s Permission System section):

> Effective Permission = MIN(document_permission, corpus_permission).
> A private document in a public corpus is still hidden from users
> without document-level access.

These are different semantics, and one of them needs to become the documented default.

**Options**:
- **A.** Make the service's `get_corpus_documents` follow MIN. Stronger by default. **Risk**: changes behaviour for current callers — documents in a public corpus that are individually private would disappear from MCP, discovery, etc.
- **B.** Keep corpus-as-gate as default; add a separate `CorpusObjsService.get_corpus_documents_visible_to_user` method for MIN-semantic callers (the GraphQL CAML resolver migrates to this).
- **C.** Add an `enforce_document_visibility: bool = False` kwarg.

**Recommendation**: Option B — explicit separate method avoids hidden semantic flips and lets each caller declare intent.

**Acceptance**:
- Decision documented in `CLAUDE.md`'s Permission System section.
- The GraphQL CAML resolver and any other MIN-semantic callers migrate to the chosen service method.
- `corpus_types.py:resolve_caml_articles` no longer emits `DeprecationWarning` (depends on the related CAML-aware-method issue too).

Issue #1655, opened 2026-05-15
Phase A — Per-model user_can() overrides + deprecate user_has_permission_for_obj (follow-up to #1637)


JSv4:
## Context

This is **Phase A** of the multi-phase permission-centralization plan tracked in PR #1637. Step 0 (FolderService vertical slice) landed in #1637; the API surface `_default_user_can` / `Manager.user_can` / `obj.user_can` already exists on `BaseVisibilityManager`, `PermissionedTreeQuerySet`, and `BaseOCModel`. Phase A pushes that API out to every visibility-managed model and demotes the parallel `user_has_permission_for_obj` helper to a deprecation shim.

The drift this is fixing: today `visible_to_user(u).filter(pk=x).exists()` and `user_has_permission_for_obj(u, x, READ)` can disagree (the helper intentionally ignores creator status — see docstring on `permissioning.py:505`). After Phase A there is one decision point per model, on the manager, alongside `visible_to_user`.

### 1. Per-model `user_can` overrides

Add `user_can` overrides matching each existing `visible_to_user` override. Each override must produce the same boolean for READ that the queryset filter would include, and must be defined in the same module as its sibling.

- [ ] `DocumentManager.user_can` — `opencontractserver/shared/Managers.py` — mirror `DocumentQuerySet.visible_to_user` (`shared/QuerySets.py:201`), including the public-corpus auto-inheritance branch.
- [ ] `AnnotationManager.user_can` — `opencontractserver/annotations/models.py` — order: superuser-bypass → structural-write-deny → privacy via `created_by_analysis`/`created_by_extract` → `MIN(doc, corpus)`. Move the body of the annotation branch from `permissioning.py:587-650` (the two recursive `user_has_permission_for_obj` calls become `Analysis.objects.user_can(...)` / `Extract.objects.user_can(...)`).
- [ ] `RelationshipManager.user_can` — `opencontractserver/annotations/models.py` — create the manager (currently uses base) and pin structural-write-deny for non-superusers.
- [ ] `ConversationManager.user_can` — `opencontractserver/conversations/models.py` — CHAT/THREAD bifurcation (CHAT visible only to its creator; THREAD visible per corpus READ).
- [ ] `ChatMessageManager.user_can` — `opencontractserver/conversations/models.py` — moderator access path.
- [ ] `NoteManager.user_can` — `MIN(doc, corpus)`.
- [ ] `PermissionManager.user_can` — see **Gotcha** below.
[...]

Issue #1450, opened 2026-05-02
Identify and consolidate inline access checks that bypass visible_to_user / user_has_permission_for_obj


JSv4:
The codebase has standardized on `Model.objects.visible_to_user(user)` (manager method) and `user_has_permission_for_obj` (utility) for access checks. CLAUDE.md notes that the older `resolve_oc_model_queryset` helper is deprecated.

Worth a sweep for the patterns that drift from this:

- Resolvers/mutations that hand-roll permission checks inline (e.g. `if obj.creator == user or user.is_superuser`) instead of calling the centralized helpers — these are where logic falls behind as the permission model evolves (sharing, corpus inheritance, MIN(doc, corpus)).
- Any remaining call sites of `resolve_oc_model_queryset`.
- Case-sensitivity bugs in email-based sharing comparisons (one side lowercases, the other doesn't).
- `Model.objects.get(pk=...)` / `Model.objects.filter(...)` calls in resolvers/mutations that ought to go through the permissioned manager or optimizer (cross-references #1449).

### Proposal

1. Produce an inventory in this issue grouped by module:
    - inline ad-hoc permission checks
    - direct `.objects.get` / `.objects.filter` in resolver/mutation paths
    - residual `resolve_oc_model_queryset` callers
    - email-comparison case mismatches
2. Convert each cluster into a focused PR that swaps to the centralized helper or pushes filtering down into the manager/optimizer (per #1449).
3. Flag any behavior changes uncovered during conversion separately so they get individual review (a "consolidation" PR should not also be a behavior-change PR).

### Acceptance

- Inventory checklist completed in this issue.
- Per-cluster PRs merged.
- `resolve_oc_model_queryset` removed once the last caller is gone.

JSv4:
## Inventory pass — round 1

Did a full sweep of the four categories called out in the issue. Summary first, details below.

| Category | Result |
|---|---|
| A. Inline ad-hoc permission checks | **14 instances** across 12 files — list below |
| B. Direct `.objects.get` / `.objects.filter` in resolvers/mutations | Mostly compliant; the remaining cases are either gated by `visible_to_user()`, `creator=user` filter, or `can_moderate()`. No new offenders found. |
| C. Residual `resolve_oc_model_queryset` callers | **None.** The function has no live callers — only a comment in `opencontractserver/shared/Managers.py:98` and a docstring reference in `opencontractserver/tests/test_visibility_managers.py:6`. Safe to remove. |
[...]

Issue #820, opened 2026-01-24
Agent Tools Not Fault Tolerant


JSv4:
If a tool fails mid-conversation, we don't have retry logic or graceful handling: 

```
django-1             | 2026-01-24 05:11:17,603 ERROR opencontractserver.llms.agents.pydantic_ai_agents [pydantic_ai_agents.py:480] [DIAGNOSTIC] Processing function_tool_call event
django-1             | 2026-01-24 05:11:17,603 ERROR opencontractserver.llms.agents.pydantic_ai_agents [pydantic_ai_agents.py:553] [DIAGNOSTIC] Tool 'ask_document' does not require approval - emitting ThoughtEvent
django-1             | 2026-01-24 05:11:17,603 ERROR opencontractserver.llms.agents.pydantic_ai_agents [pydantic_ai_agents.py:568] [DIAGNOSTIC] Finished processing function_tool_call for 'ask_document' - continuing iteration
django-1             | 2026-01-24 05:11:17,625 INFO opencontractserver.utils.permissioning [permissioning.py:331] get_users_permissions_for_obj() - Starting check for google-oauth2|114688257717759010643 with model type corpus forpermission PermissionTypes.READ
django-1             | 2026-01-24 05:11:17,625 INFO opencontractserver.utils.permissioning [permissioning.py:337] get_users_permissions_for_obj - App name: corpuses
django-1             | 2026-01-24 05:11:17,625 INFO opencontractserver.utils.permissioning [permissioning.py:209] get_users_permissions_for_obj() - Starting check for google-oauth2|114688257717759010643 with model type corpus
django-1             | 2026-01-24 05:11:17,625 INFO opencontractserver.utils.permissioning [permissioning.py:214] get_users_permissions_for_obj - App name: corpuses
django-1             | 2026-01-24 05:11:17,625 INFO opencontractserver.utils.permissioning [permissioning.py:250] get_users_permissions_for_obj - this_user_perms: corpuses.CorpusUserObjectPermission.None
[...]

PR #1985, opened 2026-06-12
Close permissioning-guide audit gaps: relationship privacy, anonymous extracts, group grants

JSv4:
# Permissioning audit gap closure

A full audit of `docs/permissioning/consolidated_permissioning_guide.md` against the GraphQL and service layers surfaced substantive guide/code divergences. Four intent decisions were made with the maintainer, and this PR makes code and docs agree.

### 1. Relationship privacy recursion implemented (closes the #1655 Phase-C deferral)
`RelationshipManager.user_can` and `RelationshipManager.visible_to_user` now enforce `created_by_analysis` / `created_by_extract` source permissions exactly like annotations, via a shared module-level `_source_privacy_recursion_passes` (`opencontractserver/shared/Managers.py`). Previously a user with doc+corpus grants could see and write analysis-/extract-private relationships through the manager surfaces. The relationship **creator** keeps access (creator short-circuit, mirrored by `Q(creator=user)` in the queryset gate) so filter/check parity holds.

### 2. Extracts are never anonymous-visible
The guide's documented contract ("Extract - never") is now enforced everywhere: a new `ExtractManager` (`opencontractserver/extracts/models.py`) denies anonymous users on both `visible_to_user` and `user_can`, and `ExtractService.get_visible_extracts` / `check_extract_permission` lock down the service boundary. Previously the service exposed `is_public` extracts in public corpora to anonymous callers (drift copied from the analysis service, reachable through the un-gated `extracts` resolver).

### 3. Group-granted source permissions unlock private rows in lists
The `created_by_*` privacy gates consulted only the USER object-permission tables, so a viewer whose analysis/extract grant arrived via a Django group passed `user_can` (groups resolve by default) but never saw the rows in lists - a filter/check parity drift. New shared builders in `opencontractserver/utils/source_visibility.py` join the group tables and replace four divergent inline copies (`AnnotationQuerySet.visible_to_user`, `AnnotationService.get_document_annotations` / `get_corpus_annotations`, `RelationshipService.get_document_relationships`). **Bonus fix found during consolidation:** `get_corpus_annotations` skipped privacy exclusion entirely for anonymous users - anonymous viewers of a public corpus could see analysis-/extract-private annotations on public documents. The exclusion now applies unconditionally.

### 4. Thread moderation: code is right, guide was wrong
`Conversation.can_moderate` does NOT grant moderation via EDIT permission on the corpus/document; the guide claimed it did. Guide corrected; the deliberate `Conversation.can_moderate` vs `Corpus.user_can_moderate` divergence (#1450) is now documented.

## Also in this PR

- **Annotation row-creator source-privacy parity**: `AnnotationManager.user_can` now mirrors `AnnotationQuerySet.visible_to_user`'s `Q(creator=user)` source-privacy exemption. This fixes the #1986 item-1 edge where a row creator could see their own analysis-/extract-rooted annotation in lists but fail `user_can(READ)`. The exemption only skips source recursion; document/corpus permissions still apply.
- **Geographic pin aggregation source privacy**: `GeographicAnnotationService.aggregate_for_corpus` now applies `apply_source_privacy_gate` after the visible-document corpus gate, so analysis-/extract-private annotations cannot leak through map pin aggregates. Regression coverage lives in `test_geographic_annotation_service.py`.
- **`myPermissions` mirrors the structural-write break-glass**: `AnnotationService` masked structural `_can_update`/`_can_delete` to `False` for everyone (under-reporting for superusers - a stale comment claimed superusers were "handled upstream", untrue since scoped admin 2026-05); `RelationshipService` had **no** structural mask (over-reporting doc+corpus writes for normal users). Both now mask to `user.is_superuser`, and tests now pin structural `_can_read=True` as well.
- **Extract manager integer-user parity**: `ExtractManager.visible_to_user` now resolves integer user ids before the anonymous guard, matching `user_can` and preserving the base manager's id shortcut.
- **Source-permission hot-path indexes**: the concrete Guardian tables already have `(user, permission, content_object)` / `(group, permission, content_object)` unique indexes; this PR adds concurrent companion indexes for the source privacy subqueries: `anl_uop_perm_user_obj_idx`, `anl_gop_perm_grp_obj_idx`, `ext_uop_perm_user_obj_idx`, and `ext_gop_perm_grp_obj_idx`.
- **Defensive extract permission lookup**: `ExtractService.check_extract_permission` now catches malformed ids (`TypeError` / `ValueError`) as denial, so an undecoded Relay id fails closed even though all production GraphQL callers already decode with `from_global_id(...)` before calling the service.
- **`analysis_id is not None` correctness**: `_source_privacy_recursion_passes` intentionally uses `is not None`, not truthiness, so a falsy-but-present id cannot skip the privacy recursion branch.
- **Review cleanup**: the stale `test_annotation_creator_parity_gap_sentinel` reference was removed from `source_visibility.py`; the comment now points at `test_annotation_creator_source_private_row_has_filter_check_parity`.
- **Guide overhaul**: every stale `*QueryOptimizer` name -> current per-app service; dead `<app>/query_optimizer.py` paths -> `<app>/services/...`; resolver locations after the `queries.py` split; `ws/agent-chat/` URL; awaited `_check_user_permissions` + issue #820 fault-tolerance split; real agent-test paths; `get_users_permissions_for_obj` group default (`True`); Pitfall 1 (structural-only, not empty); CorpusCategory GraphQL mutations; structural Enforcement Locations -> `Managers.py`. Newly documented: the **#1682 `get_corpus_documents` (corpus-as-gate) vs `get_corpus_documents_visible_to_user` (MIN)** dual semantics, `opencontracts.E001` dual enforcement + scan-scope nuance, `BaseService.filter_visible_qs`, and reference-table rows for Note / UserFeedback / Embedding / CorpusAction / Notification.
- **Stale superuser-bypass docstrings removed** in `shared/QuerySets.py`, `shared/Managers.py`, `conversations/models.py`, `llms/tools/image_tools.py`, `documents/services/relationships.py`; CLAUDE.md pitfall 2 -> `ConversationService`.
- **Caller audit**: `get_label_distribution_for_corpus` has one production caller (`config/graphql/corpus_queries.py`), and it passes `user=info.context.user`; the other call sites are tests.
- **Changelog fragments**: `changelog.d/perm-audit-2026-06.security.md` + `.fixed.md`.

## Tests

- `test_authorization_invariants.py`: relationship matrix extended with analysis-/extract-rooted fixtures + user- and group-granted readers; the `test_no_privacy_widening_via_created_by_analysis` sentinel (which pinned the old deferral) is **replaced** by six real privacy invariants per the agreed behavior change; new `ExtractAuthorizationInvariantsTestCase`; annotation group-grant parity test.
- New / updated: `test_relationship_privacy_scoping.py`, `test_extract_anonymous_lockdown.py`, `test_source_visibility_group_grants.py`, `test_structural_mypermissions_breakglass.py`, `test_privacy_gate_regressions.py`, `test_analysis_listing_grant_parity.py`, `test_geographic_annotation_service.py`, and the annotation creator parity coverage in `test_authorization_invariants_coverage.py`.
- Local validation for the latest follow-up commits:
  - `git diff --check`
  - `python -m compileall ...` on touched Python modules, tests, and migrations
  - `docker compose -f test.yml run --rm django pytest --create-db opencontractserver/tests/permissioning/test_authorization_invariants_coverage.py::AnnotationUserCanLeafBranchesTestCase opencontractserver/tests/permissioning/test_extract_anonymous_lockdown.py -q` -> 24 passed
  - `docker compose -f test.yml run --rm django pytest opencontractserver/tests/permissioning/test_authorization_invariants.py::AnnotationAuthorizationInvariantsTestCase opencontractserver/tests/permissioning/test_source_visibility_group_grants.py opencontractserver/tests/permissioning/test_privacy_gate_regressions.py -q` -> 29 passed
  - `docker compose -f test.yml run --rm django pytest opencontractserver/tests/permissioning/test_relationship_privacy_scoping.py opencontractserver/tests/permissioning/test_structural_mypermissions_breakglass.py opencontractserver/tests/permissioning/test_extract_anonymous_lockdown.py opencontractserver/tests/test_geographic_annotation_service.py -q` -> 40 passed
  - `docker compose -f test.yml run --rm django isort --check-only ...` on touched files
  - `docker compose -f test.yml run --rm django flake8 ...` on touched files
  - Live test DB index check confirmed all four new hot-path indexes exist.

## Follow-up observations (out of scope, flagged during audit)

- `ExtractType` has no `get_queryset` override; related-field traversal (e.g. datacell -> extract) relies on parent gating. Pre-existing.
- Mention-autocomplete resolvers still inline guardian `get_objects_for_user` in `config/graphql/search_queries.py` (legal under E001's three-token scan; now documented in the guide as a known exception; service migration is a candidate follow-up).

---
_Generated by [Claude Code](https://claude.ai/code/session_01SqVgjvU9qNPyrf3Vjb9zbM)_


JSv4:
Addressed the review in `7b1069b`:

- **🔴 `require_permission` example** — the example's *logic* was already correct, and the two suggested rewrites would not be: `require_permission` returns `""` (falsy) on **grant** and a denial string (truthy) on **deny** — it never raises — so `if require_permission(...): bail` is the documented house idiom (see the "Truthiness inversion" section of `BaseService.require_permission`'s docstring), and `if not require_permission(...)` would flip the gate. The real problem was my compressed `(falsy = granted)` comment; the example now spells out the inversion explicitly and points readers at `BaseService.user_has` when they want a boolean grant.
- **🟡 extract-positive pin** — added `test_privacy_recursion_grants_via_extract_grant`: explicit extract READ unlocks the extract-rooted relationship on both surfaces and does not bleed into the analysis-rooted sibling.
- **🟡 PERF note** — added at the `RelationshipManager.user_can` call site (the `AnnotationManager.user_can` docstring already carries the full version of this note).
- **🟡 `ExtractManager` kwargs** — verified non-issue: `BaseVisibilityManager.visible_to_user(self, user=None, lightweight=False, with_doc_label_annotations=False)` declares both kwargs.
- **Annotation-creator parity asymmetry** — agreed it's now conspicuous with the relationship side pinned; it's listed in the PR's follow-up section and intentionally left out of scope here (changing it is an annotation-side behavior change that deserves its own invariant work).

---

JSv4:
Dispositions for the latest review (addressed in `eec7f04` and `7b189db`):

1. **Any-permission grant shape** — upgraded from "document the intent" to a real fix: this was a live filter/check parity violation, not a defensible design choice. `user_can`'s recursion resolves READ via the read codename, so an `update_analysis`-only grantee cleared the old list gates while failing `user_can(READ)`. The builders now filter `permission__codename="read_analysis"` / `"read_extract"` on both the user- and group-level tables; pinned by `test_update_only_source_grant_does_not_unlock_lists`.
2. **`ExtractManager` None-handling asymmetry** — added an equivalence comment; the invariant suite already pins the surfaces agreeing.
3. **Deleted-source posture** — the review's premise doesn't hold: both `created_by_*` FKs are `on_delete=SET_NULL` (`annotations/models.py`), so deleting the source NULLs the FK and the row becomes a normal row — nobody is permanently locked out. The `source is None` branch fires only in the read-vs-SET_NULL race window and fails closed for that window by design. Now documented in the `_source_privacy_recursion_passes` docstring; no tombstone logic needed.
4. **Unconditional `privacy_gate` construction** — the builders are lazy querysets composed into the same single SQL statement for every caller class; there's no separate owner-only query to optimize. No change.
5. **N+1 note** — already addressed in the previous round (`7b1069b`): PERF comment at the relationship call site; the annotation manager docstring carries the full version.
6. **`is_public` extract gap** — this review raced the push: fixed in `eec7f04` exactly as described (authenticated branch of `visible_extracts_for` now includes `is_public`, pinned by `test_public_extract_source_passes_both_surfaces`). Anonymous remains fully denied via `ExtractManager`.
7. **`get_corpus_annotations` call-site comment** — added: the exclusion is intentionally outside any auth guard; the old anonymous guard was itself the leak.

Docs note on `require_permission` was handled in `7b1069b` (it returns `""`/denial-string and never raises — the example now states this explicitly).

---

JSv4:
Dispositions for the deep review (addressed in `bdef2f2`; every claim verified against code before acting):

**🔴 1 — `codename__contains="read"`**: real and fixed, with one correction: it's pre-existing code, not new (this PR had only touched the anonymous branches of that file). Not exploitable today — no corpus codename besides `read_corpus` contains "read" — but tightened to exact matches anyway.
**🔴 2 — missing group-level corpus grants in `get_visible_extracts`**: confirmed and fixed — plus the twin gap the review missed: the *extract* leg also consulted only user-level rows (any codename). Both legs now join the group tables and match `read_extract`/`read_corpus` exactly, with `ExtractListingGrantParityTestCase` pinning both directions (group grants unlock; update-only extract grant does not).
**🟠 3 — broad `except (ImportError, Exception)`**: confirmed pre-existing in `BaseVisibilityManager`, but the severity claim is inverted — the fallback drops guardian rows, so it *under*-discloses rather than over-discloses. Tightening the exception envelope on load-bearing base-manager code mid-PR risks surfacing unrelated latent errors; deferred to follow-ups.
**🟠 4 — aggregate privacy gates**: confirmed for both. `get_label_distribution_for_corpus` (user-facing via `corpusIntelligenceAggregates`) and `get_relationship_summary` now apply the `created_by_*` gate; regression pins in the new `test_privacy_gate_regressions.py`.
**🟠 5 — `get_corpus_relationships` corpus-as-gate**: the conditional resolves to "keep": its sole caller is `mcp/tools.py`, and per issue #1682 corpus-as-gate is the documented default for pipeline-facing callers. Documented inline with an explicit warning that any future GraphQL caller must switch to the `_visible_to_user` variant. It *was* missing the privacy gate though — added and pinned.
**🟠 6 — anonymous ChatMessage THREAD restriction**: confirmed real (ChatMessage's anonymous branch lacks the THREAD filter its Conversation counterpart has) but pre-existing and outside this PR's audited scope — added to follow-ups alongside the conversations group-grant gaps the review itself files as pre-existing.
**🟠 7 — `Datacell._validate_manual_entry`**: confirmed real (`validation_config.get` before the `or {}` guard) but unrelated to permissioning; follow-up.
**🟡 8 — `analysis_id == 0` sentinel**: pre-existing API contract with the frontend's `"__none__"` mapping; renaming it touches the GraphQL contract — follow-up.
**🟡 9 — single-filter + `.distinct()`**: the single-filter form was already adopted in `bdef2f2`'s predecessor; note the premise is off — queryset `|` OR-merges the WHERE clause (it's `.union()` that emits SQL UNION), and `.distinct()` is unnecessary because all predicates are single-table (no join fan-out).
**🟡 10 — "Superuser bypass" docstring**: fixed — now phrased as the structural-write break-glass with an explicit "NOT a blanket bypass" note.
**Test gaps**: group→extract counterparts added in both files. The weak assertions in `test_query_optimizer_methods.py` are pre-existing tests outside this change's behavior — left untouched per repo convention (noted as follow-up).

---

JSv4:
Re the latest review round — no code changes warranted, two requests fulfilled:

- **Follow-up tracking**: all out-of-scope items from the review rounds (annotation-creator parity edge, ChatMessage anonymous THREAD asymmetry, conversations group grants, the broad `except` fallback, `ExtractType.get_queryset`, `Datacell` validator, sentinel/API nits, and the `claude-review` workflow malfunction) are now durably tracked in #1986 rather than living only in the PR description.
- **`get_label_distribution_for_corpus` callers**: grep confirms exactly two call sites — the `corpusIntelligenceAggregates` resolver (passes `info.context.user`) and the new regression tests (pass `user=`). Nothing regresses to the anonymous fallback.
- **`user.groups` "evaluated twice"**: declined as written — the builders return lazy querysets, so both group-id subselects compile into the *single* SQL statement of the enclosing query; there are no extra round-trips, and `auth_user_groups` is a small indexed table. Threading a `user_group_ids=` parameter through a security-critical module for an unmeasured micro-win isn't worth the API surface; tracked as item 11 in #1986 should profiling ever disagree.
- **Docstring reshuffling nits**: leaving as-is deliberately — for security-critical helpers the consolidated docstring is the audit surface; scattering the deleted-source and performance notes across branches optimizes for casual reading at the cost of the next auditor.

---

JSv4:
Both mediums addressed in `f614ed9`:

- **Builder shape inconsistency** — real oversight from the earlier conversion (only the analyses builder was updated): `visible_extracts_for` now uses the identical single-filter OR shape. One precision note: Django's queryset `|` OR-merges the WHERE clause for same-model unannotated filters — it never emits SQL `UNION` (that's `.union()`) — so the change is shape-consistency, not a query-plan fix; behavior and SQL are equivalent either way.
- **`user=None` regression risk** — docstring now states explicitly that `None` is the most-restrictive anonymous shape (omission under-counts for authenticated viewers; it can never leak) and names both current call sites, which remain the only ones (re-grepped).
- **Corpus-as-gate caveat** — now names the sole caller (`opencontractserver/mcp/tools.py`) so the #1682 constraint is searchable.

The two observations (annotation-creator parity, `ExtractManager` double-resolution coupling) are already tracked as items 1 and (inline-documented) in #1986.

---

JSv4:
Round addressed in `e4a0fbe`:

- **Item 1 (`user=None` default)** — went further than the warn suggestion: `user` is now **required and keyword-only** on `get_label_distribution_for_corpus`. Omission is a `TypeError` at the call site rather than a silent under-count; an explicit `None`/`AnonymousUser` still resolves to the most-restrictive anonymous shape inside the builders (defense in depth).
- **Item 4 (exclusion repetition)** — adopted: the six call-site copies are consolidated as `apply_source_privacy_gate(qs, user)` in `source_visibility.py`, with a docstring note that `AnnotationQuerySet.visible_to_user` deliberately keeps its positive-Q composition (it needs the creator disjunct and single-WHERE composition with the doc/corpus EXISTS predicates) and must stay in sync.
- **Item 2 (double resolution)** — declined as proposed, with the inverse-fragility rationale: forwarding the **resolved** `User` keeps `ExtractManager.user_can` correct even if a future refactor removes resolution from the base class, whereas forwarding the original raw input (possibly an int/str id) would then break. The idempotence is documented at the call site; the invariant suite pins the surfaces.
- **Item 3 (type annotation)** — declined for file consistency: `relationship_service.py` leaves `user` untyped on every method; annotating one signature would diverge from the module's convention. Worth a file-wide pass under the typing roadmap rather than a one-off here.
- **Verification request (datacell → extract traversal)** — already tracked: #1986 item 5.

---

JSv4:
Latest round addressed in `0a7ea9c`:

- **🔴 1 (fragile implicit contract)** — taken as a documentation fix, per the review's own "not a bug" assessment: the structural-READ short-circuit in `_source_privacy_recursion_passes` now names both list-side counterparts (the queryset's `Q(structural=True)` disjunct and the gate's `structural=False` conjuncts) and states that removing either side alone breaks parity for structural rows rooted in invisible sources.
- **🔴 2 (parity test)** — adopted with a necessary scoping correction: the recommendation as written (`exists() == exists()` for *each* fixture) would fail by design on creator-owned rows, because the queryset gate's `Q(creator=user)` disjunct is precisely the annotation-side asymmetry tracked as #1986 item 1. `test_gate_matches_queryset_visibility_for_non_creator` pins the contract for non-creator viewers (where the shapes must agree exactly), before and after a source grant, and documents the creator exclusion. Verified green locally.
- **🟡 3** — wording upgraded: loop-callers `MUST select_related(...)`; single-object call sites noted as unaffected. (One factual note: `RelationshipManager.visible_to_user` uses the builder subqueries directly — it never routes through `_source_privacy_recursion_passes`, so the bulk path was already clean, as the review half-observes.)
- **🟡 4** — declining a third comment on the resolution idempotence; the call site already documents the assumption and the invariant suite pins both surfaces. Tracked context lives in #1986.
- **🟡 5** — stale: `test_public_extract_source_passes_both_surfaces` has existed (by exactly that name) since `eec7f04` — the `is_public` branch is covered today, not zero-coverage.
- **Nit** — taken: `Q` lifted to module level; only the model imports remain deferred.

---

JSv4:
Latest round addressed in `66d267f`:

- **Creator-parity tracking** — the issue has existed since earlier in this PR (#1986, item 1); what was genuinely missing was the in-code pointer, now added at the `Q(creator=user)` disjunct in `AnnotationQuerySet.visible_to_user`'s privacy gate.
- **`get_label_distribution_for_corpus` callers** — confirmed (third grep, same result): exactly two call sites, the `corpusIntelligenceAggregates` resolver and the privacy regression tests, both passing `user` by keyword. No management commands, tasks, or utilities call it.
- **`documentrelationship` in the mixin** — confirmed in the real code, not just the guide: `config/graphql/permissioning/permission_annotator/mixins.py:200`. The guide edit was made to *match* the code during the audit, not the other way around.
- **Codename string literals** — declined for this PR with evidence: no codename constants exist anywhere (`opencontractserver/constants/permissioning.py` holds only cache attrs), and bare literals / `f"read_{model_name}"` shapes are the prevailing convention across `shared/QuerySets.py` and the services. Introducing constants for only the new call sites would diverge from that convention inside a security PR; added as #1986 item 12 for a repo-wide pass.
- **Guide snippet `annotation_pk`** — fixed (now derived from `annotation_id` via `from_global_id`).
- **N+1 inline-comment suggestion** — declined as a third copy: the MUST-strength note lives in the function docstring and at the `RelationshipManager.user_can` call site already.
- The `ExtractManager` docstring-note suggestion is already present at the call site (added two rounds ago).

---

JSv4:
Latest round addressed in `4198242`:

- **Item 1 (creator parity gap)** — the follow-up issue has existed for several rounds (#1986 item 1, plus the in-code pointer at the queryset gate); the genuinely new suggestion was the **sentinel test**, now added: `test_annotation_creator_parity_gap_sentinel` pins the current divergence (creator-of-row visible in lists, denied by `user_can(READ)`) with both assertions referencing the issue — same idiom as the Phase-C deferral sentinel this PR retired. It also documents why the invariant matrix excludes that fixture shape. Verified locally (coverage file 51/51).
- **Item 5 (caller grep)** — answered for the fourth time, same result: the resolver and the regression tests are the only call sites; both pass `user` by keyword. The grep result is also recorded in the method's docstring.
- **Item 7 (queryset-path parity test "missing")** — factually stale: `test_gate_matches_queryset_visibility_for_non_creator` (added several rounds ago) compares `apply_source_privacy_gate` membership against `Annotation.objects.visible_to_user` membership row-by-row, pre- and post-grant — that *is* the queryset path.
- **Item 6** — previously declined twice with rationale (forwarding the resolved User is the robust direction); the reviewer's own assessment this round is "not a bug".

With this, every distinct suggestion across all twelve review rounds has been either implemented, pinned by a test, tracked in #1986, or declined with evidence on this thread.

---

JSv4:
No changes needed for this round — all three tracked items predate it:

1. **CI-legible fixture for the creator asymmetry** — landed one commit before this review posted: `test_annotation_creator_parity_gap_sentinel` (`4198242`) pins the divergence at CI time with both assertions referencing #1986 item 1, exactly as suggested.
2. **"No `totalExtracts: 0` assertion for anonymous"** — factually incorrect against the diff: `test_stats.py:145-153` contains the dedicated anonymous block (`self.assertEqual(stats, {**expected_stats, "totalExtracts": 0})`, with the BEHAVIOR CHANGE comment) — added in `3b7df6f` when the anonymous case was split out of the shared loop. The review appears to have quoted the loop without the block that follows it.
3. **N+1 loop-caller enforcement** — there are no loop callers of `user_can` over privacy-rooted rows today (all call sites are single-object checks), so a fixture verifying "known loop callers prefetch" would assert over an empty set. The MUST-strength requirement is documented at the function and at the `RelationshipManager` call site; adjacent concern is already #1986 item 11.

Checklist status: CI was fully green on the `bdef2f2` line (full-suite pytest 8,598 passing + codecov patch checks); the subsequent commits are locally verified (permissioning directory, invariants, gate consumers, coverage file 51/51) and the current tip's run is in flight. #1986 item 1 exists and is now triply referenced (issue, in-code pointer, sentinel test).

---

JSv4:
Round addressed in `344d608`:

- **Deleted-source semantics in the guide** — adopted, the round's one genuinely new finding: the Source Privacy banner now states the `on_delete=SET_NULL` consequence (deleting a private Analysis/Extract un-gates its rows into ordinary doc+corpus-governed rows; delete or re-scope the rows first if they must stay private). One precision note: the rows don't become "visible to everyone" — doc+corpus permissions still apply — but the operator-surprise point stands and is now documented where operators look.
- **`get_visible_extracts` None-guard asymmetry** — adopted: both `user.is_anonymous` reads (lockdown branch and corpus-scope branch) now use the same fail-closed `user is None or getattr(...)` shape as `check_extract_permission` and the manager. (A bare read would have failed *loud*, not open, on `None` — but consistency is right.)
- **`DocumentRelationshipService.get_visible_relationships` myPermissions under-report** — verified real (creator-only `Case` at `documents/services/relationships.py:263-280` vs MIN-based `user_has_permission`) and pre-existing in a method this PR never touched; now tracked as #1986 item 6. Same mirror principle as this PR's structural-mask fix, but fail-safe in direction and out of scope here.
- Items 2 and 4 were already resolved (sentinel `4198242`; documented idempotence).

---

JSv4:
Round addressed in `902a3fb` (second consecutive approval — thank you):

- **Pre-merge verification (loop callers of `RelationshipManager.user_can`)** — verified clean: there are none. `AnnotatePermissionsForReadMixin`'s fallback never calls `user_can` (it resolves per-row `myPermissions` from prefetched guardian metadata — zero `user_can` references in `mixins.py`); every GraphQL relationship list flows through the annotated services (`_can_*` pre-computed, no per-row checks); and all ten relationship permission gates in the mutation files are single-object. The privacy recursion's per-call FK dereference therefore has no N+1 surface today, and the MUST-prefetch note guards future loop callers.
- **MCP boundary reinforcement** — adopted: the `get_corpus_relationships` call in `opencontractserver/mcp/tools.py` now carries its own #1682 context note (sanctioned corpus-as-gate caller; do not copy into GraphQL), so the warning exists on both sides of the boundary.
- **Follow-up items** — both already tracked: the annotation-creator gap is #1986 item 1 (sentinel-pinned at CI time), and the mention-resolver E001 exception is #1986 item 10.

---

JSv4:
Round addressed in `37d8f3c`:

- **Dead-code cleanup** — adopted, with a sharper fix than suggested: the anonymous guard in `get_visible_extracts` now **early-returns** `Extract.objects.none()` (the fall-through `qs = none()` shape was what made the corpus-block's inner anonymous check redundant-in-effect — it gated an already-empty queryset). The inner branch is removed and the corpus permission check is documented as authenticated-only.
- **Aggregate grant matrix** — adopted: `LabelDistributionPrivacyRegressionTestCase` gains the group-granted case, completing anonymous / ungranted / owner / group-granted coverage for the surface.
- **Guide banner cross-link** — adopted: the deleted-source note now points to `_source_privacy_recursion_passes` for the full mechanics including the race-window guard.
- **"Stronger fix" for the two-shape sync** — premise correction: `AnnotationQuerySet.visible_to_user` has consumed the shared `visible_analyses_for` / `visible_extracts_for` builders since the first group-grant commit (see `shared/QuerySets.py` — the queryset imports from `source_visibility`). What differs is only the gate *composition* (positive-Q with the creator disjunct vs the exclude pair), which cannot be unified: the queryset shape needs the creator disjunct (#1986 item 1) and single-WHERE composition with the doc/corpus EXISTS predicates. That residual difference is exactly what `test_gate_matches_queryset_visibility_for_non_creator` pins.
- **Double-resolution comment** — already present at the call site since two rounds ago ("idempotent no-op for the User instance").

---

JSv4:
Round 17 addressed in `f8a2b6e` — its item 1 was the most substantive finding in many rounds, and it was right:

- **Creator-disjunct gap in `apply_source_privacy_gate`** — confirmed real and fixed. The deeper framing: this PR's own relationship manager changes created a 2-vs-1 surface disagreement (both manager surfaces honoured a creator who lost source access; the service listings hid the row). The gate now carries the same **authenticated-only** creator exemption as the queryset disjunct — anonymous callers get an always-false predicate, since `Q(creator=AnonymousUser())` is not a valid lookup and anonymous users can never be a row's creator (the safety caveat in the review was on the right track but the mechanism differs: Django rejects the unsaved-instance lookup rather than null-matching it). After the fix: relationships agree across all three surfaces; annotations agree across both list surfaces, with `user_can` remaining the single tracked divergence (#1986 item 1, sentinel-pinned). The missing test the review named exists now for both models.
- **Caller greps** — `get_label_distribution_for_corpus`: two call sites, both pass `user` (fourth grep, same answer). `get_relationship_summary`: its `user` parameter predates this PR (positional, unchanged signature) — sole caller is `document_types.py:1201`, which has always passed it; no API break occurred there.
- **Double-resolution pin** — the requested tests already exist: `test_manager_user_can_denies_anonymous` covers both `user_can(None, …)` and `user_can(AnonymousUser(), …)`, and the parity test covers the surfaces agreeing.
- **`require_permission` docstring** — the inversion warning already lives there: the guide's `TRUTHINESS INVERSION` comment was copied *from* `BaseService.require_permission`'s own "Truthiness inversion" docstring section, which includes the same do-not-transcribe warning.
- **Remaining nits** (gate-shape placement, NOT-IN profiling, N+1 CI grep, `is_public` dead-branch note) — placement is already adjacent to both consumers; the `is_public` branch is not untested (pinned by `test_public_extract_source_passes_both_surfaces`); profiling/CI-grep enforcement are #1986 territory (items 11/12 adjacent).

Full `tests/permissioning/` suite running against the change locally; CI runs the same head.

---

JSv4:
Round 18 (third consecutive approval) — one nit adopted in `68679d2`, the rest closes with cites:

- **`is not None` idiom** — adopted (the truthiness checks were moved verbatim from the original annotation implementation; behavior-identical for positive PKs, but the explicit form is right for FK id fields).
- **Sentinel present?** — yes: `test_annotation_creator_parity_gap_sentinel` landed in `4198242` and is on the branch.
- **Anonymous label-distribution test** — already exists by intent and name: `test_private_labels_hidden_from_anonymous` in `test_privacy_gate_regressions.py` passes `AnonymousUser()` (not `None`) and asserts the private label is absent — present since the file was created.
- **Double-application scan** — confirmed clean: six call sites, each applies `apply_source_privacy_gate` exactly once (3 in `annotation_service.py`, 3 in `relationship_service.py`); the summary builds its own queryset rather than layering on `get_document_relationships`.
- **`resolved is None` comment** — the call-site comment already explains it ("the resolver maps None / unresolvable ids to None and passes AnonymousUser through, so both surfaces deny exactly the same caller set").
- **`# TODO(#1682)` promotion** — declined: the corpus-as-gate scope is a permanent semantic boundary, not pending work; tagging it TODO would misrepresent it as unfinished, and the `#1682` reference is already greppable on both sides of the boundary (service docstring + MCP call site).
- **Release communication for the anonymous-extract change** — the security changelog fragment carries the user-visible consequence prominently; per the repo's release process those fragments ARE the release-notes feeder (collated into `CHANGELOG.md` at release), so the note will surface exactly where release readers look.
- **N+1 enforcement mechanism** — declined as in prior rounds (no loop callers exist; documented MUST at the function and call site; #1986-adjacent if one ever appears).

---

JSv4:
Round 19 — one test adopted in `f4374fd` (`check_extract_permission(None, …)` now pinned at the service layer; the manager `None` cases were already covered). Everything else closes with cites:

1/4. **Double-resolution & call-site N+1 notes** — both comments already exist at the call sites (rounds 10–14); `resolve_user_for_user_can` is a pure resolver and stamps nothing.
2. **Gate re-evaluation note** — the gate is applied exactly once at each of its six call sites (verified by grep last round) and the builders embed as subqueries in a single statement; standard queryset laziness applies.
3. **`Q(pk__in=[])` comment** — the requested comment exists verbatim at that line: "Always-false predicate: anonymous callers get no exemption…".
5. **Comment density** — the quoted CLAUDE.md rule ("never write multi-paragraph docstrings… one short line max") does not appear in CLAUDE.md; the repo's actual house style for security-critical permission code is dense audit-trail docstrings (`Managers.py`, `QuerySets.py`, `BaseService` are the precedents these new docstrings follow).
6. **`is_public` cross-reference** — `visible_extracts_for`'s docstring already references `ExtractManager` twice, including the manager-parity rationale.
7. **Anonymous regression sentinel** — `test_anonymous_sees_only_non_private_rows` is the first test in `test_privacy_gate_regressions.py` and asserts exactly the requested case; it fails on pre-fix code by construction (the old guard skipped the exclusion entirely for anonymous viewers).
8. **Changelog coverage of the resolver `user=` change** — it is the headline of the security fragment's aggregate-surfaces bullet (label-name/count leak in `corpusIntelligenceAggregates` + the required keyword-only `user`), not an unflagged drive-by.

With three consecutive approvals and the deltas reduced to single-test/single-idiom items, treating the review loop as converged; the gate to merge is the routine CI pass on the tip.

---

JSv4:
Round 20 addressed in `1c88b6b` — its item 1 was a genuine catch, and better than "track as follow-up": **fixed in-PR**. `AnalysisService.get_visible_analyses` carried the identical drift its sibling `get_visible_extracts` had before round 6 (user-table-only guardian `Exists` with any codename on the analysis leg; `codename__contains="read"` substring on the corpus leg) — leaving it unfixed after this PR's gate changes would have produced exactly the wedge the review describes: a group-granted reader passing the privacy gates while unable to see the analysis row in the list view. Both legs now join the group tables and match `read_analysis` / `read_corpus` exactly, with `test_analysis_listing_grant_parity.py` pinning both directions (group grants unlock + single-object agreement; update-only grant does not unlock). Anonymous semantics deliberately unchanged (public analyses stay visible — the documented contract, unlike extracts). Changelog bullet extended to cover both siblings.

Remaining items were confirmations (2, 3, 4, 5, 7 — agreed on all counts) or already covered: item 6's release-notes ask is carried by the security fragment's `is_public` sentence (fragments are the release-notes feeder), and the `_NEVER_MATCH` constant nit is served by the existing inline always-false comment. Verification of the change running locally against the new test plus the legacy analysis-visibility suites; CI runs the same head.

---

JSv4:
Round 21 — no changes warranted; every item verifies against the branch:

- **Caller grep** (5th request): two call sites, both pass `user=` — also stated in the function's docstring itself since round 8.
- **`get_relationship_summary` coverage**: explicit — `RelationshipAggregatePrivacyRegressionTestCase.test_summary_excludes_private_rows_for_ungranted_viewer` / `..._includes_private_rows_for_owner` in `test_privacy_gate_regressions.py`.
- **`filter_visible_qs` fail-closed**: confirmed in source — `BaseService.filter_visible_qs` raises `TypeError` on inputs lacking `visible_to_user` (`shared/services/base.py`, the SECURITY-commented branch) rather than returning unfiltered rows.
- **`corpus.get_documents()` deprecation**: present on main before this PR — `DeprecationWarning` at `corpuses/models.py` (verified during the original audit; correctly absent from this diff).
- **Label-distribution "now runs on all paths" perf framing**: the gate previously ran on *no* path there — that absence was the leak being fixed; the subquery cost is the fix working. Staging plans worth watching as with any aggregate, agreed.
- **Creator-gap UX scheduling** and **CI as the merge gate**: both are maintainer-process items already tracked (#1986 item 1 with sentinel; CI cycling the tip).

With rounds 14–21 producing three approvals and otherwise only items that already exist on the branch by name, I'll respond to future automated rounds only if they contain new, verifiable findings — the disposition ledger above and #1986 cover everything else.

---

PR #2007, opened 2026-06-15
feat: add LiteLLM as AI gateway provider

RheagalFire:
## Summary

Add LiteLLM as a provider in `SimpleLLMClient`, enabling access to 100+ LLM providers (Anthropic, Azure, Bedrock, Vertex, Groq, Ollama, etc.) via the [LiteLLM](https://github.com/BerriAI/litellm) Python SDK.

The `Provider` enum already had `ANTHROPIC` and `GOOGLE` marked as "Future support". Rather than implementing each provider individually, LiteLLM provides a unified SDK that translates between all major LLM APIs, so adding it once covers all of them.

## Changes

- `opencontractserver/llms/client.py`
  - Added `LITELLM = "litellm"` to `Provider` enum
  - Added `_init_litellm()` for lazy litellm import and API key resolution (from settings or constructor arg)
  - Added `_chat_litellm()` using `litellm.completion()` with `drop_params=True` for cross-provider compatibility
  - Wired into `__init__` and `chat()` dispatch
- `requirements/base.txt` - added `litellm>=1.80.0,<1.87.0`
- `opencontractserver/tests/test_litellm_provider.py` - 7 unit tests covering enum, init, chat dispatch, API key handling, and factory function

## Tests

**Pre-commit:** all hooks pass (black, isort, flake8, mypy).

**Unit tests (7 tests):**
- `TestLiteLLMProviderEnum` - enum value, string parsing
- `TestLiteLLMClientInit` - init with API key, settings fallback, no key
- `TestLiteLLMChat` - completion dispatch with `drop_params=True`, API key omission
- `TestLiteLLMFactory` - `create_client()` with `LLM_CLIENT_PROVIDER="litellm"` setting

## Risk / Compatibility

- Additive only. OpenAI provider path untouched.
- `litellm` is added to `requirements/base.txt` (not optional, since the existing OpenAI dep is also in base).
- `drop_params=True` silently drops per-provider-unsupported kwargs, preventing errors when switching between providers.

## Example usage

```python
from opencontractserver.llms.client import SimpleLLMClient, ChatMessage

client = SimpleLLMClient(
    provider="litellm",
    model="anthropic/claude-sonnet-4-6",
    # api_key auto-resolved from settings.LITELLM_API_KEY if not provided
)

response = client.chat([
    ChatMessage(role="system", content="You are a legal document analyst."),
    ChatMessage(role="user", content="Summarize this clause."),
])
print(response.content)
```

Or via Django settings:
```python
# settings.py
LLM_CLIENT_PROVIDER = "litellm"
LLM_CLIENT_MODEL = "anthropic/claude-sonnet-4-6"
LITELLM_API_KEY = "sk-..."

# Then anywhere in code:
from opencontractserver.llms.client import create_client
client = create_client()
```

### DIFF
--- a/opencontractserver/llms/client.py
+++ b/opencontractserver/llms/client.py
@@ -19,6 +19,7 @@ class Provider(Enum):
     """Supported LLM providers."""
 
     OPENAI = "openai"
+    LITELLM = "litellm"
     ANTHROPIC = "anthropic"  # Future support
     GOOGLE = "google"  # Future support
 
@@ -78,6 +79,8 @@ class SimpleLLMClient:
         # Initialize provider-specific client
         if provider == Provider.OPENAI:
             self._init_openai(api_key)
+        elif provider == Provider.LITELLM:
+            self._init_litellm(api_key)
         else:
             raise ValueError(f"Provider {provider} not yet supported")
 
@@ -93,6 +96,15 @@ class SimpleLLMClient:
             raise ValueError("OpenAI API key not found in settings")
 
         self.client = OpenAI(api_key=api_key)
+
+    def _init_litellm(self, api_key: Optional[str] = None) -> None:
+        """Initialize LiteLLM client."""
+        try:
+            import litellm  # noqa: F401
+        except ImportError:
+            raise ImportError("LiteLLM library not installed. Run: pip install litellm")
+
+        self._litellm_api_key = api_key or getattr(settings, "LITELLM_API_KEY", None)
 
     def chat(
         self,
@@ -128,6 +140,8 @@ class SimpleLLMClient:
 
         if self.provider == Provider.OPENAI:
             return self._chat_openai(messages, model, temperature, max_tokens)
+        elif self.provider == Provider.LITELLM:
+            return self._chat_litellm(messages, model, temperature, max_tokens)
         else:
             raise ValueError(f"Provider {self.provider} not yet supported")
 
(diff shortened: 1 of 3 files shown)


RheagalFire:
cc @JSv4

JSv4:
Nice! Will take a look.

JSv4:
Thanks for the PR. I'd like to close this one as superseded, and if you're game would love a fresh PR rebuilt against the pattern below.

**1. `opencontractserver/llms/client.py` was deleted after this PR was opened.**

We removed `SimpleLLMClient` (the file this PR extends) in favor of routing all LLM calls through a registry (`opencontractserver/llms/llm_registry.py`) that resolves per-call → per-agent → per-corpus → install-default settings. `SimpleLLMClient` predated that. It was a standalone OpenAI-only wrapper. This PR no longer applies, and even resolving the conflict by keeping your version of the file would put back the standalone path we'd just consolidated out.

**2. The registry already has a clean extension point for this, and the `litellm` package isn't even needed.**

Providers are pluggable components under `opencontractserver/pipeline/llm_providers/` (see `openai_provider.py`, `ollama_provider.py` for the pattern). Each declares a `provider_key`, a `Settings` dataclass (`api_key` + `base_url`, both admin-editable live via System Settings, zero extra frontend code), and gets threaded into a `pydantic-ai` model by a provider-key branch inside `model_factory.py::_construct_model`, next to the existing `anthropic` and `google-gla` cases.

`pydantic-ai` (already pinned here as `pydantic-ai-slim[openai,anthropic,google,mcp]`) ships its own `pydantic_ai.providers.litellm.LiteLLMProvider`, built purely on the `openai` extra we already install. It's meant for exactly this: pointing at a self-hosted LiteLLM **proxy** (an actual gateway deployment translating 100+ providers into an OpenAI-shaped API), not importing the `litellm` SDK into our Django process. So:

- No `requirements/base.txt` change at all.
- A new `opencontractserver/pipeline/llm_providers/litellm_provider.py` (~30 lines, mirror `ollama_provider.py`): `provider_key = "litellm"`, `base_url` with no default (a gateway has none to assume), `api_key` optional (`requires_api_key = False`, same as Ollama, since some people run an unauthenticated internal proxy).
- One new branch in `model_factory.py::_construct_model` using `pydantic_ai.providers.litellm.LiteLLMProvider` + `OpenAIChatModel`. Two details to match from the existing `ollama` handling:
  - It needs a placeholder `api_key` string when none is configured. The underlying OpenAI client requires some non-empty value even against an unauthenticated server.
  - It should be exempted from the "warn if `base_url` is set without an `api_key`" check, for the same reason.

Your test file's approach (stubbing the module in `sys.modules`, checking `api_key` gets dropped from the call kwargs when it's unset) is worth carrying over into tests for the new provider file.

One thing for the new PR: make sure it's rebased off latest `main`. Quite a bit has moved since this one was opened.

PR #2030, opened 2026-06-20
Batch empty_corpus & folder cascade-delete into a shared bulk-trash primitive (#1951)

### LINKED ISSUE #1951: Bulk-trash primitive: batch per-document remove_document loops (empty_corpus, folder cascade-delete, empty-trash)
### Background

Three soft-delete paths currently loop over `Corpus.remove_document()` once per document inside a single `transaction.atomic()`:

- `DocumentLifecycleService.empty_corpus` (`opencontractserver/corpuses/services/lifecycle.py:~407`)
- `FolderCRUDService._trash_documents_in_subtree` (folder cascade-delete) (`opencontractserver/corpuses/services/folders.py:~927`)
- the legacy "empty trash" path

Each `remove_document` call issues several queries (history row, signals, path update) and holds row locks for the whole loop inside the caller's transaction. For corpora/sub-trees with thousands of documents this risks hitting the DB statement/connection timeout.

### Scope note (from PR #1941)

The 25,000-document `MAX_SELECT_ALL_DOCUMENT_IDS` cap bounds *Select All* but does **not** bound `emptyCorpus` or folder cascade-delete — a very large corpus can still be emptied in one interactive call.

### Proposed work

Introduce one shared bulk-trash primitive that batches the path/history updates and signal emission, and route all three call sites through it. Then it's safe to raise (or remove) the interactive document-count ceiling on these paths.

### Acceptance

- Single bulk-trash primitive replaces the three per-document loops.
- Trash listing / restorability behavior unchanged (history nodes + signals preserved).
- Benchmarked against a multi-thousand-document corpus without statement-timeout risk.

_Filed as the tracking issue referenced by the `TODO(perf, deferred)` comments in `lifecycle.py` and `folders.py`._

JSv4:
## Summary

Closes #1951.

Two soft-delete paths looped over `Corpus.remove_document()` once per document **inside a single `transaction.atomic()`**:

- `DocumentLifecycleService.empty_corpus` (`opencontractserver/corpuses/services/lifecycle.py`)
- `FolderCRUDService._trash_documents_in_subtree` (folder cascade-delete) (`opencontractserver/corpuses/services/folders.py`)

Each `remove_document` call issues several queries (active-path lookup, `is_current` flip, soft-delete insert, signals) and holds row locks for the **whole loop**. Because `ATOMIC_REQUESTS=True`, the entire mutation already runs in one request transaction, so on a multi-thousand-document corpus this risks exhausting the DB statement/connection timeout. Neither path is bounded by the 25,000-document `MAX_SELECT_ALL_DOCUMENT_IDS` Select-All cap (per the scope note from #1941). Both were marked `TODO(perf, deferred)` pointing at this issue.

## What changed

Added the shared **bulk-trash primitive** `DocumentLifecycleService.bulk_soft_delete_documents(corpus, document_ids, user)` — the batched counterpart of `Corpus.remove_document(document=...)`. For every active, non-deleted `DocumentPath` of the in-scope documents it:

1. supersedes every current head in **one** `UPDATE (is_current=False)`,
2. inserts the immutable soft-delete history nodes (`is_deleted=True, is_current=True`, parented to the superseded row) in **one** `bulk_create`,
3. replays `post_save(created=True)` per new row via the existing `CorpusPathService._dispatch_document_path_created_signals` helper (the same mechanism the bulk move/reconcile paths already use) — so the document-text embedding + `Readme.CAML` cache-refresh side-effects fire exactly as on the per-document path, and
4. revokes `is_public` on now-orphaned documents in a single query pair (mirrors `remove_document`'s tail).

`empty_corpus` and `_trash_documents_in_subtree` now delegate to it, retiring both per-document loops. The new rows are byte-for-byte identical in shape to what `remove_document` wrote, so the `select_related("document")` cache keeps the replayed signals query-free per row.

The query count is now **independent of the document count** — a fixed handful of statements instead of O(N) — which is the structural guarantee against the statement/connection timeout the loops risked.

## Acceptance criteria

- ✅ **Single bulk-trash primitive replaces the per-document soft-delete loops.** Both `remove_document` loops now route through `bulk_soft_delete_documents`.
- ✅ **Trash listing / restorability unchanged (history nodes + signals preserved).** Identical soft-delete history nodes; `post_save(created=True)` replayed per node (covered by a test). The bulk `UPDATE` skips only the `created=False` post_save on superseded rows, which both DocumentPath receivers either early-return on or re-trigger idempotently from the new node.
- ✅ **No statement-timeout risk on large corpora.** Proven at the unit level by `test_query_count_independent_of_document_count` (same query count for 2 docs as for 6) rather than a slow multi-thousand-doc CI benchmark.

## Scope note — why "empty trash" is deliberately excluded

The issue also names the legacy "empty trash" path. That path is `permanently_delete_all_in_trash` → `permanently_delete_document`, which **does not** call `remove_document`. It is a *permanent* delete that destroys history nodes (plus cascades annotations/relationships/summaries/document) and keeps **deliberate per-document partial-success semantics** (documented in `versioning.py`). It therefore cannot share a primitive whose contract is "history nodes + signals preserved / restorability unchanged" — the two are opposite operations. Folding it into an all-or-nothing batch would change its documented partial-success behavior, which would be out of step with a surgical fix. Batching that cascade is a separate, riskier follow-up if desired.

`CorpusDocumentService.remove_documents_from_corpus` (Select-All → Remove) is likewise left untouched: it is already bounded by the 25k cap, runs each removal in its own transaction, and intentionally allows per-document partial success.

## Tests

New `TestBulkSoftDeletePrimitive` in `test_corpus_objs_service.py`:

- restorable soft-delete history nodes created (old head superseded, new node `is_deleted`/`is_current`/`parent`/version/creator, 2 rows kept, document survives)
- distinct-document count semantics (re-trashing already-trashed ids returns 0)
- empty input short-circuits with **zero** DB access
- ids with no active path are a no-op
- `post_save` dispatched once per created node, none with `created=False` (signal-preservation + supersession invariant)
- **query count independent of document count** (the core #1951 guarantee, with a non-vacuous floor)
- `is_public` revoked when a doc leaves its only public corpus, and preserved when it remains in another public corpus
- `empty_corpus` wiring (delegates to the primitive) + folder cascade wiring (`_trash_documents_in_subtree`)

Existing `TestEmptyCorpus` and `TestFolderDelete_CascadeToTrash` assertions are unchanged and continue to hold (the resulting DB state is identical). `black`/`isort`/`flake8` pass on all changed files.

> Note: the full backend suite was not run here (no Docker in this environment); the changes are syntax-checked, lint-clean, and covered by the new + existing unit tests described above.


JSv4:
Thanks for the thorough review — addressed across `80b37a7` and `8a29fbf`:

**Done**
- **Multi-corpus `is_public` keep-test** — added `test_keeps_is_public_when_still_in_another_public_corpus` covering the `still_in_public` safety-net branch (doc with an active path in a second public corpus keeps `is_public` when trashed from the first).
- **Hardcoded `+2` query-count delta** — softened to `assertGreater`; the cross-size O(1) equality assertions remain the real guarantee.
- **`post_save(created=False)` invariant** — documented on `_dispatch_document_path_created_signals` (applies to all three bulk callers) and added a negative assertion in the dispatch test.
- **Rollback safety** — added a comment noting both replayed receivers defer via `transaction.on_commit`, so pre-commit dispatch is rollback-safe (Django discards the callbacks if the block rolls back — no embedding jobs for uncommitted rows).
- **`document_ids` snapshot / TOCTOU** — documented that the locked SELECT re-validates active state (concurrently-removed docs are skipped, added docs left alone), matching the prior per-document loop's semantics. No upstream-lock assumption is required for correctness.
- **`transaction.atomic()` nesting** — kept (self-contained/standalone safety, matching `Corpus.remove_document`) with a comment explaining the savepoint intent.

**Deliberately not changed (non-blocking)**
- **Promote `_dispatch_document_path_created_signals` to a public name** — out of scope for #1951; it's a broader refactor touching the two existing move/reconcile callers. The cross-service call matches established precedent.
- **`Iterable[int]` → `Collection[int]`** — keeping `Iterable`: the arg is `list()`-materialized immediately, so `Iterable` is the widest correct type; `Collection` would needlessly reject one-shot generators for no functional gain.

---

JSv4:
On the one item flagged as worth resolving before merge — **#1 `version_number` parity** — verified, no change needed:

- `Corpus.remove_document` **copies** the version (no increment): `version_number=active_path.version_number` (`models.py:1161`) and `version_number=path_record.version_number` (`models.py:1195`).
- The primitive does the same: `version_number=p.version_number`.
- `test_creates_restorable_soft_delete_history_nodes` seeds `version_number=1` and asserts the new soft-delete head is `1`, so an erroneous `+1` (→ 2) would fail the test.

The remaining points (#2–#6) are non-blocking polish, and I'm intentionally not pushing further commits for them — each push re-triggers the review bot, and the reviews have converged on approve. Quick notes: #6 is already done (the `trashed_doc_ids` computation is hoisted directly under the locked SELECT, before the writes); #3's asymmetry intentionally mirrors `remove_document`'s `is_public` tail; `folder_id` propagation (#5) matches `remove_document` (`models.py:1159/1193`) and is exercised end-to-end by `TestFolderDelete_CascadeToTrash`. Happy to fold in the `set`-difference tidy (#4) or the explicit `folder_id` assertion (#5) if a maintainer would like them before merge.

---

PR #1990, opened 2026-06-15
Open-vocabulary authority discovery (Phase 0+1): taxonomy + Tier-2a detection

JSv4:
## Summary

Makes the reference-enrichment engine **open-vocabulary** — it now detects citations to legal authorities outside the built-in registry (US Code, CFR, Federal Register, Public Law, Statutes at Large, state codes, and named Acts), classified by **jurisdiction** and **authority type**, surfaced through a new read-only `discover_authorities` agent tool. The trusted Tier-1 registry path is unchanged; open-vocabulary detection is additive and opt-in.

This is **Phase 0 + Phase 1** of a larger initiative (design + plan kept locally under `docs/superpowers/`).

### Phase 0 — taxonomy foundation (additive, no behaviour change)
- `CorpusReference` gains indexed `jurisdiction`, `authority_type`, `detection_tier`, `detection_confidence` (migrations `0080`, `0084`).
- New `AuthorityNamespace` registry model (`0081`), seeded from the static authority map (`0082`); existing rows classified by prefix (`0083`).
- `authority_alias_registry()` now reads the namespace table (global namespaces always; corpus-linked only when visible) — adding a body of law no longer needs a code change.

### Phase 1 — Tier-2a open-vocabulary detection
- `opencontractserver/enrichment/grammars.py` — generic citation-shape grammars + `opencontractserver/enrichment/abbreviations.py` state-code table.
- `opencontractserver/enrichment/reconcile.py` — registry wins on span overlap (scoped to matching `reference_type`).
- `EnrichmentService.discover()` + the `discover_authorities` agent tool group detected authorities by jurisdiction/type and flag prefixes not yet in the registry (`new_namespaces`).

## Test plan
- [ ] CI green (full backend suite)
- Comprehensive enrichment suite passes locally — **136 passed** (new: `test_generic_grammars`, `test_reconcile`, `test_abbreviations`, `test_authority_namespace`, `test_corpusreference_classification`, `test_corpusreference_backfill`, `test_enrichment_discovery`, `test_enrichment_classification_constants`; plus all existing enrichment suites as regression).
- No migration drift (`makemigrations --check`); `manage.py check` clean (no E001); black/isort/flake8/mypy pass.

## Follow-ups (not in this PR)

**Pre-existing issues** found during review in code this PR did not modify (`enrichment/services/enrichment_service.py`):
- `apply()` marks the `Analysis` `COMPLETED` *before* `_link_external()` runs — a `_link_external` failure leaves a permanently-wrong provenance status.
[...]

PR #2031, opened 2026-06-20
Add municipal citation grammar (Tier-2a authority discovery)

### LINKED ISSUE #1995: Authority discovery — municipal citation grammar (Phase 1 follow-up)
Deferred follow-up from Phase 1 (#1990).

Phase 1 shipped federal + state Tier-2a shape grammars but **deliberately deferred municipal citations**: their forms are too heterogeneous for a single precise regex, and a low-precision municipal matcher was judged worse than shipping none (it would silently misclassify or over-match).

## Scope
- Add a generic municipal grammar (`<Place> Municipal Code § N`, `<Place> Mun. Code §`, ordinance forms) + a `MUNICIPAL_ABBREVIATIONS` table in `opencontractserver/enrichment/`, same data-driven pattern as `abbreviations.py`.
- `authority_type="municipal-ordinance"`; jurisdiction codes like `us-ca-san-francisco`.
- Hold the same precision bar as the federal/state grammars: golden cross-municipality test corpus, explicit false-positive guards, confidence calibrated below structured federal cites.

See `opencontractserver/enrichment/grammars.py` + `abbreviations.py` from #1990 for the established pattern.

JSv4:
Closes #1995.

Completes the federal/state/**municipal** trio of Tier-2a citation-shape grammars deliberately deferred from Phase 1 (#1990). Municipal citation forms are too heterogeneous for a single precise regex, so this follows the **exact pattern already established** for state codes (`STATE_CODE_ABBREVIATIONS` table) + bare Acts (open-vocabulary grammar) — no new architecture.

## What it does

Two complementary layers, both emitting `authority_type=municipal-ordinance`:

**1. Known municipal codes — `MUNICIPAL_CODE_ABBREVIATIONS` table** (`opencontractserver/enrichment/abbreviations.py`)

Exact code names → a `muni-` canonical-key prefix, the **full hierarchical jurisdiction** (e.g. `us-ca-san-francisco`), at confidence `0.8`. Seeded across six municipalities / five states (San Francisco, Los Angeles, New York City, Chicago, Seattle, Houston), with both spelled-out and Bluebook-abbreviated forms (`San Francisco Municipal Code`, `S.F. Mun. Code`, `N.Y.C. Admin. Code`, `Houston Code of Ordinances`, …). Adding a city is a data edit, never a grammar change.

**2. Open-vocabulary shape grammar** (`opencontractserver/enrichment/grammars.py`)

Matches `[City] Municipal Code § N`, `Mun. Code §`, `Code of Ordinances §`, and `Ordinance No. N` for cities **outside** the table, keyed under the **same** `muni-` namespace at confidence `0.6`. Jurisdiction is left `None` (free text reveals a city but not its state). Because the prefix is shared with the table, a captured city never fragments from its table entry, and tabling that city later **upgrades** existing mentions instead of orphaning them.

## Precision bar (held to the federal/state standard)

- A `§`/`Section`/`Sec.` + number **anchor is mandatory** — prose like *"the city adopted a new Municipal Code last year"* never matches.
- The code phrase must be **capitalised** (lowercase "municipal code" excluded).
- **"Administrative Code" is intentionally kept out of the open-vocab grammar** so `Texas Administrative Code § 1.5` (a *state* regulation) is never mis-tagged municipal; named municipal admin codes (NYC) live in the table instead.
- The table pass **claims its span** so a known code never double-emits with its generic shadow (verified: exactly one candidate).
- Every municipal confidence is **below structured federal cites** (< 0.9).

`classify_prefix` (`constants.py`) recognises the `muni` / `muni-` prefix family **by shape**, so a municipal key is never stranded at `(None, None)` authority_type in the frontier or governance graph. **No migration or `AuthorityNamespace` seed is required** — table candidates carry their taxonomy inline, exactly like the state codes.

## Tests

`opencontractserver/tests/test_generic_grammars.py`, `test_abbreviations.py`, `test_enrichment_classification_constants.py`:

- **Golden cross-municipality corpus** — one realistic multi-city paragraph asserting all six known codes resolve with the correct jurisdiction.
- False-positive guards (no anchor, lowercase, `Texas Administrative Code`, bare ordinance, plain numbers).
- NYC namespace-sharing regression (the `New York City Municipal Code` form resolves to the table's `muni-new-york`, not a fragmented `muni-new-york-city`).
- Bare-placeholder collapse, trailing city-qualifier preservation, ordinance precision-bar tradeoff, confidence calibration, OCR double-space tolerance, dedup, and `muni`-family classification.

All tests in the affected files pass under the Django `SimpleTestCase` runner (no regressions); `black`/`isort`/`flake8`/`pyupgrade` clean.

### DIFF
--- a/changelog.d/1995-municipal-citations.added.md
+++ b/changelog.d/1995-municipal-citations.added.md
@@ -1,0 +1,38 @@
+- **Authority discovery: municipal citation grammar (Tier-2a, issue #1995).**
+  The reference-enrichment engine now detects municipal-code and ordinance
+  citations, completing the federal/state/municipal trio deliberately deferred
+  from Phase 1 (#1990). Two layers, mirroring the established state-code table +
+  bare-act open-vocabulary pattern:
+  - **Known municipal codes** — a new `MUNICIPAL_CODE_ABBREVIATIONS` table
+    (`opencontractserver/enrichment/abbreviations.py`) maps exact code names
+    (e.g. "San Francisco Municipal Code", "S.F. Mun. Code", "N.Y.C. Admin.
+    Code", "Houston Code of Ordinances") to a `muni-<city-slug>` canonical-key
+    prefix, the full hierarchical jurisdiction (`us-ca-san-francisco`), and
+    `authority_type=municipal-ordinance`, at `_CONF_MUNICIPAL` (0.8). Adding a
+    city is a data edit, never a grammar change.
+  - **Open-vocabulary shape grammar** — `GenericCitationExtractor` now matches
+    `[City] Municipal Code § N`, `Mun. Code §`, `Code of Ordinances §`, and
+    `Ordinance No. N` forms for cities outside the table
+    (`opencontractserver/enrichment/grammars.py`), keyed under the SAME
+    `muni-<city-slug>` namespace (jurisdiction left `None` — free text reveals a
+    city but not its state) at `_CONF_MUNICIPAL_GENERIC` (0.6). A captured city
+    therefore shares its prefix with any table entry for the same city, so the
+    two never fragment, and adding the city to the table later upgrades existing
+    mentions instead of orphaning them.
+  - **Precision guards** (held to the federal/state bar): a `§`/`Section`/`Sec.`
+    + number anchor is mandatory (so prose like "the city adopted a new
+    Municipal Code" never matches); the code phrase must be capitalised
+    (lowercase "municipal code" is excluded); "Administrative Code" is kept out
+    of the open-vocab grammar so "Texas Administrative Code" (a STATE
+    regulation) is never mis-tagged municipal; the table pass claims its span so
+    a known code never double-emits with its generic shadow; and every municipal
+    confidence is calibrated below structured federal cites (< 0.9).
+  - `classify_prefix` (`opencontractserver/enrichment/constants.py`) recognises
+    the `muni` / `muni-<city>` prefix family by shape, so a municipal key is
+    never stranded at `(None, None)` authority_type in the frontier or
+    governance graph. No migration or `AuthorityNamespace` seed is required
+    (table candidates carry their taxonomy on the candidate, exactly like the
+    state codes). Tests: golden cross-municipality corpus, false-positive
+    guards, confidence calibration, OCR tolerance, and dedup in
+    `opencontractserver/tests/test_generic_grammars.py`,
+    `test_abbreviations.py`, `test_enrichment_classification_constants.py`.
--- a/opencontractserver/enrichment/abbreviations.py
+++ b/opencontractserver/enrichment/abbreviations.py
@@ -57,6 +57,13 @@ MUNICIPAL_CODE_ABBREVIATIONS: dict[str, tuple[str, str, str]] = {
         "us-ca-san-francisco",
         C.AUTHORITY_TYPE_MUNICIPAL,
     ),
+    # The abbreviated city + spelled-out "Municipal Code" form is tabled too: the
+    # abbreviation slug ("s-f") would otherwise fragment from "san-francisco".
+    "S.F. Municipal Code": (
+        "muni-san-francisco",
+        "us-ca-san-francisco",
+        C.AUTHORITY_TYPE_MUNICIPAL,
+    ),
     # Los Angeles, CA
     "Los Angeles Municipal Code": (
         "muni-los-angeles",
@@ -64,6 +71,12 @@ MUNICIPAL_CODE_ABBREVIATIONS: dict[str, tuple[str, str, str]] = {
         C.AUTHORITY_TYPE_MUNICIPAL,
     ),
     "L.A. Mun. Code": (
+        "muni-los-angeles",
+        "us-ca-los-angeles",
+        C.AUTHORITY_TYPE_MUNICIPAL,
+    ),
+    # As with S.F. above — "l-a" must not fragment from "los-angeles".
+    "L.A. Municipal Code": (
         "muni-los-angeles",
         "us-ca-los-angeles",
         C.AUTHORITY_TYPE_MUNICIPAL,
--- a/opencontractserver/enrichment/abbreviations.py
+++ b/opencontractserver/enrichment/abbreviations.py
@@ -71,17 +71,34 @@ MUNICIPAL_CODE_ABBREVIATIONS: dict[str, tuple[str, str, str]] = {
     # New York City, NY — the consolidated code IS the "Administrative Code"
     # ("Administrative Code" is intentionally kept OUT of the open-vocab grammar
     # because "Texas Administrative Code" is a STATE regulation, not municipal).
-    # NB: the slug is ``muni-new-york`` (NOT ``muni-new-york-city``). The
-    # open-vocab grammar would slug a literal "New York City Municipal Code" cite
-    # as ``muni-new-york-city`` — a DIFFERENT prefix — so any "New York City
-    # Municipal Code"/"Mun. Code" form added here later MUST also key
-    # ``muni-new-york`` to keep the two from fragmenting into rival authorities.
+    # ALL NYC forms key ``muni-new-york`` (NOT ``muni-new-york-city``). The
+    # informal "New York City Municipal Code"/"Mun. Code" spellings are tabled
+    # HERE as aliases precisely so they resolve to this one authority instead of
+    # fragmenting to the open-vocab ``muni-new-york-city`` slug — "New York City"
+    # ends in "City", which the open-vocab city-capture would otherwise carry
+    # into the key. Tabling them eliminates the fragmentation (vs merely warning
+    # about it). Any further NYC spelling added later must also key muni-new-york.
     "New York City Administrative Code": (
         "muni-new-york",
         "us-ny-new-york",
         C.AUTHORITY_TYPE_MUNICIPAL,
     ),
     "N.Y.C. Admin. Code": (
+        "muni-new-york",
+        "us-ny-new-york",
+        C.AUTHORITY_TYPE_MUNICIPAL,
+    ),
+    "New York City Municipal Code": (
+        "muni-new-york",
+        "us-ny-new-york",
+        C.AUTHORITY_TYPE_MUNICIPAL,
+    ),
+    "New York City Mun. Code": (
+        "muni-new-york",
+        "us-ny-new-york",
+        C.AUTHORITY_TYPE_MUNICIPAL,
+    ),
+    "N.Y.C. Mun. Code": (
         "muni-new-york",
         "us-ny-new-york",
         C.AUTHORITY_TYPE_MUNICIPAL,
--- a/opencontractserver/enrichment/abbreviations.py
+++ b/opencontractserver/enrichment/abbreviations.py
@@ -71,6 +71,11 @@ MUNICIPAL_CODE_ABBREVIATIONS: dict[str, tuple[str, str, str]] = {
     # New York City, NY — the consolidated code IS the "Administrative Code"
     # ("Administrative Code" is intentionally kept OUT of the open-vocab grammar
     # because "Texas Administrative Code" is a STATE regulation, not municipal).
+    # NB: the slug is ``muni-new-york`` (NOT ``muni-new-york-city``). The
+    # open-vocab grammar would slug a literal "New York City Municipal Code" cite
+    # as ``muni-new-york-city`` — a DIFFERENT prefix — so any "New York City
+    # Municipal Code"/"Mun. Code" form added here later MUST also key
+    # ``muni-new-york`` to keep the two from fragmenting into rival authorities.
     "New York City Administrative Code": (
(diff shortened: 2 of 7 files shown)


JSv4:
Thanks for the reviews — addressed in 938de93. Summary of what changed and one correction:

**Applied**
- **Public API in tests** — `test_abbreviations.py` now asserts `classify_prefix(prefix)[1] == AUTHORITY_TYPE_MUNICIPAL` instead of matching the private `_MUNI_PREFIX_RE`, pinning the contract rather than the implementation.
- **`_is_claimed` readability** — rewritten via De Morgan (`start < ce and end > cs`) with a comment; behaviour identical.
- **Bare `City Municipal Code § N` placeholder** — added `city`/`county` to `_CITY_STOPWORDS` so a template with no real city collapses to the honest `muni:` key instead of inventing a `muni-city` authority. Only *leading* stopwords drop, so `Kansas City` → `muni-kansas-city` and `Marin County` → `muni-marin-county` are preserved (two new tests).
- **Ordinance precision bar** — new false-positive test documents that internal numbering stays confidence `<0.8` with `jurisdiction=None`, plus an OCR double-space test and a docstring note that ordinance keys aren't table-upgradeable.

**Not applied (claim was incorrect)** — the "`City of Portland Municipal Code § 1` → `muni-city`" false-negative. I verified against the live regex: it yields `muni-portland:1` (and `City of Oakland …` → `muni-oakland:5.04.010`). The lowercase "of" breaks the capitalised run, so backtracking starts the match at the real city and "City of" is dropped. The genuine edge case is only the bare `City Municipal Code` form (no "of"/no real city), which the stopword change above now handles. Adding "of"/"City" handling to the *city-of-X* path would have addressed a non-existent bug.

**Deferred (agreed out of scope)** — `_is_claimed` → `bisect` (fine at ~12 entries) and additional Bluebook abbreviations (`LAMC`/`SMC`/`SFMC`), which are additive data edits for a follow-up.

61 tests pass; mypy/black/flake8/isort clean.

---

JSv4:
Both "before merge" items from the full review are addressed in 706076d (docs only, no logic change):

1. **NYC slug** — added a note on the `New York City Administrative Code` table entry: the slug is `muni-new-york` (not `muni-new-york-city`), and any literal `New York City Municipal Code`/`Mun. Code` form added later must key `muni-new-york` so it doesn't fragment from the open-vocab slug. (Confirmed empirically: open-vocab `New York City Municipal Code § 5` → `muni-new-york-city`.) I kept this as a comment rather than adding a table alias, since NYC's official code is the Administrative Code, not a "Municipal Code".
2. **Optional `§` asymmetry** — added a note in the `_muni_re` construction that the optional `§` is intentional (mirrors `_state_re`; a named table code is already trusted) and deliberately asymmetric with the open-vocab `_MUNI_CONN`, which requires the anchor as its core precision guard.

Items 3 (ordinance precision bar) and 4 (`LAMC`) need no change per the review — the former is tested as an accepted tradeoff, the latter is an additive data follow-up. 61 tests pass; mypy/black/flake8/isort clean.

---

JSv4:
Thanks — all three items addressed.

1. **Blocker (CLAUDE.md footer)** — removed the `_Generated by [Claude Code]_` footer from the PR description. (CLAUDE.md rule #3 takes precedence over the default footer.)
2. **NYC fragmentation (Medium)** — went with the fuller fix in e1ab07b: tabled the informal `New York City Municipal Code` / `New York City Mun. Code` / `N.Y.C. Mun. Code` spellings as aliases of `muni-new-york` (parallel to the `S.F. Mun. Code` forms). The table pass claims the span, so these now resolve to the single NYC authority with full jurisdiction and never fragment to `muni-new-york-city`. Added `test_nyc_municipal_code_form_resolves_to_table_prefix` as the regression guard (asserts `muni-new-york:27-2004` present, `muni-new-york-city:27-2004` absent). I preferred resolving over the deferred-comment approach precisely because, as you noted, a "MUST also key …" note is a fragmentation accident waiting to happen.
3. **Ordinance downstream guidance (Low)** — added a "Downstream note" to the `_municipal_generic` docstring: low confidence + `None` jurisdiction are the filter signals; consumers should treat such mentions as provisional (surfaced for review, never promoted to the trusted tier) until corroborated — called out specifically for the anchorless ordinance pseudo-city case.

62 tests pass; mypy/black/flake8/isort clean.

---

JSv4:
Thanks for the review. Status on each item:

**Must Fix — Claude Code footer:** Already removed from the PR description in the previous round (the review may have rendered against the pre-edit body). ✅

**Should Fix — "comment/docstring verbosity":** Holding here, because the cited rule does not exist. I grepped `CLAUDE.md` for the quoted text — *"Never write multi-paragraph docstrings or multi-line comment blocks — one short line max"* — and there is no such rule (no "one short line", "multi-paragraph", or "multi-line comment" prohibition anywhere in the file). The actual guidance is to *match the surrounding code's comment density*, and this module's established, pre-existing style is multi-line explanatory docstrings — e.g. `extractor.py::_terms` (predates this PR) is a 4-line paragraph; `grammars.py::_canonical_act_prefix` is a ~20-line docstring. My comments deliberately mirror that convention. Collapsing them would make this file inconsistent with its own module. Happy to trim anything genuinely redundant, but not against a rule that isn't in the repo.

**Observation — `No\.?\s*` in `_MUNI_ORDINANCE_RE`:** Keeping `\s*`. The suggested `\s+` is a regression — it rejects the valid compressed form `Ordinance No.5` / `Ord. No.2021-15`:

| input | current `\s*` | suggested `\s+` |
|---|---|---|
| `Ordinance No. 5` | ✅ | ✅ |
| `Ordinance No.5` | ✅ | ❌ |
| `Ordinance No5` | ✅ | ❌ |

The only thing `\s*` over-accepts is the unnatural `No5` concatenation, which (if it ever occurred) yields a 0.6-confidence, `jurisdiction=None` provisional candidate — exactly the documented downstream-filter contract — so it's harmless. Tightening to reject `No5` while keeping `No.5` is possible but not worth the added regex for a form that doesn't occur.

**Observation — `_is_claimed` O(n·m):** No action (you concurred; n is single-digit per document).

**Observation — `L.A. Municipal Code` absent:** Good catch — fixed in 4921cbb. The table already lists `L.A. Mun. Code`/`S.F. Mun. Code` precisely to stop the abbreviation slug from fragmenting, so I added the spelled-out `L.A. Municipal Code` / `S.F. Municipal Code` variants (which would otherwise key `muni-l-a`/`muni-s-f`) plus a regression test. The table targets common forms, not an exhaustive cross-product, so rarer combinations still rely on the open-vocab safety net at 0.6 confidence.

63 tests pass; mypy/black/flake8/isort clean.

---

JSv4:
Thanks — good batch. Status per item:

**Issue 2 (sentence-lead absorption) — fixed in 4428abd.** This was the real one: `See Oakland Municipal Code § 5` (very common in legal prose) was keying `muni-see-oakland`. Added the frequent single-word leads (`see`/`under`/`per`/`accord`/`cf`/`compare`/`contra`) to `_CITY_STOPWORDS` (leading-strip only, so `Kansas City` is untouched; `Pursuant to X` was already fine since the lowercase "to" breaks the run). Documented it as a pragmatic denylist, not exhaustive — the open-vocab capture stays provisional (0.6) for the long tail. Added a regression test.

**Open-vocab OCR coverage gap — added.** New `test_open_vocab_tolerates_ocr_double_space` (`Oakland  Municipal  Code § 5`) pins the open-vocab path's double-space tolerance, paralleling the table-path test.

**Issue 1 (NYC "Code of Ordinances") — declining.** Verified it does fragment to `muni-new-york-city`, but NYC's consolidated code is the *Administrative Code* — it is never called a "Code of Ordinances" (that's Houston's term). Tabling `New York City Code of Ordinances → muni-new-york` would assert a code name that doesn't exist, which is worse than the near-zero-probability fragmentation it would prevent. This is the same line I drew when adding the L.A./S.F. aliases: table *plausible real* spellings, not invented ones. I also didn't add the suggested test, since it would pin fragmented behavior as "expected" and codify a wart for a string that won't occur — the open-vocab provisional contract already covers it.

**Issue 3 (ordinance `\s*` → `\s+`) — declining (same as the prior review).** `\s+` is a regression: it rejects the valid compressed form `Ordinance No.5` / `Ord. No.2021-15`. The only thing `\s*` over-accepts is `No5` (no separator), which yields a 0.6-confidence, `jurisdiction=None` provisional candidate per the documented contract — harmless.

**Issue 4 (Houston Bluebook abbreviation) — none exists.** There's no established Bluebook short form for the Houston Code of Ordinances (unlike `S.F.`/`L.A.`/`N.Y.C.`), so none is tabled — consistent with Chicago, which likewise has no abbreviated entry. The open-vocab path covers `Houston Code of Ordinances`-style cites for the spelled form regardless.

**Minor observations** (`_MUNI_SEC` dash position, `_is_claimed` linear scan, None-guard parity) — no action; you flagged each as not-a-bug / no-concern, and I concur.

65 tests pass; mypy/black/flake8/isort clean.

---

JSv4:
Good review — Issue 1 was a real regression I introduced last round. Status:

**Issue 1 (`contra`/`accord` stopword collision) — fixed in 0d75069.** Confirmed: with `contra` as a leading stopword, `Contra Costa Code of Ordinances § 4.1` keyed `muni-costa` (the leading "contra" was stripped) — a silent corruption of a real CA jurisdiction. Trimmed `_CITY_STOPWORDS` back to the unambiguous, zero-collision leads only — `see` and `under` — and dropped `per`/`accord`/`cf`/`compare`/`contra`. Added a regression test asserting `Contra Costa [County] Code of Ordinances` stays `muni-contra-costa[-county]`, plus an inline note on why colliding signals are excluded. Thanks for catching this.

**Issue 2 (NYC ordinance fragmentation) — added a test.** Framed around the *contract* rather than pinning the fragmented slug as desirable: `test_nyc_ordinance_form_is_provisional_not_table_resolved` asserts an NYC ordinance cite carries the provisional contract (confidence `<0.8`, `jurisdiction=None`). Ordinance numbers are explicitly not table-upgradeable (the `_municipal_generic` docstring says so) — they're discovery signals, not resolved authorities — so this isn't fragmentation of a *resolved* authority; the test makes that behavior explicit and guards against silent regression.

**Issue 3 (`No\.?` dot optional vs comment) — fixed the comment.** The optional dot is correct (`Ordinance No 7` is a valid form, so I kept `No\.?`), but the comment overstated "No." as required. Updated it to "`No`/`No.` … the trailing dot is optional".

**Issue 4 (`{0,3}` → `{0,2}`) — declining, with a counterexample.** Tightening the bound corrupts genuine 3-word municipalities: `San Luis Obispo Municipal Code § 5` becomes `muni-luis-obispo` under `{0,2}` (added a regression test pinning `muni-san-luis-obispo` under `{0,3}`). The `Proposed Draft Municipal Code` false positive stays a provisional 0.6 / `None`-jurisdiction candidate per the documented contract, which is the better tradeoff than truncating real multi-word city names.

68 tests pass; mypy/black/flake8/isort clean.

---

JSv4:
Thanks. Status:

**Recommended — non-stopword-lead test:** Added in 76d21bb. `test_non_stopword_lead_lands_in_slug_at_provisional_confidence` pins exactly the case you described — `Regarding Oakland Municipal Code § 5` → `muni-regarding-oakland:5` (not silently `muni-oakland`), with `jurisdiction=None` and `confidence<0.8`. This documents the minimal-stopword design (keep the denylist small to avoid place-name collisions; non-city leads land in the slug at provisional confidence) as a first-class invariant.

**Downstream-filtering question (raised twice) — confirmed, with the precise mechanics:** Grammar-tier municipal candidates are persisted on `CorpusReference` with `detection_tier="grammar"`, `detection_confidence` (0.8 table / 0.6 open-vocab), and `jurisdiction` — those signals are durable and load-bearing. To be precise about the gating: the `needs_review` review-bucket path is **LLM-tier-specific** (`llm_citation_extractor.py` sets `needs_review = conf < LLM_CONFIDENCE_FLOOR`); grammar-tier candidates (municipal *and* the existing federal/state ones) are applied as mentions stamped at `tier=grammar`, never as `registry` (the trusted tier). So "provisional" here means *tier=grammar + low confidence + None jurisdiction*, which the inventory / governance-graph / frontier consumers read to tier-filter — not a separate review queue. The municipal grammar doesn't change this contract; it follows the established grammar-tier behavior exactly. (If we later want a confidence floor that routes sub-threshold *grammar* hits to the review bucket too, that's a system-wide enrichment change, out of scope for this PR.)

**Optional — `_keys` test-helper dedup:** Holding. The per-class `_keys` helper is the established convention in this file — every pre-existing grammar test class (`FederalGrammarTests`, `StateGrammarTests`, `BareActGrammarTests`, `GrammarRobustnessTests`) defines its own. A shared base/module-level helper would either fragment the convention (new classes differ from old) or require editing the pre-existing classes (touching old tests). Matching the file's existing pattern is the more consistent call for a trivial 2-line helper.

69 tests pass; mypy/black/flake8/isort clean.

---

Issue #1986, opened 2026-06-12
Permissioning audit follow-ups (from PR #1985)


JSv4:
Durable tracking for the out-of-scope items surfaced during the 2026-06 permissioning audit and PR #1985's review rounds. Each was verified against code; none is a regression introduced by that PR. Ordered roughly by security relevance.

### Parity / security posture (pre-existing)

1. **Annotation-creator privacy parity edge.** `AnnotationQuerySet.visible_to_user`'s privacy gate carries `Q(creator=user)` (annotated with a pointer to this issue; divergence pinned at CI time by `test_annotation_creator_parity_gap_sentinel`), but `AnnotationManager.user_can`'s privacy recursion (`_source_privacy_recursion_passes`) has no creator exemption — an annotation creator lacking source READ sees their own analysis-/extract-rooted annotation in lists while `user_can(READ)` denies it. Relationships handle this deliberately (creator short-circuit mirrored by the queryset gate). Fixing annotations means choosing a semantic, flipping both sentinel assertions together, and adding the fixture to the invariant matrix.
2. **ChatMessage anonymous branch lacks the THREAD restriction.** `ChatMessageQuerySet.visible_to_user` anonymous path filters `conversation__is_public=True` without `conversation_type=THREAD`, while `ConversationQuerySet.visible_to_user` is THREAD-only for anonymous — a public CHAT's *messages* would be anonymous-visible while the conversation itself is hidden (`opencontractserver/conversations/models.py`).
3. **Conversations querysets miss group-level guardian grants.** `ConversationQuerySet` / `ChatMessageQuerySet` consult user-level object permissions only — same filter/check parity class PR #1985 fixed for annotations/relationships/extracts.
4. **`BaseVisibilityManager.visible_to_user` broad exception fallback.** `except (ImportError, Exception)` (`opencontractserver/shared/Managers.py` ~line 339) silently downgrades to creator/public filtering on ANY error — under-disclosing (drops guardian rows) and hiding programming errors. Intended catch is `(ImportError, LookupError)`. Needs its own change since tightening the envelope may surface latent errors suite-wide.
5. **`ExtractType` has no `get_queryset` override.** Related-field traversal (e.g. `DatacellType.extract`) relies on parent gating rather than `ExtractManager.visible_to_user`. Consider a `get_queryset` guard for defense-in-depth after the PR #1985 anonymous lockdown.
[...]

PR #2032, opened 2026-06-20
Permissioning audit follow-ups (#1986): conversation/message visibility parity, safe manager fallback, datacell crash

JSv4:
## Summary

Addresses the **surgical, still-present security + correctness items** from the #1986 permissioning-audit tracking issue. #1986 lists 14 follow-ups of widely varying scope; this PR takes the four that are unambiguous, high-value, and genuinely surgical, and dispositions the rest below (rather than forcing risky/architectural changes into one PR — per the "don't run wild" guidance).

Refs #1986 (intentionally **not** `Closes`, since 14 items remain on the tracker).

### Item 2 — anonymous chat-message visibility now follows the CHAT/THREAD bifurcation *(security)*
`ChatMessageQuerySet.visible_to_user`'s anonymous branch filtered `conversation__is_public=True` alone (`opencontractserver/conversations/models.py`). But `ConversationQuerySet.visible_to_user` keeps a public **CHAT** hidden from anonymous users (anonymous can only see THREADs), so a public CHAT's *messages* were anonymous-visible while the conversation itself was not. The branch now delegates to `Conversation.objects.visible_to_user` — the single source of truth — so messages inherit the bifurcation. This also fixes a latent **under**-disclosure: messages in context-inherited public-resource threads are now visible, matching conversation visibility and the authenticated path (which already delegates the same way).

### Item 3 — conversation/message querysets honour group-level READ grants *(security)*
`ConversationQuerySet` / `ChatMessageQuerySet` consulted only the **user**-level object-permission tables. Because `ConversationManager.user_can(READ)` / `ChatMessageManager.user_can(READ)` route *back through* `visible_to_user`, a group-only READ grant was **entirely non-functional** for these models — invisible in lists *and* denied by `user_can(READ)` — even though non-READ writes (`_default_user_can`) honour the same group grant. Both querysets now join the `*groupobjectpermission` tables (mirroring `BaseVisibilityManager`), closing the same filter/check group-grant gap PR #1985 closed for annotations/relationships/extracts.

### Item 4 — `BaseVisibilityManager.visible_to_user` stops swallowing unexpected errors *(security)*
The guardian-lookup fallback caught `(ImportError, Exception)` — equivalent to a bare `except Exception` (`ImportError ⊂ Exception`). Any error (programming bug, DB error, malformed grant row) silently degraded the queryset to creator/public filtering, **under-disclosing** guardian-granted rows while hiding the defect (`opencontractserver/shared/Managers.py`). Narrowed to `(ImportError, LookupError)`: a genuinely-absent permission table still degrades gracefully, but real errors now propagate.

### Item 7 — `Datacell._validate_manual_entry` AttributeError *(correctness)*
The "required" check read `self.column.validation_config.get("required")` *before* the `config = ... or {}` guard (`opencontractserver/extracts/models.py`). `validation_config` is a nullable JSONField, so validating a manual-entry cell with no `value` key on a config-less column raised `'NoneType' object has no attribute 'get'`. The `or {}` guard is hoisted and reused.

## Tests
- `test_conversation_permissions.py`: anonymous message bifurcation (item 2) + new `TestConversationGroupGrants` (item 3, asserting filter↔`user_can` parity for both conversations and messages).
- `test_shared_managers.py`: item 4 — unexpected error propagates; missing permission table still falls back gracefully (both halves of the narrowed envelope).
- `test_metadata_columns.py`: item 7 — null-config cell validates cleanly; a `required` config still raises.

**Validation note:** `python -m compileall`, `black --check`, `flake8`, and `scripts/collate_changelog.py --check` all pass locally. The dockerized test suite could **not** be executed in this environment — base-image pulls hit Docker Hub's unauthenticated rate limit (HTTP 429) — so the new/affected tests rely on CI to run. They're written against the existing fixtures/patterns in each file.

## Disposition of the remaining #1986 items (not in this PR)

| # | Item | Disposition |
|---|------|-------------|
| 1 | Annotation-creator privacy parity | ✅ **Already fixed** by PR #1985 (`AnnotationManager.user_can` now mirrors the queryset's `Q(creator=user)` source-privacy exemption; the sentinel was replaced by `test_annotation_creator_source_private_row_has_filter_check_parity`). |
| 5 | `ExtractType` has no `get_queryset` | Deferred. graphene-django's `get_queryset` does not guard single-FK traversal (the cited `DatacellType.extract` case), and the real protections (`get_node` + service-filtered `resolve_extracts`/`resolve_analyses`) already exist; adding it risks double-filtering connections. Marginal defense-in-depth value vs. regression risk — wants its own PR + test design. |
| 6 | `DocumentRelationshipService` myPermissions under-report | Deferred. Fail-safe (the mutation gate `user_has_permission` is correct; only the UI affordance under-reports). A correct fix needs per-permission *editable-by-user* subqueries (create/update/delete × source/target doc + corpus) — a new queryset concept that's not surgical. Own PR. |
| 8 | `analysis_id == 0` sentinel | Code-quality; touches the GraphQL contract (`__none__` mapping). Separate PR. |
| 9 | `require_permission` truthiness inversion | Code-quality rename + deprecation shim across call sites. Separate PR. |
| 10 | Mention-autocomplete inline guardian | Service migration; legal under E001 today and documented as an exception. Separate PR. |
| 11 | Weak assertions in `test_query_optimizer_methods.py` | Test tightening; do when next touching that file. |
| 12 | `source_visibility.py` micro-dedup | Explicitly deferred on the issue ("only if profiling shows it matters" — compiles to one SQL statement today). |
| 13 | Guardian codename string literals repo-wide | Large repo-wide pass; explicitly declined in the PR #1985 review to avoid diverging convention in a security PR. Own PR. |
| 14 | `claude-review` workflow malfunction | Tooling/workflow bug, not a codebase fix verifiable here. |

---
_Generated by [Claude Code](https://claude.ai/code/session_01Y4cYLCSNpV7ApzDggAF63Y)_

### DIFF
--- a/changelog.d/1986-perm-followups.fixed.md
+++ b/changelog.d/1986-perm-followups.fixed.md
@@ -1,0 +1,11 @@
+- Fixed an `AttributeError` crash in `Datacell._validate_manual_entry`
+  (`opencontractserver/extracts/models.py`, issue #1986 item 7). The "required"
+  check in the missing-`value` branch called
+  `self.column.validation_config.get("required")` *before* the
+  `config = self.column.validation_config or {}` guard, so validating a
+  manual-entry datacell with no `value` key on a column whose `validation_config`
+  is `None` (a nullable JSONField — legitimately null on explicitly-cleared or
+  legacy rows) raised `'NoneType' object has no attribute 'get'`. The `or {}`
+  guard is now resolved once at the top of the method and reused, so a config-less
+  column validates cleanly (nothing is required) while a column that *does* set
+  `required` still raises the expected `ValidationError`.
--- a/changelog.d/1986-perm-followups.security.md
+++ b/changelog.d/1986-perm-followups.security.md
@@ -1,0 +1,30 @@
+- Permissioning audit follow-ups from issue #1986 (filter/check parity + safe
+  error handling):
+  - **Anonymous chat-message visibility now follows the conversation
+    bifurcation (item 2).** `ChatMessageQuerySet.visible_to_user`
+    (`opencontractserver/conversations/models.py`) previously filtered the
+    anonymous branch on `conversation__is_public=True` alone, which exposed a
+    public **CHAT**'s messages even though `ConversationQuerySet.visible_to_user`
+    keeps the conversation itself hidden from anonymous users (anonymous can
+    only see THREADs). The branch now delegates to
+    `Conversation.objects.visible_to_user`, so message visibility inherits the
+    single-source-of-truth CHAT/THREAD rules — no public-CHAT message leak, and
+    messages in context-inherited public-resource threads are correctly visible.
+  - **Conversation and chat-message querysets now honour group-level READ
+    grants (item 3).** `ConversationQuerySet` / `ChatMessageQuerySet`
+    `visible_to_user` consulted only the USER object-permission tables. Because
+    `ConversationManager.user_can(READ)` / `ChatMessageManager.user_can(READ)`
+    route back through `visible_to_user`, a group-only READ grant was both
+    invisible in lists AND denied by `user_can(READ)` — even though non-READ
+    writes (via `_default_user_can`) honoured the same group grant. Both
+    querysets now join the `*groupobjectpermission` tables (mirroring
+    `BaseVisibilityManager`), closing the same group-grant gap PR #1985 fixed for
+    annotations / relationships / extracts.
+  - **`BaseVisibilityManager.visible_to_user` no longer swallows unexpected
+    errors (item 4).** The guardian-lookup fallback caught
+    `(ImportError, Exception)` — equivalent to a bare `except Exception` — so any
+    error silently degraded the queryset to creator/public filtering,
+    under-disclosing guardian-granted rows while hiding the defect
+    (`opencontractserver/shared/Managers.py`). The catch is narrowed to
+    `(ImportError, LookupError)` so a genuinely-absent permission table still
+    falls back gracefully, but programming/database errors now propagate.
--- a/opencontractserver/conversations/models.py
+++ b/opencontractserver/conversations/models.py
@@ -223,7 +223,15 @@ class ConversationQuerySet(SoftDeleteQuerySet):
             )
             return queryset.filter(anon_filter).distinct().order_by("-created")
 
-        # Get explicitly permitted conversation IDs via guardian
+        # Get explicitly permitted conversation IDs via guardian (user- AND
+        # group-level READ grants). ``ConversationManager.user_can(READ)``
+        # routes back through this queryset, so the filter IS the check — but
+        # non-READ writes delegate to ``_default_user_can``, which honours group
+        # grants by default. Consulting only the USER object-permission table
+        # here therefore made a group-only READ grant entirely non-functional:
+        # invisible in lists AND denied by ``user_can(READ)`` (issue #1986
+        # item 3 — the same filter/check group-grant gap PR #1985 closed for
+        # annotations / relationships / extracts).
         model_name = self.model._meta.model_name
         app_label = self.model._meta.app_label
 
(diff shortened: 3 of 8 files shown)


JSv4:
Thanks for the thorough review. Both flagged questions are about deliberate, pre-existing behavior, so no code change — but to put the rationale on record:

**Q1 — a group-level message grant is visible even when the parent conversation isn't.** This is the existing message-level access path, not new in this PR. The authenticated branch already ORs the *user*-level direct message grant with conversation visibility:

```python
return queryset.filter(conversation_visible | base_conditions | moderator_conditions)
#                       ^ base_conditions already contains has_permission (user-level message grant)
```

So a message shared directly (without sharing its conversation) was already visible to a user-level grantee. Item 3 only extends that same OR to *group* grants, for filter/check parity. Switching OR→AND would change established user-level semantics and drop legitimate direct-share access — out of scope for this audit follow-up. (Whether the frontend lets you navigate to the parent conversation is a separate UI concern; the backend correctly exposes what was explicitly shared.)

**Q2 — anonymous subquery cost.** The new anonymous path runs the *identical* subquery the long-standing authenticated path already runs (`Conversation.objects.visible_to_user(user).values_list("id")` → `conversation_id__in`). It's not a new cost class — both pay it deliberately to keep conversation visibility the single source of truth. I intentionally didn't add a perf note to only the anonymous branch, since that would imply it's special when it's the same query the authenticated branch has always used.

Minor: the redundant `public_corpus.delete()` in the anon test mirrors the existing `test_anonymous_user_sees_threads_on_public_corpus` in the same file; happy to drop it if you'd prefer.

---

JSv4:
Thanks — incorporated the one actionable item and clarifying the rest:

- **Strengthened `test_missing_conversation_group_table_falls_back_gracefully`** (commit `b7d8744`): it now also asserts that a conversation reachable via a *user-level* grant **still appears** under the simulated missing-group-table condition, pinning the "separate try blocks → a missing group table never discards already-resolved user grants" guarantee — not just the negative case.
- **Perf / `(conversation_type, is_public)` index (item 2):** no new index need is introduced by this PR. `ConversationQuerySet.visible_to_user` already compiled to this `conversation_id IN (SELECT ...)` subquery for the authenticated message path and for anonymous *conversation* queries; item 2 only routes anonymous *messages* through the same pre-existing path. Any index tuning here is a pre-existing consideration, not introduced by this change.
- **Message-level codename shape (item 3)** and the **`addCleanup` nit:** intentional/pre-existing as you noted — no change, consistent with the existing `has_permission` shape and the file's existing teardown pattern respectively.

---

JSv4:
Addressed in `305bd2c`:

- ✅ **Added `test_conversation_group_grant_makes_messages_visible`** — pins the common path (a group grant on the *conversation* surfaces its *messages* via the `conversation_visible` inheritance branch), distinct from the existing direct message-level grant test.
- ✅ **Trimmed the item-7 comment** to the invariant ("nullable config → resolve the `{}` fallback once before reading any key"); the historical detail lives in the changelog fragment.

Two points I'm intentionally **not** changing:

- **`order_by()` strip on the anonymous subquery:** the pre-existing *authenticated* branch immediately below runs the **identical** `Conversation.objects.visible_to_user(user).values_list("id")` subquery without stripping ordering, and has done so in production. Adding `.order_by()` to only the anonymous branch would make the two identical subqueries visibly asymmetric for a purely cosmetic gain. If the ordering-in-subquery noise is worth removing, it should be done to both in a dedicated cleanup, not lopsidedly here.
- **`id__in` vs `pk__in`:** the new message group branch already mirrors the existing user-level branch exactly — both use `Q(id__in=<subquery>)` for the positive case and `Q(pk__in=[])` for the `except LookupError` fallback. There's no new inconsistency introduced.

The `setUpClass`/`setUpTestData` and longer-comment notes match the existing conventions in this test file, so I've left them for consistency.

---

JSv4:
Addressed in `8c2c435`:

- ✅ **Strengthened `test_missing_message_group_table_falls_back_gracefully`** with the symmetric positive assertion — a *user-level* message grant now provably survives the simulated missing group table, matching the conversation counterpart.

Intentionally not changing (evaluated, declined):

- **App-label hard-coding in `ChatMessageQuerySet`:** the new group line mirrors the **adjacent pre-existing user-level lookup**, which hard-codes both `"conversations"` *and* the model name. Deriving only the app label would be half-measure; fully deriving (like `ConversationQuerySet`) means refactoring the pre-existing user-level lookup — cosmetic scope creep for a security PR. The values are correct.
- **Redundant `tearDown`:** matches the explicit-delete pattern used by the other two test classes in this same file; removing only this one would be locally inconsistent.
- **Codename asymmetry / long comments:** the asymmetry is already documented in the method comment, and the item-7 comment was trimmed in `305bd2c`; the remaining `conversations/models.py` comments carry security-parity rationale that justifies their length.

This is my last substantive change — all three reviews approved with no blockers, and the remaining notes are nitpicks I've evaluated. I'll keep monitoring CI to merge-readiness.

---

PR #2033, opened 2026-06-20
Fix enrichment linking/provenance robustness (#1996)

### LINKED ISSUE #1996: enrichment_service.py: linking/provenance robustness (pre-existing)
Three pre-existing bugs in `opencontractserver/enrichment/services/enrichment_service.py`, surfaced during the #1990 review. **Not introduced by #1990** — they live in the original `apply()` / `_link_external()` code, which that PR did not modify. Filing separately to keep scope clean.

- [ ] **`apply()` marks the Analysis COMPLETED before `_link_external()` runs.** `analysis.status = COMPLETED` is saved, then `_link_external()` runs outside the try/except. If `_link_external` raises (it does two `bulk_update`s over potentially thousands of rows), the exception propagates to the caller while the `Analysis` provenance row is permanently `COMPLETED` with `law_references_linked=0`. **Fix:** move the status save to after `_link_external()` returns, or extend the try/except to cover it.

- [ ] **`path_corpus_cache = dict(...)` keeps only the last corpus for a multi-corpus document.** `dict()` of `DocumentPath...values_list("document_id","corpus_id")` silently retains the last row when an authority document has current paths in >1 corpus → nondeterministic `target_corpus_id` (Postgres storage order) and a mention `link_url` that 404s for the other corpus. **Fix:** deterministic `.order_by("corpus_id")`, or scope to the specific authority corpus being linked.

- [ ] **`refs` queryset iterated twice as a lazy queryset.** The first loop builds `target_cache`, the second assigns targets; rows inserted between the two evaluations by a concurrent `apply()` on the same corpus appear in pass 2 but not the cache → silently skipped. **Fix:** materialize `refs = list(CorpusReference.objects.filter(...).select_related(...))` once before the first loop.

JSv4:
Closes #1996.

Fixes the three pre-existing bugs flagged in #1996, all in
`opencontractserver/enrichment/services/enrichment_service.py`. Each was confirmed
still present on `main`. The changes are localized to `apply()` and
`_link_external()` — no architectural changes.

## 1. `Analysis` stranded `COMPLETED` when linking failed
`apply()` stamped the provenance `Analysis` `COMPLETED` and saved it **before**
calling `_link_external()`, which issues two `bulk_update`s over potentially
thousands of rows. If linking raised, the exception propagated to the caller while
the `Analysis` row stayed permanently `COMPLETED` with `law_references_linked=0`.

**Fix:** `_link_external()` now runs **inside** `apply()`'s `try` (after
`writer.write`, before the `COMPLETED` stamp), so a linking failure marks the
`Analysis` `FAILED` and re-raises. The direct agent-tool path (which creates its
own `Analysis`) was fully affected; this also keeps the analyzer-framework path
consistent.

## 2. Nondeterministic / non-navigable `target_corpus_id` for multi-corpus authority documents
`_link_external()` built `path_corpus_cache` with
`dict(DocumentPath…values_list("document_id", "corpus_id"))`. When an authority
document has current paths in **more than one corpus**, `dict()` silently keeps
whatever row Postgres returns last → a nondeterministic `target_corpus_id` and a
mention `link_url` that 404s for the audience.

**Fix:** the target corpus is now chosen **deterministically *and* navigably** —
prefer a corpus the citing corpus's *audience floor* can actually open
(anonymous for a public corpus, the creator otherwise), breaking ties on the
lowest `corpus_id`; fall back to the lowest `corpus_id` so a resolvable target
never loses its corpus. Still one query for the paths plus one for the
visible-corpus set — no per-target N+1. (This addresses both options the issue
suggested: deterministic ordering *and* scoping to a navigable corpus.)

## 3. `refs` queryset iterated twice while lazy
The `CorpusReference` queryset was walked once to build the target cache and again
to promote/demote. Rows inserted between the two evaluations by a concurrent
`apply()` on the same corpus appeared in pass 2 but not the cache → silently
skipped.

**Fix:** materialize it once with `list(...)` before the first loop.

## Tests
Added three regression tests in `opencontractserver/tests/test_enrichment_linking.py`:

- `test_apply_marks_analysis_failed_when_linking_raises` — bug 1; reliably fails
  on the original code (status was `COMPLETED`).
- `test_link_target_corpus_is_deterministic_for_multi_corpus_authority` — bug 2
  determinism; reliably fails on the original code.
- `test_link_target_corpus_prefers_audience_visible_corpus` — bug 2 navigability;
  a private mirror corpus with the *lower* `corpus_id` is correctly skipped in
  favour of the public authority corpus, guarding against a naive "lowest-id only"
  fix.

Bug 3 is a defensive materialization against a true insert race that can't be
cleanly simulated in a unit test; its normal-path correctness is covered by the
existing linking suite.

All 13 tests in `test_enrichment_linking.py` pass, plus 150 adjacent
enrichment/authority/governance tests run green (`test_enrichment_writer`,
`test_enrichment_authorities`, `test_enrichment_analyzer_integration`,
`test_crawl_authorities`, `test_governance_graph`, `test_enrichment_backfill`,
`test_enrichment_tools`, `test_enrichment_run_mutation`,
`test_corpus_reference_model`, `test_corpusreference_backfill`,
`test_corpusreference_classification`). `black`, `isort`, and `flake8` are clean.

Changelog fragment: `changelog.d/1996-enrichment-linking-robustness.fixed.md`.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Fg68y5wUE2y5vmwbVW8A3x)_

### DIFF
--- a/changelog.d/1996-enrichment-linking-robustness.fixed.md
+++ b/changelog.d/1996-enrichment-linking-robustness.fixed.md
@@ -1,0 +1,31 @@
+- **Enrichment linking/provenance robustness — three pre-existing bugs in
+  `opencontractserver/enrichment/services/enrichment_service.py` (issue #1996).**
+  - **Analysis stranded `COMPLETED` when linking failed.** `apply()` stamped the
+    provenance `Analysis` `COMPLETED` and saved it *before* calling
+    `_link_external()` (which issues two `bulk_update`s over potentially
+    thousands of rows). A failure there propagated to the caller while the
+    `Analysis` row stayed permanently `COMPLETED` with `law_references_linked=0`.
+    Fix: `_link_external()` now runs inside `apply()`'s `try` (after
+    `writer.write`, before the `COMPLETED` stamp), so a linking failure marks the
+    `Analysis` `FAILED` and re-raises (`enrichment_service.py:444-458`).
+  - **Nondeterministic / non-navigable `target_corpus_id` for multi-corpus
+    authority documents.** `_link_external()` built `path_corpus_cache` with
+    `dict(DocumentPath…values_list("document_id","corpus_id"))`, which silently
+    kept whatever row Postgres returned last when an authority document had
+    current paths in more than one corpus — yielding a nondeterministic
+    `target_corpus_id` and a mention `link_url` that 404s for the other corpus.
+    Fix: the target corpus is now chosen deterministically *and* navigably —
+    prefer a corpus the citing corpus's audience can actually open, breaking ties
+    on the lowest `corpus_id` (one query for paths + one for the visible-corpus
+    set; no per-target N+1) (`enrichment_service.py:620-655`).
+  - **`refs` queryset iterated twice while lazy.** The `CorpusReference`
+    queryset was walked once to build the target cache and again to
+    promote/demote; rows inserted between the two evaluations by a concurrent
+    `apply()` on the same corpus appeared in pass 2 but not the cache and were
+    silently skipped. Fix: materialize it once with `list(...)` before the first
+    loop (`enrichment_service.py:604-613`).
+  - Regression coverage added in
+    `opencontractserver/tests/test_enrichment_linking.py`
+    (`test_apply_marks_analysis_failed_when_linking_raises`,
+    `test_link_target_corpus_is_deterministic_for_multi_corpus_authority`,
+    `test_link_target_corpus_prefers_audience_visible_corpus`).
--- a/opencontractserver/enrichment/services/enrichment_service.py
+++ b/opencontractserver/enrichment/services/enrichment_service.py
@@ -662,12 +662,21 @@ class EnrichmentService:
                 continue
             target = target_cache.get(key)
             if target is not None:
+                target_corpus_id = path_corpus_cache.get(target.id)
+                if target_corpus_id is None:
+                    # find_authority_target only returns a document with a
+                    # current, non-deleted path, so target_corpus_id is normally
+                    # present; it can go missing only if that path is removed
+                    # between resolution and the path query above. Don't promote
+                    # to RESOLVED without a navigable target_corpus — that would
+                    # render the broken link this pass exists to prevent.
+                    continue
                 if (
                     ref.target_document_id != target.id
                     or ref.resolution_status != C.STATUS_RESOLVED
                 ):
                     ref.target_document = target
-                    ref.target_corpus_id = path_corpus_cache.get(target.id)
+                    ref.target_corpus_id = target_corpus_id
                     ref.resolution_status = C.STATUS_RESOLVED
                     # bulk_update bypasses auto_now — stamp ``modified``.
                     ref.modified = now
--- a/opencontractserver/enrichment/services/enrichment_service.py
+++ b/opencontractserver/enrichment/services/enrichment_service.py
@@ -444,13 +444,18 @@ class EnrichmentService:
         writer = EnrichmentWriter(corpus, creator_id, analysis=analysis)
         try:
             res = writer.write(resolutions)
+            # Cross-corpus linking is part of a successful apply and can raise
+            # (two bulk_updates over potentially thousands of rows), so it must
+            # run INSIDE the try. Stamping the Analysis COMPLETED before it ran
+            # left a permanent COMPLETED provenance row with
+            # law_references_linked=0 whenever _link_external failed (#1996).
+            link = self._link_external(user, corpus)
         except Exception:
             analysis.status = JobStatus.FAILED.value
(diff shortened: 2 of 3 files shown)


JSv4:
Thanks for the review — addressed in `2f940ee`, along with the `linter` (mypy) failure.

**Issue 1 (null `target_corpus`)** — added a guard in `_link_external()`: a resolved target with no entry in `path_corpus_cache` is now skipped rather than promoted to `RESOLVED` with a null `target_corpus` (which `_restamp_mention_links` would otherwise render as a link into the *citing* corpus). In normal flow this is unreachable — `find_authority_target` only returns a document that has a current, non-deleted path, which the path query then re-finds — so it's defensive against a path removed mid-pass.

I kept the annotation as `dict[int, int]` rather than `dict[int, int | None]`: the dict *values* are genuinely always `int` (`corpus_id` is a non-null FK), so `int | None` would be less accurate. The `None` you noted comes from `dict.get()` on a missing key, not from a `None` value — and mypy already models that (`dict[int, int].get(k)` is typed `int | None`), so the call site stays correctly narrowed.

**Issue 2 (test premise)** — added the explicit `DocumentPath` existence assertion. One correction: the test wouldn't actually pass vacuously if the bootstrapper stopped creating the auth-corpus path — `target_corpus_id` would become `second.id`, but `min(auth["corpus_id"], second.id)` is `auth["corpus_id"]` (auth is created first), so the assertion would *fail*. The guard is still worth it for a clearer failure message, so it's in.

**Nitpicks** — left the `apply()` comment as-is (it explains the non-obvious provenance/`bulk_update` WHY, which your Positives section agreed meets the bar) and `_doc_id` over `_` (slightly clearer at the unpack site). Both genuinely trivial.

CI re-running now.

---

JSv4:
Thanks — good catch on the TOCTOU demote gap. Decided to close it now rather than defer, in `10d9bfb`.

Rather than duplicate the demote block inside the guard, I restructured the loop slightly: compute `target_corpus_id` up front, promote only when there's a navigable target (`target is not None and target_corpus_id is not None`), and otherwise fall through to the **existing** `elif` demote branch. Net effect:

- **Promote** — unchanged (navigable target).
- **No target** — unchanged (demote if currently resolved).
- **Target found, EXTERNAL ref, no current path** — unchanged (stays EXTERNAL).
- **Target found, RESOLVED ref, no current path (the TOCTOU case)** — now **demoted** instead of left stale. ← the gap you flagged.

So the only behavioral change is that one case, and there's a single demote block (no duplication). Confirmed `path_corpus_cache` can only miss a resolved target when *all* its current paths are gone (the cache is seeded on first sight regardless of audience-visibility), so a miss genuinely means "no navigable path" → demote is correct.

Added `test_resolved_ref_demoted_when_target_loses_navigable_path` for the demote case and kept the EXTERNAL-stays-EXTERNAL test. 15 linking tests + 79 adjacent enrichment/authority/governance tests pass; mypy/black/flake8 clean.

On the other notes — no action needed: the Bug 1 re-raise semantics are intentional (FAILED is correct since `apply()` didn't return success); `visible_to_user(None)` returning the published authority corpus is verified by `test_link_target_corpus_prefers_audience_visible_corpus` (it asserts the public corpus wins and the link resolves); and I've left the `#1996` breadcrumbs in the comments since they point readers straight at the rationale.

---

JSv4:
Thanks for the review (and the streamed parts). Net: no code change needed — here's the status on the three non-blocking nits, two of which are already in the PR:

1. **`writer.write()` not rolled back if `_link_external` fails** — intentional, and matches the earlier review's read. `writer.write()` commits inside its own `transaction.atomic()` (`writer.py:330`) and `apply()` has no outer atomic, so on a linking failure those rows persist and only the `Analysis` is stamped `FAILED`. That's safe by design: enrichment is idempotent and `_link_external` / `link_external_references` are re-runnable, so a re-run reconciles the partial state; the `FAILED` stamp records that the run didn't complete. Leaving as-is.

2. **Two-pass loop comment** — already documented at the `refs = list(...)` site (*"walked twice — build the target cache below, then promote/demote"*) plus *"Resolve each distinct key once under the audience floor"* on pass 1. Given the earlier note flagging comment verbosity, I'd rather not expand further.

3. **Two-visible-corpora lowest-id tie-break test** — already covered by `test_link_target_corpus_is_deterministic_for_multi_corpus_authority`: both the auth corpus and the mirror are owned by (visible to) the audience, and it asserts `target_corpus_id == min(auth_corpus_id, second_id)`.

Since the last review round I also closed the TOCTOU demote gap (`10d9bfb`), so the PR now carries the three #1996 fixes + the demote edge case, with five regression tests; mypy/black/isort/flake8 clean. Ready to merge from my side.

---

PR #2034, opened 2026-06-20
Address PR #2008 review: enrichment + notification fixes (#2029)

### LINKED ISSUE #2029: ## Code Review — PR #2008
## Code Review — PR #2008

**Stacking note**: This review focuses on the delta above #2001 (the Phase 5 base). The overall approach is sound and the PR is very well-prepared — good test coverage, explicit changelog fragments, adversarial review artefacts already documented. Below are the issues I think should be addressed before merge, followed by smaller nits.

---

### Issues

**1. Race condition in signal idempotency guard (`opencontractserver/notifications/signals.py:166-167`)**

```python
if Notification.objects.filter(analysis=instance, notification_type=ntype).exists():
    return
# ... then ...
notification = Notification.objects.create(...)
```

Two concurrent `Analysis.post_save` signals (e.g. a Celery worker + an in-process save during a test tear-down) can both pass the `exists()` check before either `create()` commits, producing duplicate notifications. Replace the check+create pair with `get_or_create`:

```python
notification, created = Notification.objects.get_or_create(
    analysis=instance,
    notification_type=ntype,
    defaults={
        "recipient_id": instance.creator_id,
        "data": {...},
    },
)
if not created:
    return
```

**2. Partial-success semantics not surfaced to callers (`config/graphql/enrichment_mutations.py:152-161`)**

When `run_enrichment` dispatches successfully but `run_crawl` fails, the mutation returns `ok=False` + the enrichment `Analysis` row in `analyses`. The enrichment job is actually running — the caller needs to know this or they may retry both jobs. Two options: (a) return `ok=True` + a non-fatal `message` when enrichment succeeded but crawl failed; or (b) document the partial-success contract on the mutation docstring more explicitly and add a dedicated `partialFailure: Boolean` field. Option (a) is simpler and matches what the UI actually needs (show the running job).

**3. Silent `first: 50` cap in `GET_CORPUS_ANALYSES` (`frontend/src/graphql/queries.ts:1072`)**

```graphql
analyses(... first: 50) { ... }
```

A corpus with more than 50 enrichment runs will silently lose history. The component has no way to know it's truncated. Either surface a `totalCount` field (the existing `CountableConnection` pattern already supports this) and show a "showing N of M" label, or document the limit clearly and add a corresponding test that asserts the cap is intentional.

**4. Invalid reference-type codes silently dropped without user feedback (`config/graphql/enrichment_mutations.py:124-127`)**

```python
valid_types = [t for t in ref_types if t in C.ALL_REFERENCE_TYPES]
if valid_types:
    input_data["types"] = valid_types
```

If a caller passes `["BOGUS"]`, `valid_types` is empty, `types` is not set, and the enrichment runs against ALL reference types — a very different outcome than requested. At minimum, log a warning. Better: return a validation error listing the unknown codes so the caller knows their request was modified.

**5. Missing WHY comment on `useEffect` deps exclusion (`frontend/src/components/admin/enrichment/useOptimisticRows.ts:38`)**

```ts
}, [jobs]); // eslint-disable-line react-hooks/exhaustive-deps
```

`optimistic` is intentionally excluded to avoid an infinite loop (the effect prunes `optimistic` → triggers the effect again → ...). The suppression is correct but a future reader won't know why. Add a single-line comment: `// optimistic excluded: including it would re-trigger when the prune below updates state`

---

### Minor nits

**6. Signal fires an extra DB query on every non-enrichment `Analysis.save()`**

The `Analyzer.objects.filter(pk=instance.analyzer_id, task_name__in=_ENRICHMENT_TASK_NAMES)` query runs on every `Analysis.post_save`, including doc-analysis and extract `Analysis` rows. This is fine at current scale (Analyzer table is tiny and query is PK-indexed), but a short-circuit like `if instance.status not in _ANALYSIS_STATUS_NOTIFICATION` at the top of the handler saves the I/O for the majority of saves. You already check `ntype = _ANALYSIS_STATUS_NOTIFICATION.get(instance.status)` before the analyzer query, so this short-circuit just moves the return one line earlier — the logic is already correct, this is purely about comment accuracy vs. the "Guard on analyzer_id FIRST (cheap, no I/O)" claim.

**7. `elapsedLabel` always shows raw seconds (`frontend/src/components/admin/enrichment/EnrichmentJobList.tsx:195-201`)**

A crawl that runs for 127 seconds displays as `127s`. Consider a small formatter: `< 60s → "Ns"`, `≥ 60s → "Nm Ns"`. Not a blocker, but authority crawls can take several minutes and the raw seconds makes the elapsed column hard to scan at a glance.

**8. `NumberInput` allows fractional values (`frontend/src/components/admin/enrichment/EnrichmentRunner.tsx`)**

`type="number"` + no `step` attribute means `3.14` is a valid entry. The mutation input schema defines these as `graphene.Int`, so Graphene will truncate/error. Add `step={1}` to each numeric input to close the gap at the UI layer.

**9. `analysis` FK on `Notification` uses `CASCADE` — confirm intent (`opencontractserver/notifications/migrations/0006_...py:20`)**

Deleting an `Analysis` row cascades to its notification rows. This is probably intentional (no point keeping "analysis 42 is RUNNING" if analysis 42 no longer exists), but it's worth a one-line comment in the migration — the original notification models all use `SET_NULL` or `PROTECT`. If there's a case where an `Analysis` gets deleted but the notification should survive (audit log), this will silently destroy it.

**10. `user_can` in `analysis_lifecycle_service.py` doesn't forward `request` for Tier-2 cache**

```python
if require_corpus_update and not corpus_obj.user_can(user, PermissionTypes.UPDATE):
```

The `request` is passed to `start_document_analysis` but not forwarded to `user_can`. Per CLAUDE.md the `request` parameter engages the Tier-2 permission cache in GraphQL resolvers. This is a performance miss (extra DB hit) rather than a correctness bug — and is consistent with how the document check is done just above it in the same function — but flagging for awareness.

---

### What's well done

- The relay type-prefix validation in `RunCorpusEnrichmentMutation` is exactly the right defense; the changelog entry for it is a good artifact.
- The `AnalysisLifecycleService.require_corpus_update` opt-in flag is a clean way to gate write-scoped operations without polluting the general analysis path.
- Scoping the signal to `_ENRICHMENT_TASK_NAMES` (module-level tuple, built once at import) is the right performance call.
- The `classify_canonical_key` ladder and `namespace_classification_cache` batch-fetch correctly align the read (discover) and write (persist) classification paths — gap-4 was a subtle invariant violation.
- The `whole-act` fallback (`":" not in canonical_key`) in `find_authority_target` is correctly narrow-scoped so it can't silently resolve a genuinely missing section citation to a wrong section.
- `EnrichmentPanel` extracted as a separate component to ensure hooks are called unconditionally is the right hook-rules fix.
- The `useEnrichmentJobs` unit test covering the 5 notification-routing cases (matching corpus, missing corpus_id, different corpus, wrong type, undecodable id) is exactly the right set.

_Originally posted by @claude[bot] in https://github.com/Open-Source-Legal/OpenContracts/issues/2008#issuecomment-4758814061_

JSv4:
Closes #2029.

Issue #2029 is the code review of PR #2008 (now merged). Every item it raised was still present on `main`; this PR addresses all of them — the 5 "Issues" and the 5 "Minor nits" — with surgical changes scoped to the enrichment feature and the analysis-notification signal.

## Review items → fixes

| # | Review item | Fix |
|---|-------------|-----|
| 1 | Race in signal idempotency guard (`notifications/signals.py`) | Replaced check-then-create with atomic `get_or_create`, backed by a **new partial unique constraint** `uniq_notification_per_analysis_type` on `(analysis, notification_type)` (`condition=analysis NOT NULL`). Migration `0007` de-dupes pre-existing rows first. `get_or_create` alone isn't race-safe without the constraint, so both are needed. |
| 2 | Partial-success not surfaced (`enrichment_mutations.py`) | When enrichment dispatches but the crawl fails, returns `ok=True` + the running row + a non-fatal message (was `ok=False`). The Runner shows it as a warning toast. Crawl-only failures still return `ok=False`. |
| 3 | Silent `first: 50` cap (`queries.ts`) | Query now requests `totalCount`; `EnrichmentJobList` shows "Showing the N most recent of M runs." when truncated. Threaded `useEnrichmentJobs` → `useOptimisticRows` → both consumers. |
| 4 | Invalid reference-types silently dropped (`enrichment_mutations.py`) | Unknown codes are now **rejected** (`ok=False`, names the bad codes) instead of filtered — an all-bogus list previously fell through to scanning ALL types. |
| 5 | Missing WHY comment on `useEffect` deps (`useOptimisticRows.ts`) | Added the comment explaining `optimistic` is excluded to avoid a prune→re-trigger loop. |
| 6 | Extra DB query / comment accuracy (`signals.py`) | The status+creator guards already short-circuit before the analyzer query; reworded the comment so the "cheap, no I/O" claim is accurate (logic already correct). |
| 7 | `elapsedLabel` raw seconds (`EnrichmentJobList.tsx`) | Multi-minute durations now render `Nm Ns` / `Nh Nm` (e.g. `127s` → `2m 7s`); sub-minute unchanged (`47s`). |
| 8 | `NumberInput` allows fractional values (`EnrichmentRunner.tsx`) | Added `step={1}` to the crawl-bound inputs (plus an integer guard in `handleRun`). |
| 9 | `analysis` FK CASCADE intent (migration/model) | Documented the deliberate CASCADE on `Notification.analysis` (transient UI signal, not an audit log) in `models.py`. |
| 10 | `user_can` not forwarding `request` (`analysis_lifecycle_service.py`) | Forwarded `request` to the corpus UPDATE check so it shares the Tier-2 permission cache. |

## Testing

- **Backend**: `test_notifications.py` + `test_enrichment_run_mutation.py` pass (unique-constraint idempotency incl. partial/NULL behaviour, invalid-type rejection, partial-success, crawl-only failure, uniform `partial=False`). `makemigrations --check` reports no missing `notifications` migration (model Meta ↔ migration `0007` agree); the GraphQL service-layer architecture invariant still passes.
- **Frontend**: `tsc --noEmit` clean; `formatters` + `useEnrichmentJobs` Vitest pass; `EnrichmentJobList` / `CorpusEnrichmentCard` / `EnrichmentRunner` CT suites pass, including multi-minute-elapsed, truncation-note (show/hide/optimistic-count), partial-success warning toast, fractional-input rejection, and the `ok=false` error-toast path.

## Notes

- The migration adds a partial unique index; the cleanup step keeps it safe against any pre-existing duplicates.
- No `CHANGELOG.md` edit — two `changelog.d/2029-enrichment-review.{fixed,changed}.md` fragments added per the repo convention.
- An unrelated, pre-existing `makemigrations` discrepancy on `annotations.authorityfrontier.discovery_state` (introduced by #2008's merge) is **not** touched here.

### DIFF
--- a/changelog.d/2029-enrichment-review.changed.md
+++ b/changelog.d/2029-enrichment-review.changed.md
@@ -1,0 +1,17 @@
+- **The enrichment job list no longer silently truncates history (issue #2029 —
+  PR #2008 review).** `GET_CORPUS_ANALYSES` now requests `totalCount`
+  (`frontend/src/graphql/queries.ts`) and `EnrichmentJobList`
+  (`frontend/src/components/admin/enrichment/EnrichmentJobList.tsx`) renders a
+  "Showing the N most recent of M runs" note whenever the server holds more
+  analyses than the `first: 50` page cap returns. `totalCount` is threaded
+  through `useEnrichmentJobs` → `useOptimisticRows` to both the Admin panel and
+  the corpus enrichment card.
+- **Multi-minute enrichment elapsed times now read as `Nm Ns` / `Nh Nm`
+  (`frontend/src/components/admin/enrichment/EnrichmentJobList.tsx`).** A crawl
+  that ran 127 seconds previously rendered as the unscannable "127s"; it now
+  renders "2m 7s". Sub-minute durations are unchanged ("47s").
+- **Crawl-bound number inputs reject fractional values
+  (`frontend/src/components/admin/enrichment/EnrichmentRunner.tsx`).** Added
+  `step={1}` to the Advanced crawl-bound inputs so the browser flags
+  non-integer entries at the UI layer before they reach the `graphene.Int`
+  mutation schema.
--- a/changelog.d/2029-enrichment-review.fixed.md
+++ b/changelog.d/2029-enrichment-review.fixed.md
@@ -1,0 +1,33 @@
+- **Closed a duplicate-notification race in the analysis-status signal
+  (`opencontractserver/notifications/signals.py`, issue #2029 — PR #2008
+  review).** The handler guarded duplicate `ANALYSIS_RUNNING/COMPLETE/FAILED`
+  notifications with a `Notification.objects.filter(...).exists()` check
+  followed by a separate `.create()`. Two concurrent `Analysis.post_save`
+  signals (e.g. a Celery worker plus an in-process save) could both pass the
+  `exists()` check before either committed, leaking duplicate rows. The
+  check-then-create pair is now a single atomic `get_or_create`, backed by a new
+  partial unique constraint `uniq_notification_per_analysis_type` on
+  `(analysis, notification_type)` (`condition=analysis NOT NULL`) added in
+  `opencontractserver/notifications/migrations/0007_notification_analysis_idempotency_constraint.py`.
+  The migration de-dupes any pre-existing rows (keeping the earliest) before
+  adding the constraint; the partial condition leaves the many analysis-less
+  notification types unconstrained.
+- **`runCorpusEnrichment` now rejects unknown `reference_types` instead of
+  silently dropping them (`config/graphql/enrichment_mutations.py`).** A request
+  whose `reference_types` contained only unrecognised codes previously filtered
+  to an empty list, left `types` unset, and ran enrichment against ALL reference
+  types — the opposite of the caller's intent. The mutation now returns
+  `ok=False` naming the unknown code(s) and dispatches nothing.
+- **`runCorpusEnrichment` surfaces partial success
+  (`config/graphql/enrichment_mutations.py`).** When the enrichment analyzer
+  dispatches but the authority crawl fails, the mutation now returns `ok=True`
+  with the already-running enrichment row and a non-fatal message — previously
+  it returned `ok=False` with the enrichment row mixed in, leading callers to
+  treat the whole request as failed and retry (double-dispatching enrichment).
+  The Runner (`frontend/src/components/admin/enrichment/EnrichmentRunner.tsx`)
+  surfaces the message as a warning toast.
+- **Forwarded `request` to the corpus UPDATE check in
+  `AnalysisLifecycleService.start_document_analysis`
+  (`opencontractserver/analyzer/services/analysis_lifecycle_service.py`).** The
+  `user_can(user, UPDATE)` gate now shares the Tier-2 permission cache when
+  called from a GraphQL mutation, removing a redundant permission DB hit.
--- a/config/graphql/enrichment_mutations.py
+++ b/config/graphql/enrichment_mutations.py
@@ -122,6 +122,7 @@ class RunCorpusEnrichmentMutation(graphene.Mutation):
             # distinguish "malformed id" from "exists but not visible" (IDOR).
             return RunCorpusEnrichmentMutation(
                 ok=False,
+                partial=False,
                 message="Resource not found or you do not have permission.",
                 analyses=[],
             )
@@ -129,6 +130,7 @@ class RunCorpusEnrichmentMutation(graphene.Mutation):
         if not run_enrichment and not run_crawl:
             return RunCorpusEnrichmentMutation(
                 ok=False,
+                partial=False,
                 message="Select at least one job (runEnrichment or runCrawl).",
                 analyses=[],
             )
@@ -150,6 +152,7 @@ class RunCorpusEnrichmentMutation(graphene.Mutation):
                 if unknown:
                     return RunCorpusEnrichmentMutation(
                         ok=False,
+                        partial=False,
                         message=(
                             "Unknown reference type(s): "
                             + ", ".join(unknown)
@@ -180,6 +183,7 @@ class RunCorpusEnrichmentMutation(graphene.Mutation):
             if not res.ok:
                 return RunCorpusEnrichmentMutation(
                     ok=False,
+                    partial=False,
                     message=res.error,
                     analyses=[],
                 )
@@ -234,6 +238,7 @@ class RunCorpusEnrichmentMutation(graphene.Mutation):
                     )
                 return RunCorpusEnrichmentMutation(
                     ok=False,
+                    partial=False,
                     message=res.error,
                     analyses=[],
                 )
--- a/config/graphql/enrichment_mutations.py
+++ b/config/graphql/enrichment_mutations.py
@@ -128,16 +128,32 @@ class RunCorpusEnrichmentMutation(graphene.Mutation):
         created = []
 
         if run_enrichment:
-            analyzer = EnrichmentService.get_or_create_analyzer(user.id)
             input_data: dict[str, Any] = {
                 "use_llm": bool(getattr(options, "use_llm_tier", False) or False),
             }
             ref_types = getattr(options, "reference_types", None)
             if ref_types:
-                valid_types = [t for t in ref_types if t in C.ALL_REFERENCE_TYPES]
-                if valid_types:
-                    input_data["types"] = valid_types
+                # Reject unknown codes rather than silently dropping them. If we
+                # filtered to an empty ``valid_types`` and left ``types`` unset,
+                # the analyzer would fall through to scanning ALL reference types
+                # — the opposite of the caller's intent. Surface the bad codes so
+                # the caller knows their request was rejected, not modified.
+                unknown = [t for t in ref_types if t not in C.ALL_REFERENCE_TYPES]
+                if unknown:
+                    return RunCorpusEnrichmentMutation(
+                        ok=False,
+                        message=(
+                            "Unknown reference type(s): "
+                            + ", ".join(unknown)
+                            + ". Valid types: "
+                            + ", ".join(C.ALL_REFERENCE_TYPES)
+                            + "."
+                        ),
+                        analyses=[],
+                    )
+                input_data["types"] = list(ref_types)
 
+            analyzer = EnrichmentService.get_or_create_analyzer(user.id)
             logger.info(
                 "RunCorpusEnrichmentMutation: dispatching enrichment analyzer "
                 "analyzer_pk=%s corpus_pk=%s user=%s",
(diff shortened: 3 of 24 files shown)


JSv4:
Thanks for the review — addressed in 60e706c.

**Issue 1 (fragile `"SUCCESS"` sentinel)** — Added an explicit `partial: Boolean` field to the `RunCorpusEnrichmentMutation` payload (set `True` on the enrichment-ok/crawl-failed path, `False` on clean success). The Runner now keys its warning toast off `payload.partial`, so the UI no longer couples to the message text. Asserted in the backend tests.

**Issue 2 (inline utility)** — Moved `formatElapsedSeconds` to `frontend/src/utils/formatters.ts` next to `formatDuration`, with dedicated unit tests in `formatters.test.ts`.

**Nit 3 (hardcoded codes)** — `test_valid_reference_types_are_accepted` now sources the list from `C.ALL_REFERENCE_TYPES`.

**Nit 5 (reverse docstring)** — Expanded `reverse_cleanup` to note the constraint/index is dropped automatically on reverse; only the one-time dedup is irreversible.

**Nit 4 (truncation note vs. optimistic rows)** — Left as-is. The note only renders when `totalCount > fetchedJobs.length`, and with a `first: 50` cap `fetchedJobs.length` is 50 whenever it's truncated — so the realistic case reads "Showing the 50 most recent of 73 runs" with at most one transient optimistic row appended (a 50-vs-51 discrepancy that the prune effect clears within ~1s of the refetch). The literal "1 of 73" example can't occur, since a 73-row corpus returns 50 server rows, not 1. Hiding the note while optimistic rows are present would make it flicker on every dispatch, which seemed worse than the brief off-by-one. Happy to revisit if you'd prefer it hidden during dispatch.

Validation: backend enrichment-mutation tests 12 passed; frontend `formatters` 44 passed (incl. 3 new) and the EnrichmentRunner + EnrichmentJobList CT suites 16 passed; tsc/prettier/black/isort/flake8 clean.

---

JSv4:
Thanks for the thorough second pass — addressed in 307efd5.

**Acted on:**
- **#1 `get_or_create` staleness** — added a `# NOTE` that `data` is the first-writer snapshot under a concurrent race (safe, since `notification_type` is one-per-status).
- **#2 `step={1}` not a hard block** — `handleRun` now rejects non-integers via `Number.isInteger` ("… must be a whole number") before any dispatch; covered by a new CT test that types `1.5` into Max depth.
- **#3 truncation count vs. optimistic rows** — the note now counts `sorted.length` (the rows actually rendered) and gates on the same value, so "N most recent" always matches what's on screen and stays below the server total; new CT test asserts the count includes an optimistic row.
- **Nit — negative `formatElapsedSeconds`** — clamps with `Math.max(0, …)`; unit test added.
- **Coverage gap — partial-success toast** — new CT test mounts a `ToastContainer` and asserts the warning surfaces (and a small `React.Children.only` gotcha in the wrapper is handled by wrapping the two children in a fragment).

**Already covered / no action (verified against the code):**
- **Mixed valid/invalid types test** — `test_invalid_reference_types_are_rejected` already sends a *mixed* list `["LAW", "INVALID_CODE"]`, so the any-unknown rejection path is exercised.
- **`reverse_cleanup` comment** — expanded in the previous commit (60e706c) to note the index drops automatically on reverse.
- **`analyses=[]` contract change** — not actually a contract change: in every `ok=False` return, `created` is empty (enrichment hadn't dispatched), so `analyses=[]` equals the old `analyses=created`. The only non-empty-on-failure case is now `ok=True, partial=true`.

Validation: EnrichmentRunner CT 9 passed, EnrichmentJobList CT 10 passed, formatters 45 passed; tsc/prettier/black/isort/flake8 clean.

---

JSv4:
Addressed the one actionable item in 04b289b: `partial=False` is now set explicitly on all five `ok=False` return paths, so the field is always a concrete boolean (never `null`). Locked the uniform contract with `assert data["partial"] is False` in the invalid-types and crawl-only-failure tests (12 passed; lint clean).

---

JSv4:
Both clarifications added in 75e6a84 (both confirmed intentional, comment-only):

- **`recipient_id` in `get_or_create` defaults** — intentional. Extended the inline `NOTE` to explain the idempotency key is exactly `(analysis, notification_type)` (to match the partial unique constraint), and the recipient is the stable analysis creator, so it's a value to set on create rather than a lookup key.
- **Empty `ref_types` list** — documented that an omitted field (`None`) and an explicitly empty list both mean "no restriction" (only a non-empty list is validated, equivalent to omitting the field). Left as a comment rather than a hard guard since `[]` ≡ omitted is the sensible interpretation and the UI never sends an empty list (LAW is always included).

Re: issue 3 (`useEffect` deps) — the `// eslint-disable-line react-hooks/exhaustive-deps` is already on the `}, [jobs]);` line, and the project's `lint` script is `prettier . --check` (no eslint), so nothing surfaces there.

---

JSv4:
Addressed both the review items and the codecov note in 0110251.

**Review:**
- **`Math.floor` in `formatElapsedSeconds`** — input is now floored (and clamped) so a fractional value can't render "2m 7.5s"; added a unit test for the fractional case.
- **Constraint single-recipient assumption** — added a note on the `UniqueConstraint` that the key is intentionally global (no `recipient`): analysis-status notifications always target the single analysis creator, and `recipient` would need to join the key for any future multi-recipient fan-out. (Comment-only — `makemigrations --check` confirms no new migration.)

**Codecov** — the flagged patch lines were the `ok=false` → error-toast branch in `EnrichmentRunner.tsx`. Added a CT test that mocks an `ok:false` response and asserts the backend message surfaces as an error toast (and no optimistic job is recorded). Worth noting `codecov/patch` was already green at 91.9% (non-gating), but the error path was genuinely worth covering.

Local validation: formatters 46 passed, EnrichmentRunner CT 10 passed, tsc/prettier/black/flake8 clean.

---

JSv4:
Good catch on the attribution footer — **removed** from the PR body. To be clear, that `_Generated by [Claude Code]_` line was auto-appended by the PR-creation tooling, not something I authored (CLAUDE.md's "never credit Claude" rule is exactly why I'd omitted it); it's now stripped.

On the minor nits:
- **No log on the duplicate `get_or_create` path** — reasonable suggestion, but I've left it out for now rather than spin a full ~30-min CI cycle for a one-line debug log on a rare skip path. Happy to add `logger.debug("Skipping duplicate %s notification for analysis %s", ntype, instance.pk)` if you'd like it in.
- **`Math.floor(Math.round(x))` redundancy** — intentional: the internal `Math.floor` was added in response to the prior review round so the utility is bulletproof for any future caller, independent of the one call site that already rounds. Harmless and deliberate.
- **`totalCount` typed `number` vs hook returning `number | null`** — the `number` reflects the Relay field; the hook's `| null` reflects the pre-load (`data === undefined`) state via `?? null`. Functionally correct, so leaving as-is.

Everything else in the verdict table is unchanged.

---

PR #2035, opened 2026-06-20
Consume LLM Singleton registry everywhere + add per-corpus LLM admin GUI

JSv4:
## Summary

An audit-and-complete pass on how LLMs are configured at runtime via the Singleton registry, ensuring **(1)** admins have a GUI surface to control it and **(2)** corpuses consume from it so — unless a tool/infra deliberately overrides — the configured Singleton LLM is used throughout a corpus.

The registry itself already existed and was mostly wired correctly: `resolve_model_spec()` walks `per-call → agent → corpus.preferred_llm → PipelineSettings.default_llm → Django settings`, and the agent factory consumes it for every agent path (document/corpus chat, extracts, deep research, agent mentions, corpus/thread actions). This PR closes the two remaining gaps the audit surfaced.

## Gap 1 — Corpus consumption (one runtime bypass)

Conversation-title generation (`config/websocket/consumers/unified_agent_conversation.py`) was the only runtime LLM path bypassing the registry. It called the OpenAI-only `SimpleLLMClient`, which **hardcoded `gpt-4o-mini`** and ignored both `corpus.preferred_llm` and the install-wide default — even though `self.corpus` was right there.

- Added `opencontractserver/llms/completions.py::agenerate_text` — a provider-agnostic (OpenAI/Anthropic/Google/Ollama), registry-backed one-shot completion helper that walks the same resolution chain as the agent factory and builds the model via the credential-aware `model_factory`.
- Routed title generation through it, passing the corpus default.
- **Removed the dead, registry-bypassing `SimpleLLMClient`** (`client.py`) and its `LLM_CLIENT_*` settings; updated docs + tests (no-dead-code).

## Gap 2 — Admin GUI for the per-corpus LLM

`Corpus.preferred_llm` was fully plumbed in the backend (mutation arg, `CorpusType`, serializer) but had **no frontend surface** — a ghost feature only settable via the GraphQL console/Django shell.

- New shared `LlmModelPicker` component (`frontend/src/components/common/LlmModelPicker.tsx`); the admin System Settings screen now consumes it too, de-duplicating its inline provider/model chip list.
- New **Language Model** card in the Corpus Settings panel. When the corpus has no override, it shows the **inherited install-wide default** so "leave empty = inherit" is explicit. Clearing sends `""`, which the backend normalises to `NULL`.
- Threaded `preferredLlm` through `RESOLVE_CORPUS_BY_SLUGS_FULL`, `UPDATE_CORPUS`, and the `CorpusType`/`RawCorpusType` TS types.
- Two slim, secret-free `@login_required` queries (`GET_LLM_PROVIDERS`, `GET_SYSTEM_DEFAULT_LLM`) feed the picker, so non-superuser corpus owners can use it without any credential exposure.

## Testing

- Backend: `opencontractserver/tests/test_llms_completions.py` pins the resolution chain (corpus beats settings default; explicit beats corpus; hard fallback; temperature omission; None output; max_tokens forwarding). Updated the consumer title tests to mock `agenerate_text` and cover corpus-preferred threading + empty/None fallback. `pyupgrade`/`black`/`isort`/`flake8` clean.
- Frontend: `LlmModelPicker` unit tests (5 cases) pass; `tsc --noEmit` clean; existing routing tests that hit the corpus-resolution query still pass (13/13); prettier clean.

## Permissioning note

The two new GraphQL queries are `@login_required` and request only non-secret fields (no `settingsSchema`/`currentValue`), so a corpus editor can populate the picker without ever seeing provider credentials. Override precedence is preserved end-to-end: explicit per-call model > per-agent > corpus > install-wide default > Django settings.


JSv4:
Thanks for the review. Addressed in `6002071`:

- **Disabled providers in the picker** — now filtered in `CorpusSettings` (`enabled !== false`) so users aren't offered admin-disabled providers. One clarification: the "runtime error at save time" premise is overstated — `validate_model_spec` (and the agent-build path) only check that the provider is *registered*, not *enabled*, so a disabled provider's model would actually still resolve. The filter is the right UX call regardless, but it's a polish item, not a latent crash. I scoped it to the user-facing corpus picker only and left System Settings showing all providers, since that's where an admin manages enablement (matching its existing behavior of not filtering on `enabled`).
- **Changelog slugs** — renamed `2030-*` → `2035-*`.
- **Empty-title branch** — added `test_title_generation_fallback_on_empty_title` covering the `return title or f"Conversation {uuid}"` fallback.

Skipped the combined-`GetCorpusLlmContext`-query nit: the two queries are single-responsibility, `cache-first` (free after first mount), and independently reusable, so I'd rather not couple provider-list and settings-default fetching into one bespoke query.

Separately, the `linter` CI failure was unrelated to the review — CI's `pyupgrade` rewrote `Optional[X]` → `X | None` in `completions.py` and flake8 then flagged the unused import; fixed in `0955911`.

---

JSv4:
Thanks — round-2 items addressed in `87ebc5f`:

- **Medium / `LlmProviderOption` duplication** — `LlmProvidersQueryResult` now reuses the exported `LlmProviderOption` (`Array<LlmProviderOption & { enabled?: boolean | null }>`) via an `import type`, so the projection and component can't drift.
- **`temperature` always injected** — `agenerate_text` now takes `temperature: float | None = 0.7` and only adds it to `model_settings` when set (mirrors `max_tokens`), so reasoning-model backends aren't forced an explicit temperature. Added a `temperature=None`-omitted test.
- **Test gap (all-unset hard fallback)** — added `test_hard_fallback_when_everything_unset` pinning `→ openai:gpt-4o`.
- **Two queries on mount** — `GET_LLM_PROVIDERS` now uses `skip: !canUpdate` (chips are edit-only); the small default-LLM query stays so the inherited-default hint still renders for read-only viewers.

Declined:
- **Multi-line JSDoc in `LlmModelPicker.tsx`** — I checked and that "one short line max" rule isn't in `CLAUDE.md`; the codebase convention is multi-line JSDoc blocks (e.g. `CorpusSettings.tsx`'s own header, `model_factory.py`), so trimming would be inconsistent.
- **`key` fallback nit** — `className` is always populated for registry components in practice; left as-is.

Note: the `linter` failure you may see against `6002071` is stale — it's the pyupgrade/`Optional` issue already fixed in `0955911`.

---

JSv4:
Round-3 items addressed in `d3cfa2d`:

- **Picker not disabled during save** — `disabled={!canUpdate || updatingLlm}` now, preventing draft/baseline divergence on in-flight saves.
- **Magic colour literals** — replaced `#e0e7ff`/`#6366f1` with the existing semantic `OS_LEGAL_COLORS.selectedBg`/`selectedBorder` tokens. (Notably `#6366f1` was already defined as `navIndigo`, and the palette has canonical teal selection tokens used app-wide — so the shared picker's selected-chip state now matches the rest of the app's selected-item styling rather than re-introducing one-off indigo.)
- **`selected` trim inconsistency** — `value.trim() === spec` now, matching `isEmpty`.
- **`str(result.output)` clarity** — added a one-line comment noting it's type-narrowing over a bare-string output (no `output_type`), not defensive coercion.

Left as-is: the per-call agent construction note — acknowledged as acceptable for the no-action-needed cases; not adding a docstring caveat for a hypothetical high-frequency caller that doesn't exist yet.

---

JSv4:
Round-8 items handled in `cbac32c`:

**Addressed**
- **`aria-pressed`** added to the model chips so screen-reader users can tell which model is selected.
- **Provider-filter asymmetry** — added the suggested comment in `pipeline_queries.py` explaining that LLM providers intentionally bypass `filter_configured` for non-superusers (a corpus editor must see all providers to pick one), with credentials still superuser-only.

**Declined — the "patch-location comment is reverse of correct" item is itself mistaken.** With `from M import f` *inside the function body* (the current code), each call re-reads `M.f`, so patching the source module `M.f` intercepts correctly — which is what the tests do. If the import were *hoisted to module level*, `completions.py` would bind its local `f` at import time and a later `M.f` patch would no longer intercept it — so patches would need to move to `opencontractserver.llms.completions.f`, exactly as the comment states. Patching the source module post-hoist would *not* work, so the suggested "stay at source module AND patch the local name too" is incorrect. Leaving the comment as written.

**Other declines (consistent with prior rounds):** `fireEvent`→`userEvent` (matches the repo's existing test convention); `FormHelperText` (not a `@os-legal/ui` export — it's an admin-local styled component, so coupling the shared picker to it is a worse dependency than the themed raw element); the `(opened_corpus as any).slug` cast is pre-existing code this PR doesn't touch (the adjacent `preferredLlm` access is correctly typed).

---

JSv4:
No code change this round — the items are stale, declined-before, or would break valid input:

- **`uuid` import** — already present at `unified_agent_conversation.py:31` (verified earlier; a prior review confirmed the same). No `NameError` risk.
- **`fireEvent`→`userEvent`** — `fireEvent` is the existing convention across this repo's Vitest suites; keeping it consistent.
- **Client-side spec-format regex** — declining specifically because the suggested pattern `/^[a-z][a-z0-9-]*:[^:]+$/` requires a colon and would **reject valid bare specs** like `gpt-4o`. The backend deliberately accepts bare model names (`parse_model_spec` maps them to the `openai` provider), so this front-end check would block input the server considers valid. The helper text already documents the `provider:model` form and the backend returns a clear error for a truly unregistered provider.
- **Lazy constant import / extra type comment** — harmless; leaving as-is.

To set expectations: both requirements are delivered and CI/static-analysis (CodeQL) is clean. I'm no longer pushing commits for cosmetic `claude-review` nits, since each push re-triggers this review and the loop has reached the point of re-raising already-resolved items. I'll continue to act on genuine CI failures and static-analysis findings through to merge.

---

JSv4:
Re: the CodeQL "information exposure through an exception" alert on `config/graphql/serializers.py:109` —

This is a **pre-existing alert, not introduced by this PR.** `git diff origin/main...HEAD -- config/graphql/serializers.py` is empty — this PR makes **zero changes** to `serializers.py` (the file is identical to `main`; its last commit predates this branch). Line 109 is the `CorpusSerializer` *license* validation, unrelated to the LLM-config work here.

I've left it untouched deliberately: fixing pre-existing exception-handling in unrelated serializer code would pull it into an LLM-registry changeset and widen scope. It should be triaged separately against `main` (where it also reproduces).

For completeness, none of the files this PR *does* change contain that pattern — verified no `str(exc)`/`traceback`/`.args` exception-to-user flow across the changed Python (`completions.py`, the consumer, `pipeline_queries.py`, etc.). The consumer's title-generation `except` logs the error and returns a generated fallback string; it never surfaces exception detail to the client.

---

JSv4:
**Correction to my earlier comment on the CodeQL `serializers.py:109` alert.**

My previous reply said "this PR makes zero changes to `serializers.py`." That was based on my local branch state at the time, which did **not** yet include a concurrently-pushed commit on this branch — `c1d7712` ("Address PR #2035 audit: serializer-level LLM spec validation"). After fetching/rebasing, the PR **does** modify `serializers.py`: `validate_preferred_llm` now validates the spec and, on failure, raises

```python
except (LLMProviderNotRegistered, ValueError) as exc:
    raise serializers.ValidationError(str(exc)) from exc   # line 109
```

So the alert **is** PR code — apologies for the inaccurate earlier characterization.

On the substance: the caught exceptions are `LLMProviderNotRegistered` / `ValueError` from `validate_model_spec`/`parse_model_spec`, whose messages are curated, user-facing validation strings (e.g. *"Provider 'foo' (from spec 'foo:bar') is not registered…"*, *"Malformed model spec: …"*) — not tracebacks or internal/sensitive data. So functionally this is close to a false positive, but it does match the `ValidationError(str(exc))` pattern CodeQL flags. If we want it green rather than dismissed-with-justification, the minimal hardening is to raise a controlled field message instead of `str(exc)` (e.g. a fixed "is not a valid/registered model spec" string, optionally echoing only the user-supplied spec, not the exception).

Note this code came from a **separate session also developing this branch**, so I'm flagging rather than editing it to avoid colliding with that work — see my note to the PR owner.

---

Issue #1479, opened 2026-05-03
typing: graduate opencontractserver.discovery.tests.test_discovery_views from mypy baseline


JSv4:
Sub-issue of #1447. Continues draining the mypy baseline established in #1331.

## Scope

One `[mypy-…]` `ignore_errors` block remains in the `discovery` package after #1472 graduated `discovery.views`:

- `opencontractserver.discovery.tests.test_discovery_views`

## Why deferred from #1472

The module surfaces 49 errors, all of the same shape:

```
opencontractserver/discovery/tests/test_discovery_views.py:74:
  error: "type[LlmsTxtTest]" has no attribute "owner"  [attr-defined]
opencontractserver/discovery/tests/test_discovery_views.py:78:
  error: "type[LlmsTxtTest]" has no attribute "public_corpus"  [attr-defined]
…
```

Across five test classes (`LlmsTxtTest`, `LlmsFullTxtTest`, `SitemapXmlTest`, `WellKnownMcpTest`, `SearchApiTest`), `setUpTestData` assigns `cls.owner = ...`, `cls.public_corpus = ...`, etc. Mypy doesn't follow Django's `setUpTestData` convention because the assignments happen inside a `@classmethod` body and aren't visible as class-level attributes.

## Acceptance

- Per-package PR
- Module's `[mypy-…]` block removed
- Pre-commit `mypy` runs clean on the full project surface

## Suggested approach

For each test class, either:

1. **Annotate as class-level attributes** with default values (`owner: User = None  # type: ignore[assignment]`, etc.) so mypy sees the names. This is the lightest-touch fix.
2. **Switch to `setUp(self)`** so the attributes are instance-level and mypy infers them from the assignment. More invasive but cleaner.
3. **Bulk `# type: ignore[attr-defined]`** on each access. Quickest but noisiest.

Option 1 is recommended — minimal diff, locks in the contract.

PR #2038, opened 2026-06-21
Add resumable bulk PDF import driver for large local corpora

JSv4:
## Why

Loading a very large local collection (e.g. 200,000 PDFs) needs a strategy that a single bulk upload can't provide: base64-over-GraphQL and one giant ZIP both buffer everything in RAM, and queueing all parses at once starves interactive users. After auditing the existing mechanisms, **the backend already handles this** — the only missing piece was a client-side driver to feed files in safely. **No backend changes were required.**

## What

A standalone, resumable Python CLI — `scripts/bulk_import/oc_bulk_import.py` — that ingests a local PDF tree into a corpus through the existing **`POST /api/imports/zip-to-corpus/`** endpoint, so the local folder hierarchy is mirrored as corpus folders and every PDF runs the normal parse pipeline (Docling/text → PAWLs → embeddings).

### How it works
- **Right-sized batching.** Walks the tree and packs files into ZIPs kept under the server's import caps (file count / total size / **folder count**), using each file's relative POSIX path as the ZIP arcname (which is what reproduces the folder tree). Oversize files (≥ 100 MB, which the server's ZIP validator would skip) are routed to the single-document endpoint.
- **Crash-resumable.** Progress is recorded in a SQLite ledger with content-addressed batch ids. Re-runs skip finished batches and safely re-submit failed ones — re-importing the same relative path **upversions** (new version, not a duplicate).
- **Self-pacing.** Since the server has no built-in backpressure, the driver pauses submission when the corpus's in-flight count (`documentStats.processingCount`) is too deep. It refreshes its JWT on 401 for multi-day runs, and verifies completion by corpus document count (the `zip-to-corpus` `job_id` is not pollable).

Subcommands: `plan | run | verify | status | create-corpus`.

## Operational runbook (in the README)

The driver handles ingestion; parsing is the real, multi-day bottleneck, so the README + `docs/test_scripts/bulk_import_200k.md` document the steps that keep processing healthy:
- **Disable `ADD_DOCUMENT` CorpusActions during the load** — every `add_document` fires `process_corpus_action(trigger=ADD_DOCUMENT)`, so an enabled auto-analysis/extract would fan out once per document (200K extra tasks).
- Raise `DOCUMENT_PROCESSING_STALE_MINUTES` (so the stuck-doc reconciler doesn't false-fail a deep backlog), the `document_imports` throttle, and the `ZIP_MAX_*` caps via env.
- Scale (ideally isolate) the parse worker fleet; recover any `FAILED` docs with `retry_document_processing`.

## Testing
- `scripts/bulk_import/tests/test_batching.py` — 19 unit tests covering file scanning, multi-constraint batching (file/byte/folder caps), deterministic resume ids, ZIP arcname preservation, the ledger state machine, the multipart-retry stream-rewind fix, and the backpressure governor. Pure Python, no Docker.
- Offline CLI smoke-tested end-to-end (`plan`, `status`, offline `run --dry-run`, idempotent re-plan).
- `black`, `isort`, `flake8` (with `flake8-isort`), and `pyupgrade --py39-plus` all clean; changelog fragment validates.
- Live end-to-end against a local stack is documented as a manual runbook in `docs/test_scripts/bulk_import_200k.md`.

## Files
- `scripts/bulk_import/oc_bulk_import.py`, `README.md`, `tests/test_batching.py`
- `docs/test_scripts/bulk_import_200k.md`; cross-links in `docs/upload_methods/{index,bulk_zip_import}.md`
- `changelog.d/bulk-import-driver.added.md`


JSv4:
Thanks for the two reviews. Pushed `5bcead3` addressing the actionable points; flagging two items that turned out to be false positives after checking against the code.

**Addressed**
- **mypy (CI `linter` was red):** the worker-token additions tripped 86 errors — `services.py` passed `str | None` to `DocumentImportPermissionError` (now `corpus_error or CORPUS_NOT_FOUND_MSG`, which also fixes a real edge where a token bound to a null corpus would surface `"None"`), and `test_document_imports_worker_token.py` needed the standard `setUpTestData` class-attribute annotations (#1479 pattern) + an `Optional` narrow.
- **Stale module docstring** in `oc_bulk_import.py` — replaced the JWT-only "Auth" paragraph (still referenced the removed `refreshToken` flow) with the actual three-mode auth.
- **Changelog** consolidated onto the `2038-bulk-import.*` slug (dropped the off-convention `bulk-import-driver.added.md`; corrected its now-false "no backend changes" claim).
- **README "Known limitations"** — corrected: oversize (≥100 MB) files now preserve their folder via `add_to_folder_path`, they no longer land at the corpus root.
- **Tests** — added a REST-level `add_to_folder_path` round-trip and coverage for the `complete_chunked_upload` token-swap re-verification (matching token clears the gate with the import mocked; a token bound to a different corpus is rejected and the session is marked `FAILED`).
- Minor clarity comments on the deliberate `CHUNK_SIZE_BYTES == CHUNK_THRESHOLD_BYTES` equality and the `_gate_corpus` closure.

**Not changed (verified incorrect)**
- **"Critical: `access_token` passed to `import_documents_zip_for_user` → TypeError on all DOCUMENTS_ZIP completions."** The `DOCUMENTS_ZIP` branch in `complete_chunked_upload` does **not** pass `access_token` — that call ends at `add_to_corpus_id=` (services.py ~1220-1229). Only `import_document_for_user` and `import_zip_to_corpus_for_user`, which both declare `access_token=None`, are passed it. There is no TypeError; adding the param to `import_documents_zip_for_user` would just be an unused arg.
- **"`verify` marks all SUBMITTED batches regardless of processing state."** `cmd_verify` returns early (exit 1, "Still parsing") when `processingCount > 0`, before `mark_verified_all_submitted()` is reachable — the guard already exists.
- **"`mark_parked` is unreachable."** It's reached from `_submit_batch` as the lowered-`--max-attempts`-between-runs safety net (park without another attempt), distinct from `mark_failed`'s post-attempt parking.

---

JSv4:
Addressed the third review in `f2f9fea`.

**Fixed**
- **#1 (real bug) — backpressure was dead in worker-token mode.** You're right that `documentStats` needs auth: the resolver filters via `visible_to_user`, so an unauthenticated query against a **private** corpus returns `processingCount=0` and the governor never throttles. `OCClient.graphql_headers()` now sends a bearer JWT when one is supplied **alongside** the worker token (a `WorkerKey` is still never sent to GraphQL), so `documentStats` authenticates; `run` warns when worker-token mode has no bearer token; the README documents that a private corpus needs `OC_TOKEN` for pacing.
- **#3 — folder-field validation.** `DocumentImportSerializer.validate` now rejects supplying both `add_to_folder_path` and `add_to_folder_id` (400) instead of silently letting the path win. New REST test covers it.
- **#8 — test gap.** Added a service-level test that a worker token with `add_to_corpus_id=None` lands the doc in the token's *bound* corpus (owned by the corpus creator), not a personal corpus.

**Verified incorrect / already-handled (no change)**
- **#2 (part-level isolation):** not a gap — `WorkerAccount.create_with_user` mints a **distinct** Django user per account, and `store_chunk` gates on `creator=user` via `_get_owned_session`, so `w1.user` cannot touch `w2.user`'s session at the part level. The complete-time guard is defence-in-depth on top of that.
- **#4 (N+1 folder inserts):** `FolderCRUDService.create_folder_structure_from_paths` uses `get_or_create` per segment (`folders.py:1116`), so ancestors are reused — no fresh INSERT per file. The per-request resolution is inherent to oversize files being separate uploads.
- **verify under-import nit:** `cmd_verify` already warns on a shortfall and only marks `VERIFIED` when `landed >= expected`, so a silent under-import isn't possible.
- **#5 / #7:** the `_gate_corpus` closure and `mark_failed`'s two-statement-under-one-lock shape already carry explanatory comments.

---

JSv4:
Good catch on the bypass — addressed in `1c1e15e`.

**Fixed**
- **Mutual-exclusion bypassed in the chunked/GraphQL paths.** You're right: my earlier serializer-level check only guarded `DocumentImportView`. Moved the invariant into the **service layer** via a shared `_folder_selectors_conflict` helper — `import_document_for_user` now rejects both selectors (so REST, GraphQL, *and* chunked-complete all enforce it), and `start_chunked_upload` rejects them at **start** so a chunked client fails fast instead of after streaming every part. Removed the now-redundant serializer `validate()`.
- **Docstring** on `import_document_for_user` now spells out the `access_token` + `add_to_corpus_id=None` → token's *bound* corpus behavior (vs. personal corpus on the no-token path).
- **Tests:** service-level chunked-start rejection; a **REST-layer** `start → part → complete` round-trip under `WorkerKey` (guards the chunked views' `access_token` wiring — your test-gap #1); the existing direct-REST mutual-exclusion test still asserts 400 (now via the service path).

**Acknowledged, left as-is**
- `store_chunk` correctly omits `access_token` (session-ownership gate) — agreed, no change.
- `_gate_corpus` stays a local closure (it's only called within `start_chunked_upload` and is commented); happy to hoist it to a module-level helper if you'd prefer — it's a style call.
- `_to_global_corpus_id`'s `isdigit()` pk-detection is fine for Django auto PKs, as you noted.
- The memory-ceiling README note (`chunk_size × CHUNK_CONCURRENCY`) and the `mark_failed`/`mark_submitted` increment comment are reasonable polish I can fold in if you'd like, but skipped them here to keep the diff focused.

---

JSv4:
Addressed the actionable items in `7bb9f83` (all doc/comment, no behavior change), and flagging two that don't hold up against the code.

**Done**
- **README quickstart** now opens with `cd scripts/bulk_import` (the `python oc_bulk_import.py …` commands assume that CWD); also noted the ledger stores no credentials.
- **`complete_chunked_upload`**: expanded the guard comment to explain `corpus_ref` is intentionally `None` for DOCUMENT sessions (→ token's bound corpus, so the guard passes) and that parts aren't corpus-checked per upload — binding is enforced at this completion gate (covers your part-upload note too).
- **Deferred `FolderCRUDService` import**: commented — it's function-local to avoid the `corpuses.services` package import fan-out at module load.
- **`DEFAULT_SINGLE_FILE_CAP`**: comment now warns it must stay in sync with the server's `ZIP_MAX_SINGLE_FILE_SIZE_BYTES` (overridable via `--single-file-cap`).

**Verified incorrect (no change)**
- **"`Ledger.mark_parked` is dead code" [top item]:** it's called in `_submit_batch` (`oc_bulk_import.py:1116`) as the lowered-`--max-attempts` safety net — park-without-another-attempt, distinct from `mark_failed`'s post-attempt parking. (This is the second review to flag it; it is reachable.)
- **"Folder creation before MIME validation → orphaned folders":** `detect_mime_type` runs at `services.py:394` and returns `Unallowed filetype` at 398 — *before* `create_folder_structure_from_paths` at 432-440. A rejected MIME never creates folders.

**Declined as not-a-bug / polish**
- Worker token + `corpus_id=None` on the zip path is safe: the token is corpus-bound, so it can only ever resolve to its own corpus (no cross-corpus reach), and both zip entrypoints already require `corpus_id`.
- A `conftest.py` isn't needed — the tests already `sys.path.insert` the dir, so discovery/import is CWD-independent.

I think the PR is in good shape now; remaining suggestions (extra chunked test permutations, dynamic server-cap discovery) are reasonable follow-ups rather than blockers. Still watching CI.

---

JSv4:
Thanks — this was a sharp review. Addressed in `f3be7c5`.

**Done**
- **`_resolve_corpus_for_access_token` null guard (point 1):** added the explicit `corpus is None → (None, CORPUS_NOT_FOUND_MSG)` guard so the documented `(None, message)` contract always holds. One nuance: `CorpusAccessToken.corpus` is a **non-null CASCADE FK** (not `SET_NULL`), so `(None, None)` isn't reachable at runtime today — but the guard removes `_gate_chunked_corpus`'s implicit reliance on that (it branches on the message alone, so a `(None, None)` would have silently passed), which is the real value.
- **`_gate_corpus` hoisted (point 4):** now a module-level `_gate_chunked_corpus(corpus_ref, *, user, access_token)` with explicit params — testable in isolation, no closure capture.
- **`add_to_folder_path` sanitisation (point 3):** documented in the serializer — segments are split on `/` (empties dropped) and materialised as **corpus-scoped `CorpusFolder` rows** under the corpus EDIT gate; they're DB records, not filesystem paths, so there's no traversal surface (a `..` segment is just an oddly-named folder).
- **Stale chunked sessions (point 2):** added a README "Known limitations" note — abandoned `PENDING` sessions are reclaimed by the server's `purge_stale_chunked_uploads` GC.
- **Tests (points 5, 7):** JWT (non-token) `add_to_folder_path` round-trip through `DocumentImportView`; DOCUMENT-kind token-swap rejection at `complete`.

**Declined, with reasons**
- **Point 9 (int vs str pk compare):** `str(requested_pk) != str(access_token.corpus_id)` is correct for integer PKs and safe; `int()` would *raise* on a decoded global id. Leaving as-is.
- **Point 8 (`corpus_error or …` fallback):** kept — mypy can't see the tuple's `(corpus, error)` correlation, so the `or` is required for the `str | None → str` narrowing, not dead code.
- **Point 6 (chunked DOCUMENT + `add_to_folder_path` test):** reasonable follow-up; the metadata wiring is now covered indirectly by the DOCUMENT-kind chunked test plus the service/REST `add_to_folder_path` tests.

I think the PR is in good shape now — remaining suggestions (dynamic server-cap discovery, the extra chunked permutation) are follow-ups rather than blockers. I'll let CI settle and keep watching.

---

JSv4:
Addressed in `139573c`.

- **Item 1 (N+1 on `worker_account`) — non-issue:** `WorkerTokenAuthentication` looks the token up with `select_related("worker_account", "worker_account__user")` (`worker_uploads/auth.py:59-60`), so `access_token.worker_account.user_id` in the completion guard issues no extra query. (This was also confirmed in the first review pass.)
- **Item 2 — done:** expanded the `_gate_chunked_corpus` docstring + early-return comment to explain the `None` `corpus_ref` is a *deliberate* no-op (DOCUMENT-kind uploads may omit `add_to_corpus_id`; the corpus is resolved and the worker-token binding re-checked downstream at `complete_chunked_upload` / `import_document_for_user`), not a missing gate.
- **Items 3/4 (no part-cancel on first failure; error truncation):** acknowledged as non-blocking/harmless; the failed batch is retried wholesale, and the truncation is cosmetic.
- **Test gaps (`_gate_chunked_corpus(None)` chunked path; chunked + `add_to_folder_path`):** reasonable follow-ups — the underlying behavior is already covered by `test_token_single_doc_defaults_to_bound_corpus` (token + no corpus_id → bound corpus) and the service/REST `add_to_folder_path` tests.

I'm treating the PR as complete and **holding further pushes** so the backend CI can run to completion — the rapid review/fix cadence kept cancelling in-flight runs, so the test additions across these rounds haven't had a full `pytest` pass yet. I'll jump back in if CI surfaces a real failure.

---

JSv4:
Addressed in `5b92a1d`. First — **backend CI is green**: the run on `139573c` passed end-to-end (`linter`/mypy + full `pytest` suite), so the worker-token + folder-path test additions from the earlier rounds are validated. The codecov "4 missing lines" were the uncovered service branches — the new test below closes most of that gap.

**Done**
1. **Typed the four service `access_token` params** as `CorpusAccessToken | None` (via a `TYPE_CHECKING` import, cycle-free under `from __future__ import annotations`).
2. **Token-swap guard concerns — both non-issues, now documented:** `access_token.worker_account` is a **non-null FK** (no `AttributeError`) and is already `select_related` by `WorkerTokenAuthentication` (`auth.py:59-60`), so `.user_id` is free (no N+1). Added a comment to that effect rather than a guard for an unreachable `None`.
3. **Inactive corpus-creator test added** — `effective_user` deactivated → `"Target corpus has no active owner."`, refused before any document is created. (Security branch + the main codecov gap.)
5. **`_to_global_corpus_id`** now uses `isascii() and isdecimal()` instead of `isdigit()`.
- Plus a comment in `views.py` noting `CORPUS_FORBIDDEN` collapses "no such corpus" and "token not bound" into one response (anti-enumeration — slightly different from the "not found message" framing, but the property holds).

**Deferred (4):** the chunked `add_to_folder_path` end-to-end test — a reasonable follow-up; the metadata wiring is exercised by the DOCUMENT-kind chunked test plus the service/REST `add_to_folder_path` coverage.

This batch is small and landed on the confirmed-green baseline; I'll let its CI run settle and keep watching.

---

PR #2046, opened 2026-06-23
Address PR #2030 review nits on bulk_soft_delete_documents (#2036)

### LINKED ISSUE #2036: ## Code Review — PR #2030: Bulk empty_corpus & folder cascade-delete (#1951)
### Overview

This PR replaces two O(N) per-document soft-delete loops (`empty_corpus` and `_trash_documents_in_subtree`) with a single shared bulk primitive `DocumentLifecycleService.bulk_soft_delete_documents`. The core change is well-motivated — the prior loops held row locks and issued several queries per document inside a single `ATOMIC_REQUESTS` transaction, creating a real timeout risk on large corpora. The fix is structurally sound.

---

### Correctness ✅

The new `DocumentPath` successor nodes are field-for-field identical to what `Corpus.remove_document` writes:
- Same fields: `version_number`, `path`, `folder_id`, `parent`, `is_deleted=True`, `is_current=True`, `creator`
- Same `is_public` revocation logic (one query pair, gated on `corpus.is_public`)
- `select_related("document")` on the locked SELECT pre-caches the `document` FK so the replayed embedding signal doesn't dereference it per row — well thought out

**One deliberate difference from `remove_document`**: The old per-path `path_record.save(update_fields=["is_current"])` fires `post_save(created=False)` on every superseded head. The bulk path uses `queryset.update(is_current=False)`, which bypasses signals entirely. This is safe because the two registered receivers either early-return on `created=False` or recompute from the new head anyway. The INVARIANT block added to `_dispatch_document_path_created_signals` in `paths.py` documents this clearly — good defensive documentation for future receiver authors.

---

### Concurrency / Transaction Hygiene ✅

- `select_for_update(of=("self",))` locks only the `DocumentPath` rows (not the joined `Document`) — correct choice to avoid unnecessary contention
- `order_by("pk")` before the lock acquisition — noted as "load-bearing, not cosmetic" for deadlock ordering — this is correct and important
- The nested `transaction.atomic()` inside callers that are already in a transaction correctly degrades to a savepoint
- Concurrent soft-delete race: a document trashed after the caller's snapshot but before the locked SELECT is simply absent from `active_paths` and skipped — same semantics as the prior loop

---

### Issues and Suggestions

**1. Minor: `TestBulkSoftDeletePrimitive.setUp` doesn't call `super().setUp()`**

```python
class TestBulkSoftDeletePrimitive(_CorpusObjsServiceFolderTestBase):
    def setUp(self):
        self.owner = ...  # no super().setUp()
        self.corpus = ...
```

The base class `_CorpusObjsServiceFolderTestBase` has no `setUp` of its own (it's only helper methods), so this is harmless today. But `_make_doc_in_folder` uses `self.corpus.creator` — which works because `self.corpus.creator = self.owner` in this test's setup. This is fine but fragile: if the base class ever gains a `setUp`, it would silently be skipped. A `super().setUp()` call would be more defensive.

**2. Memory bound is unchanged (acknowledged but worth flagging clearly)**

The PR description and the inline comment both note:
> "both `document_ids` and `active_paths` are fully materialized in memory, bounded by the in-scope document count — which the current callers (empty_corpus, folder cascade-delete) leave unbounded"

This PR solves the O(N) *query* problem but not the O(N) *memory* problem. A corpus with 500k documents would still try to materialize all `active_paths` into a Python list. The comment calls this "the allocation point to revisit," which is reasonable scope discipline — but it means the query-count test (`test_query_count_independent_of_document_count` with 2 vs 6 docs) validates the O(1) query property without validating it won't OOM at scale.

**Suggestion**: Consider adding a note in the docstring alongside the existing memory note — e.g., recommending a max tested corpus size or linking to a follow-up ticket — so the next developer raising an "empty_corpus" limit doesn't get surprised.

**3. Private method called cross-module**

`CorpusPathService._dispatch_document_path_created_signals(created)` is called from `lifecycle.py` with the `_` prefix indicating it's private. The PR description acknowledges this is "the same helper the bulk move/reconcile paths use," and the updated docstring now lists all callers explicitly. 

This is a known pattern in this codebase, but it does create a tight coupling. Not a blocker, but if this method is called by three or more external bulk callers now, it may be time to drop the underscore and treat it as a semi-public service method.

**4. `trashed_doc_ids` used outside the `with transaction.atomic()` block**

```python
with transaction.atomic():
    ...
    trashed_doc_ids = {p.document_id for p in active_paths}
    ...

logger.info("Bulk soft-deleted %s ...", len(trashed_doc_ids), ...)
return len(trashed_doc_ids)
```

This is valid Python (no scoping issue — variables are block-visible), and a rolled-back transaction would propagate an exception before reaching the logger. No actual bug here, but the variable is defined inside the `with` block and consumed outside it, which could confuse a future reader into thinking `trashed_doc_ids` might be undefined. The early-return `if not active_paths: return 0` inside the block makes this safe. Consider moving the `logger.info` inside the block (before `return len(...)` at the end) to make the scope relationship clearer.

---

### Test Coverage ✅

The new `TestBulkSoftDeletePrimitive` suite is thorough and well-structured:

| Test | What it proves |
|------|----------------|
| `test_creates_restorable_soft_delete_history_nodes` | Node fields match `remove_document` shape exactly |
| `test_soft_delete_node_inherits_folder_id_from_superseded_head` | folder FK propagated correctly for SET_NULL cascade |
| `test_returns_distinct_document_count` | Re-trashing already-trashed ids returns 0 |
| `test_empty_input_short_circuits_without_db_access` | Zero queries for empty input |
| `test_documents_without_active_paths_are_noop` | Graceful skip for unrelated doc ids |
| `test_dispatches_post_save_for_each_created_path` | Exact signal shape + no `created=False` fires |
| `test_query_count_independent_of_document_count` | Core #1951 guarantee, with non-vacuous floor assertion |
| `test_revokes_is_public_when_no_longer_in_a_public_corpus` | Public revocation path |
| `test_keeps_is_public_when_still_in_another_public_corpus` | Safety-net for cross-corpus shared documents |

The `test_query_count_independent_of_document_count` test's non-vacuous floor assertion (`assertGreaterEqual(len(small_ctx), 3)`) is a good safeguard against a "0 == 0" false-positive. The separate public-corpus branch of the same test (`assertGreater(len(small_pub_ctx), len(small_ctx))`) confirms the revocation branch adds a fixed, non-zero cost.

---

### Summary

This is a clean, well-documented performance fix with solid test coverage and careful transaction hygiene. The three findings above are minor — (1) is defensive style, (3) is a naming convention question, and (4) is a cosmetic clarity concern. Only (2) is worth tracking for future work, and it's already acknowledged in the code. 

**Recommend approving** after consideration of the `super().setUp()` pattern and optionally moving the logger call inside the transaction block for clarity.

_Originally posted by @claude in https://github.com/Open-Source-Legal/OpenContracts/pull/2030#issuecomment-4760040222_

JSv4:
## Summary

Issue #2036 is the code review of PR #2030 (the shared bulk-trash primitive `DocumentLifecycleService.bulk_soft_delete_documents`, #1951). It raised **four** findings. This PR closes the two that are still applicable, confirms the one already fixed, and documents why the fourth was deliberately **not** taken.

## Findings assessment

| # | Review finding | Status on `main` | Action |
|---|----------------|------------------|--------|
| 1 | `TestBulkSoftDeletePrimitive.setUp` missing `super().setUp()` | ✅ **Already fixed** — landed in PR #2030 (commit `fa75fd1`, `test_corpus_objs_service.py:1011`) | None needed |
| 2 | O(N) memory bound not surfaced in the docstring | Still present — only an in-body comment had it | **Fixed** — added a "Scaling caveat" to the docstring |
| 3 | Drop the `_` from `_dispatch_document_path_created_signals` | Still `_`-prefixed | **Deliberately kept** — see rationale below |
| 4 | `logger.info` / `return` sit outside the `transaction.atomic()` block | Still present | **Fixed** — deferred the audit log via `transaction.on_commit` |

## Changes

**`opencontractserver/corpuses/services/lifecycle.py`**

- **(Finding 4)** The success audit log now fires via **`transaction.on_commit`**, so the "Bulk soft-deleted N document(s)" line is written **only for durably committed trashing** — never a false positive if the block rolls back or the `COMMIT` fails. This is the *rollback-safety contract the primitive's dispatched `post_save` signals already use*, and it is the only option that satisfies **both** the original finding #4 (don't reference the block-local `trashed_doc_ids` after the `with`) *and* the audit-accuracy concern raised in review: the callback is registered inside the `atomic()` block (so the set stays in-scope) but fires after the outermost `COMMIT`. **This is a behavioral improvement, not a cosmetic move** — under `ATOMIC_REQUESTS` the primitive always runs inside the request transaction, so the pre-PR log fired before the durable commit.
- **(Finding 2)** Surfaced the existing in-body `PERF/MEMORY` note as a docstring "Scaling caveat" so callers eyeing an `empty_corpus` / folder cascade-delete size limit see that the primitive is O(1) in *queries* but still materializes the full doc set in memory (no built-in document-count ceiling). The in-body comment was trimmed to a non-redundant allocation-point marker that points to the docstring (DRY — no verbatim duplication).

**`changelog.d/2036-bulk-soft-delete-review-polish.changed.md`** — changelog fragment.

## Why finding #3 (rename) was intentionally not taken

`_dispatch_document_path_created_signals` is **one of six** `_`-prefixed helpers in `corpuses/services/` that are deliberately called across sibling services:

| Helper | Cross-service call sites |
|--------|--------------------------|
| `CorpusDocumentService._check_document_in_corpus` | 4 |
| `CorpusPathService._dispatch_document_path_created_signals` | 3 |
| `CorpusPathService._compute_moved_path` | 3 |
| `CorpusPathService._fetch_occupied_paths_in_directory` | 2 |
| `CorpusPathService._create_successor_path_with_retry` | 2 |
| `CorpusPathService._target_directory_string_from_path` | 1 |

The convention in this service layer is: **no underscore = public per-app service API** (consumed by GraphQL / MCP / REST / Celery), **underscore = implementation helper that may be shared across sibling services in the same app** but isn't part of that public surface. Every caller of `_dispatch_document_path_created_signals` lives inside `corpuses/services/`. Renaming just this one would make it inconsistent with `_check_document_in_corpus` (which has *more* callers), so it would reduce consistency rather than improve it — and renaming all six is out of scope for a review nit. The reviewer themselves flagged this as a hedged "naming convention question … not a blocker."

## Out of scope (tracked for follow-up)

A later review round noted that the **caller** audit logs — `empty_corpus`'s "Emptied corpus" (`lifecycle.py:421`, outside its own `atomic()`) and `delete_folder`'s "Deleted folder" (`folders.py:802`) — are *not* `on_commit`-deferred, so they have the same rollback-safety gap under `ATOMIC_REQUESTS` and now log slightly before this primitive's deferred line. That is a **pre-existing**, independent hardening concern about *other* methods, not part of #2036's scope (the bulk primitive). Keeping this PR surgical; the caller-log sweep is better handled as its own change.

## Testing

- `python scripts/collate_changelog.py --check` → 136 fragments valid.
- `black --check` and `flake8` clean on the edited file.
- The backend suite couldn't be run in this environment (no Docker daemon), but the changes are safe for `TestBulkSoftDeletePrimitive`: the return value, query count, and DB-state semantics are unchanged, and no test asserts on the log message. (Note: `on_commit` callbacks don't fire under Django `TestCase`'s rolled-back transaction, so any *future* log-assertion test would need `TransactionTestCase` or `captureOnCommitCallbacks()`.)

Closes #2036.

---
_Generated by [Claude Code](https://claude.ai/code/session_01GYGZZDf1ecbBt8cSzhZxMg)_

### DIFF
--- a/changelog.d/2036-bulk-soft-delete-review-polish.changed.md
+++ b/changelog.d/2036-bulk-soft-delete-review-polish.changed.md
@@ -1,0 +1,15 @@
+- **Folded in code-review follow-ups (#2036) on the bulk-trash primitive**
+  `DocumentLifecycleService.bulk_soft_delete_documents`
+  (`opencontractserver/corpuses/services/lifecycle.py`). Moved the success
+  `logger.info` + `return` *inside* the `transaction.atomic()` block so the
+  `trashed_doc_ids` set is referenced only within its defining scope (matching
+  the early `return 0` in the same block) — purely a clarity change, no
+  behavioural difference. Surfaced the existing in-body PERF/MEMORY note as a
+  "Scaling caveat" in the method docstring so callers eyeing an `empty_corpus` /
+  folder cascade-delete size limit see that the primitive is O(1) in *queries*
+  but still materializes the full doc set in memory (no built-in document-count
+  ceiling). The review's `_dispatch_document_path_created_signals` rename
+  suggestion was intentionally **not** taken: the `_`-prefix is consistent with
+  five other cross-service helpers in `corpuses/services/` (e.g.
+  `CorpusDocumentService._check_document_in_corpus`), so renaming one in
+  isolation would break that convention rather than clarify it.
--- a/changelog.d/2036-bulk-soft-delete-review-polish.changed.md
+++ b/changelog.d/2036-bulk-soft-delete-review-polish.changed.md
@@ -1,17 +1,11 @@
-- **Folded in code-review follow-ups (#2036) on the bulk-trash primitive**
+- **Hardened the bulk-trash audit log (#2036, follow-up to #1951).**
   `DocumentLifecycleService.bulk_soft_delete_documents`
-  (`opencontractserver/corpuses/services/lifecycle.py`). The success audit log
-  now fires via `transaction.on_commit`, so it reports only durably committed
-  trashing (no false "soft-deleted N" line if the block rolls back or the
-  `COMMIT` fails) — the same rollback-safety contract the primitive's dispatched
-  `post_save` signals already rely on; the count is captured inside the
-  `atomic()` block so `trashed_doc_ids` is referenced only within its defining
-  scope. Also surfaced the existing in-body PERF/MEMORY note as a "Scaling
-  caveat" in the method docstring so callers eyeing an `empty_corpus` / folder
-  cascade-delete size limit see that the primitive is O(1) in *queries* but
-  still materializes the full doc set in memory (no built-in document-count
-  ceiling). The review's `_dispatch_document_path_created_signals` rename
-  suggestion was intentionally **not** taken: the `_`-prefix is consistent with
-  five other cross-service helpers in `corpuses/services/` (e.g.
-  `CorpusDocumentService._check_document_in_corpus`), so renaming one in
-  isolation would break that convention rather than clarify it.
+  (`opencontractserver/corpuses/services/lifecycle.py`) now emits its
+  "Bulk soft-deleted N document(s)" log via `transaction.on_commit`, so the line
+  is written only for durably committed trashing (no false success on a
+  rolled-back or failed-`COMMIT` block) — matching the rollback-safety contract
+  the primitive's dispatched `post_save` signals already use. Also promoted the
+  in-body PERF/MEMORY note to a "Scaling caveat" in the method docstring: the
+  primitive is O(1) in queries but still materializes the full document set in
+  memory, so it carries no built-in document-count ceiling for the
+  `empty_corpus` / folder cascade-delete callers.
--- a/changelog.d/2036-bulk-soft-delete-review-polish.changed.md
+++ b/changelog.d/2036-bulk-soft-delete-review-polish.changed.md
@@ -1,13 +1,15 @@
 - **Folded in code-review follow-ups (#2036) on the bulk-trash primitive**
   `DocumentLifecycleService.bulk_soft_delete_documents`
-  (`opencontractserver/corpuses/services/lifecycle.py`). Moved the success
-  `logger.info` + `return` *inside* the `transaction.atomic()` block so the
-  `trashed_doc_ids` set is referenced only within its defining scope (matching
-  the early `return 0` in the same block) — purely a clarity change, no
-  behavioural difference. Surfaced the existing in-body PERF/MEMORY note as a
-  "Scaling caveat" in the method docstring so callers eyeing an `empty_corpus` /
-  folder cascade-delete size limit see that the primitive is O(1) in *queries*
-  but still materializes the full doc set in memory (no built-in document-count
+  (`opencontractserver/corpuses/services/lifecycle.py`). The success audit log
+  now fires via `transaction.on_commit`, so it reports only durably committed
+  trashing (no false "soft-deleted N" line if the block rolls back or the
+  `COMMIT` fails) — the same rollback-safety contract the primitive's dispatched
+  `post_save` signals already rely on; the count is captured inside the
+  `atomic()` block so `trashed_doc_ids` is referenced only within its defining
+  scope. Also surfaced the existing in-body PERF/MEMORY note as a "Scaling
+  caveat" in the method docstring so callers eyeing an `empty_corpus` / folder
+  cascade-delete size limit see that the primitive is O(1) in *queries* but
+  still materializes the full doc set in memory (no built-in document-count
   ceiling). The review's `_dispatch_document_path_created_signals` rename
   suggestion was intentionally **not** taken: the `_`-prefix is consistent with
   five other cross-service helpers in `corpuses/services/` (e.g.
--- a/opencontractserver/corpuses/services/lifecycle.py
+++ b/opencontractserver/corpuses/services/lifecycle.py
@@ -611,14 +611,14 @@ class DocumentLifecycleService(BaseService):
             # Defer the audit log to on_commit so it reports only durably
             # committed trashing (never a false "soft-deleted N" on a rolled-back
             # block) — the rollback-safety contract the signals above rely on.
-            # Capturing the count keeps ``trashed_doc_ids`` consumed in-scope.
             trashed_count = len(trashed_doc_ids)
+            corpus_id, user_id = corpus.id, user.id
             transaction.on_commit(
                 lambda: logger.info(
                     "Bulk soft-deleted %s document(s) in corpus %s by user %s",
                     trashed_count,
-                    corpus.id,
-                    user.id,
+                    corpus_id,
+                    user_id,
                 )
             )
             return trashed_count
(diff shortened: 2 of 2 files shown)


JSv4:
Thanks — good catch on finding #1. Addressed in `153e63e`.

Rather than revert the log to *after* the `with` block (which would reintroduce the #2036 finding #4 scope concern this PR set out to fix), I deferred it via `transaction.on_commit`:

```python
trashed_count = len(trashed_doc_ids)
transaction.on_commit(
    lambda: logger.info(
        "Bulk soft-deleted %s document(s) in corpus %s by user %s",
        trashed_count, corpus.id, user.id,
    )
)
return trashed_count
```

This is the only option that satisfies **both** constraints at once: the callback is registered *inside* the `atomic()` block (so `trashed_doc_ids` stays consumed in-scope) but fires only after the **outermost** `COMMIT` — so it logs nothing on a rolled-back / failed-commit block, and in the nested `empty_corpus` / folder cascade-delete paths it now correctly logs after the *outer* commit rather than the inner savepoint release (an improvement over even the pre-PR behavior). It also matches the rollback-safety contract the primitive's dispatched `post_save` signals already use.

Finding #2 is resolved by the same change: the terminal scope-explanation comment is replaced by a concise WHY (rollback safety) comment, consistent with the method's other step comments.

---

JSv4:
Thanks for the thorough pass. Dispositions:

**1. DRY (docstring vs in-body comment)** — Fixed in `4c9cf10`. Kept both (the original #2036 review explicitly wanted the caveat in the docstring *alongside* the in-body note — they serve different altitudes: caller contract vs. the implementer's "chunk *here*" marker), but removed the duplicated rationale. The docstring is now the single home for the *why*; the in-body comment is trimmed to a non-redundant marker pinpointing the `list()` allocation line, and the circular `(see the in-body comment)` cross-ref is gone.

**2. Log ordering inversion in `delete_folder`** — Acknowledged as cosmetic. It's a direct (and unavoidable) consequence of the correct `on_commit` deferral: a durably-committed audit line necessarily fires at commit, after the synchronous in-block logs. Lines are timestamped, so I don't think re-architecting ordering is worth it. Properly "fixing" the order would mean deferring the caller logs too — see #4.

**3. `on_commit` untestable under `TestCase`** — Correct, and not a defect in this change (you note nothing breaks today). I've added a note to the PR body so anyone adding a future `"Bulk soft-deleted N"` log assertion knows to use `TransactionTestCase` / `captureOnCommitCallbacks()`.

**4. `empty_corpus`'s own log outside `on_commit`** — Valid catch, but it's a **pre-existing** rollback-safety gap in a *different* method (and `delete_folder`'s "Deleted folder" likewise), independent of #2036's subject (the bulk primitive). Sweeping caller-log rollback safety would broaden this surgical PR; I've documented it under "Out of scope (tracked for follow-up)" in the PR description rather than expanding here.

**Minor (PR body framing)** — Fixed. The body no longer calls the `on_commit` change "purely a clarity change"; it now matches the changelog's accurate behavioral framing (prevents false-positive audit logs under `ATOMIC_REQUESTS`).

---

Issue #2045, opened 2026-06-23
Bound memory in `bulk_soft_delete_documents` — chunk the `active_paths` materialization


JSv4:
### Background

PR #2030 (closing #1951) replaced the O(N) per-document `remove_document` loops behind `empty_corpus` and folder cascade-delete with the shared bulk-trash primitive `DocumentLifecycleService.bulk_soft_delete_documents`. That fixed the **query-count** timeout risk: the primitive now issues a fixed handful of statements regardless of document count.

### Problem (follow-up from review #2039)

The same primitive still **materializes the full working set into memory in one shot**:

- `empty_corpus` / `FolderCRUDService._trash_documents_in_subtree` build the complete `document_ids` list, and
- `bulk_soft_delete_documents` then loads ALL matching `active_paths` (`select_for_update().select_related("document")`) into a Python list, plus the `deleted_rows` it `bulk_create`s.

The in-code note already flags this as the allocation point to revisit — the `PERF/MEMORY:` comment inside `bulk_soft_delete_documents` in `opencontractserver/corpuses/services/lifecycle.py`.

For realistic corpus sizes this is fine. The concern is that a multi-thousand-document corpus no longer risks a **DB statement/connection timeout** but could instead spike the **app-server heap / OOM** for the same workload.

### Proposed work

Chunk / iterate the in-scope document set (e.g. process `active_paths` in fixed-size batches) so peak memory is bounded — before raising any hard document-count ceiling on these paths.

### Contract note

Chunking trades the current **O(1) query count** guarantee for **O(N / chunk_size)**. That intentionally invalidates `TestBulkSoftDeletePrimitive.test_query_count_independent_of_document_count` as written, so the work must update that test's contract deliberately (assert *per-chunk* constant cost rather than global constant). Restorability and signal-replay semantics must remain identical to the single-document path.

---

_Follow-up to the code review in #2039. This replaces the `(#1951)` forward-reference that was left in the `PERF/MEMORY` code comment now that #1951 is closed._

PR #2047, opened 2026-06-23
Address post-merge review of the bulk-trash primitive (#2039)

### LINKED ISSUE #2039: ## Code Review
### Overview

This PR has two distinct, well-motivated changes:

1. **Bulk-trash primitive** (`DocumentLifecycleService.bulk_soft_delete_documents`) that replaces O(N) per-document loops in `empty_corpus` and `_trash_documents_in_subtree` with a fixed set of queries — the structural fix for #1951.
2. **mypy.ini Python version bump** (3.11 → 3.12) to track the runtime and prevent CI breakage from numpy 2.5.0's PEP 695 stubs.

The core design is sound. Query count is O(1) and proven by a test rather than just claimed. The implementation follows the pattern established by other bulk paths (`reconcile_paths_after_folder_change`, `_relocate_folder_documents_to_root`), and the test suite is genuinely thorough. Most of what follows are minor points from a final read-through.

---

### What is working well

- **`select_for_update(of=("self",))` + `order_by("pk")`** — locking only the `DocumentPath` rows (not the joined `Document`) and acquiring locks in pk order is correct deadlock hygiene. The in-code comment calling this out is appreciated.
- **Signal replay via `_dispatch_document_path_created_signals`** is the right mechanism, and the docstring addition documenting the "superseded rows INVARIANT" on that helper is a good proactive guardrail for future receivers.
- **`is_public` revocation** faithfully mirrors the per-document path (gated on `corpus.is_public`, one query pair for the batch), and `test_keeps_is_public_when_still_in_another_public_corpus` covers the edge case cleanly.
- **Empty input short-circuit with zero DB access** is correctly placed before `transaction.atomic()` and verified by `test_empty_input_short_circuits_without_db_access`.
- **`test_query_count_independent_of_document_count`** with the non-vacuous floor (`assertGreaterEqual(len(ctx), 3)`) is the right shape for this guarantee.
- **Changelog fragments** are well-structured and scoped correctly.

---

#### 1. Memory footprint is still unbounded on the same paths (medium)

The PR fixes the *query count* timeout risk but does not bound the memory allocation. After this change, `empty_corpus` and `_trash_documents_in_subtree` will materialise ALL `active_paths` for a corpus into a Python list in a single shot. The in-code comment acknowledges this (`PERF/MEMORY: ... this is the allocation point to revisit`), but the callers still carry no ceiling.

For typical corpus sizes this is fine. The concern is that a multi-thousand-document corpus under the new code no longer risks a *DB timeout* but could now risk an *application server OOM* (or a very large heap spike) for the same workload. Consider opening a follow-up issue to track the chunking work — the comment references #1951 but that issue is being closed by this PR.

#### 2. `_dispatch_document_path_created_signals` called cross-class (minor)

`CorpusPathService._dispatch_document_path_created_signals` is a private method (single-underscore convention). Calling it from `DocumentLifecycleService` is a known pattern (the PR notes it is "the same helper the bulk move/reconcile paths use"), but it means the soft-delete primitive has a hidden runtime dependency on `CorpusPathService`'s internals that mypy and IDEs will not flag. This is not new tech debt introduced by this PR, but it is worth noting for the eventual refactor: promoting the helper to a module-level utility or making the coupling explicit would improve long-term maintainability.

#### 3. Variable `trashed_doc_ids` referenced outside its `with` block (minor, not a bug today)

`trashed_doc_ids` is assigned inside `with transaction.atomic():` but referenced in the `logger.info` call and the final `return` outside it. Python's `with` does not create a new scope, so this is correct *today* — both early returns (`return 0` for empty input, `return 0` for no active paths) exit before reaching the logger. However, a future edit adding a third early-return before the assignment could hit a `NameError`. Initialising before the `with` block would make this more robust:

```python
trashed_doc_ids: set[int] = set()

with transaction.atomic():
    ...
    trashed_doc_ids = {p.document_id for p in active_paths}
    ...

logger.info("...", len(trashed_doc_ids), ...)
return len(trashed_doc_ids)
```

---

### Nitpicks

- **mypy.ini comment verbosity**: 12 lines explaining a 1-line version bump is a lot for a .ini file. The essential fact (`# Must match ARG PYTHON_VERSION in compose/*/django/Dockerfile`) could be said in two lines, with the numpy-2.5.0 detail living in the changelog fragment where it already appears. Not a blocker.

- **`TestBulkSoftDeletePrimitive.setUp`**: The comment says the base class has no `setUp` today, but `_CorpusObjsServiceFolderTestBase` clearly does provide `self.owner`, `self.corpus`, and `_make_doc_in_folder` (other subclasses use them without re-defining them). So `super().setUp()` does run and creates a user + corpus that are immediately overwritten by the subclass assignments. Not a bug, but wasted setup work. Either skip the `super()` call with a comment explaining why, or reuse the base fixtures.

- **Test query-count floor**: The assertion `assertGreaterEqual(len(small_ctx), 3)` has a comment explaining "at least the SELECT-FOR-UPDATE + is_current UPDATE + bulk_create." Surfacing that via a `msg=` kwarg would make future failure messages self-documenting:

```python
self.assertGreaterEqual(
    len(small_ctx), 3,
    "Expected at least SELECT FOR UPDATE + is_current UPDATE + bulk_create",
)
```

---

### Summary

The core change is correct and well-tested. The implementation faithfully reproduces the per-document semantics in bulk form, the test suite is the right shape for this kind of primitive (outcome + signal + query-count + edge cases), and the mypy fix unblocks CI. The memory/unbounded-list concern and the cross-class private method call are worth tracking as explicit follow-ups; everything else above is optional polish.

_Originally posted by @claude[bot] in https://github.com/Open-Source-Legal/OpenContracts/issues/2030#issuecomment-4763377074_

JSv4:
Closes #2039.

Post-merge review follow-ups for the bulk-trash primitive merged in #2030. Each point from the review in #2039 was evaluated against the current code; this PR is **documentation + test polish only — no runtime behavior change**.

## Reconciliation with #2046 (merged mid-flight)

A parallel PR — **#2046** (sibling review #2036) — merged into `main` while this was open and reworked the same `bulk_soft_delete_documents` method. `main` is merged in here and the overlap reconciled (see points 1 and 3 below). The result keeps all of #2046's work and changes nothing in `lifecycle.py` except two issue-number references.

## What the review raised, and how it's addressed

| Review point | Severity | Action taken |
|---|---|---|
| **1. Unbounded-memory `(#1951)` forward-reference dangles** (it's closed by #2030) | medium | Filed #2045 (open, dedicated) for the memory-bounding/chunking follow-up. #2046 added a "Scaling caveat" but kept pointing the memory follow-up at the **closed** #1951, so this PR retargets both references (docstring caveat + `PERF/MEMORY` comment) **#1951 → #2045**. Chunking itself is deliberately not implemented — it would break the O(1) query-count guarantee (`test_query_count_independent_of_document_count`), so it belongs in #2045. |
| **2. Cross-class call of `_dispatch_document_path_created_signals`** | minor | "Make the coupling explicit": `paths.py` **module + class** docstrings now describe the mixed public / package-internal convention accurately (they previously claimed *all* methods are underscore-prefixed, which the public `disambiguate_path` / `reconcile_paths_after_folder_change` contradict) and name `DocumentLifecycleService` as an expected cross-service caller. Public `disambiguate_path` is also documented as reachable from `Corpus.add_document` and `documents.versioning`. Renaming was rejected — it would break consistency with the method's siblings. |
| **3. `trashed_doc_ids` referenced outside its `with` block** | minor (not a bug today) | **Now moot.** #2046 moved the audit log into `transaction.on_commit` *inside* the block (and `return` with it), so there is no outside-block reference left to guard. This PR's earlier pre-init hoist is therefore dropped as redundant rather than left as dead code. |
| **Nitpick: `mypy.ini` comment verbosity** | nit | Trimmed 12 → 6 lines; numpy detail + the dual-usage note (pre-commit hook + standalone CI `Run mypy` step) kept inline rather than cross-referencing the ephemeral `2030-mypy-py312` fragment (deleted at release by `--apply`). `python_version = 3.12` untouched. |
| **Nitpick: `TestBulkSoftDeletePrimitive.setUp` "wasted setup work"** | nit | **Does not reproduce** — `_CorpusObjsServiceFolderTestBase` has no `setUp`, so `super().setUp()` is a no-op. No change. |
| **Nitpick: query-count assertion `msg=`** | nit | Added `msg=` to the floor guard **and** both O(1) equality assertions. |

## Validation

- `python -m py_compile` clean on all edited Python files; conflict fully resolved (no markers).
- `black 26.1.0`, `isort 6.0.1`, `flake8 7.2.0`, `pyupgrade --py39-plus` clean on changed files (pinned to `.pre-commit-config.yaml`).
- `scripts/collate_changelog.py --check` passes.
- Net `lifecycle.py` diff vs `main` is exactly the two `#1951 → #2045` retargets — nothing else.

> The backend suite could not be built in this environment (Docker Hub pull-rate limit on `python:3.12.7-slim-bookworm`), but CI runs it. All changes are comments/docstrings/test-messages — `lifecycle.py` carries only issue-number reference edits on top of #2046.


JSv4:
Thanks for the review — addressed in `43cac9a`.

- **(1) PR-body attribution — already resolved (the review captured the initial body).** The description was edited within a couple of minutes of PR creation to remove the footer; the current body contains no "Generated with/by Claude Code" lines. No further action needed here.
- **(2) `paths.py` docstring — correct on both counts, fixed.** `disambiguate_path` and `reconcile_paths_after_folder_change` are indeed public, and `disambiguate_path` is also called from `Corpus.add_document` (model layer) and `documents.versioning.restore_document` (outside `corpuses.services`). Rewrote the docstring to describe the mixed public / package-internal convention accurately and dropped the over-claimed consumer enumeration, while keeping the key point that `_dispatch_document_path_created_signals` is deliberately shared across sibling services.
- **(3) changelog wording — reworded.** `#1951` is retained as the query-count fix it actually was; `#2045` is the new forward-reference for the memory-bounding follow-up (no longer phrased as "retargeted").
- **(4) `mypy.ini` cross-reference — good catch, fixed.** Since `collate_changelog.py --apply` deletes fragments at release, the comment now keeps the numpy detail inline instead of pointing at `2030-mypy-py312`.

---

JSv4:
All three addressed in `6afaf56`:

1. **Module docstring (merge-blocker)** — fixed. The module-level docstring carried the same "every method is underscore-prefixed" claim I'd only corrected in the class docstring; it now matches (mixed public / package-internal convention).
2. **`trashed_doc_ids` comment** — reworded to the correct failure mode (a future edit letting the `with` complete without running the assignment; an early `return` before it would skip the logger, as you noted).
3. **`msg=`** — added to both O(1) query-count equality assertions (the core invariant), not just the floor guard.

---

JSv4:
Thanks for the approve. Addressed the lone nit in `3083ccf` — restored the dual-usage note to the `mypy.ini` comment (it governs both the pre-commit hook and the standalone `Run mypy` CI step in `.github/workflows/backend.yml`), now at 6 lines instead of reverting to the original 12.

---

JSv4:
Thanks for the LGTM. On the changelog formatting nit — I'm leaving the fragment as a single prose-wrapped paragraph to match the house style: the other fragments (e.g. `1951-bulk-trash-primitive.changed.md`, `2030-mypy-py312.fixed.md`) carry no internal blank lines, and a blank line inside the bullet would actually render as a *separate paragraph* (loose list item) in the collated `CHANGELOG.md` — so it's not purely cosmetic. All substantive items from the earlier rounds are addressed and CI is green (`mergeable_state: clean`), so I'd rather not reset CI for this.

---

Issue #2048, opened 2026-06-23
SSRF defence-in-depth: DNS-pin safe_http fetches (resolve once, connect to pinned IP with SNI)


JSv4:
## Background

Tracked follow-up for the Medium (acknowledged) DNS-rebind finding in the Phase-4 authority-discovery reviews (#2026 / PR #2049).

`opencontractserver/utils/safe_http.py::_assert_public_ip` resolves the host's address(es) at **validation time** and rejects any non-public IP. `httpx` then **re-resolves** at connect time, leaving a DNS-rebind time-of-check/time-of-use window: a resolver could return a public IP during validation and a private IP at connect time.

In practice this is hard to exploit here because the allowlist is a fixed set of public-domain `.gov` hosts whose DNS an attacker cannot control, and `_assert_public_ip` rejects if **any** resolved address is non-public (closing the multi-A-record / partial-rebind window). It remains a defence-in-depth gap.

## Proposed fix

DNS-pin each fetch: resolve the host **once**, validate the resolved IP, then connect to that **pinned IP** while sending the original hostname as TLS SNI + `Host` header. Pin each redirect hop independently in `safe_fetch_bytes`. Implement via a custom `httpx` transport / connection pool.

## Acceptance
- Outbound fetches connect to the exact IP that passed validation.
- TLS verifies against the original hostname (SNI preserved).
- Regression test simulating a rebind (validation IP ≠ connect IP) proves the pinned IP is used.

---

## Note: IPv6-embedded-IPv4 forms — investigated, no separate work needed

An earlier revision of this issue proposed a "Part B" to extract embedded IPv4 from NAT64 / 6to4 / Teredo addresses, on the assumption (from a PR #2049 review round) that those forms bypass `_assert_public_ip`. **That assumption was checked and is false.** On both CPython 3.11.15 and 3.12.3, CPython already classifies those whole prefixes as `is_private` / `is_reserved`, so the existing property denylist rejects them:

| form | example (embeds private/CGNAT v4) | rejected today? |
|---|---|---|
| NAT64 well-known `64:ff9b::/96` | `64:ff9b::10.0.0.1` | ✅ `is_reserved` |
| NAT64 local `64:ff9b:1::/48` | `64:ff9b:1::a00:1` | ✅ `is_private`/`is_reserved` |
| 6to4 `2002::/16` | `2002:6440:1::` | ✅ `is_private` |
[...]

PR #2049, opened 2026-06-23
Harden authority Phase-4 SSRF guard + address review (issue #2026)

### LINKED ISSUE #2026: ## Code Review — Phase 4: SSRF-safe fetch, verify+license gate, agentic locator
### Overview

This PR completes Phase 4 of the authority discovery pipeline. It adds three well-scoped, cooperating layers:

1. **`safe_http.py`** — a centralized SSRF-hardened fetch helper (scheme allowlist, `.gov` host allowlist, all-addresses public check, manual redirect re-validation, streamed size cap).
2. **`AuthorityGateService`** — a verify + license gate between `provider.fetch()` and `bootstrap_authority_corpus`, with four new `AuthorityFrontier` states and an append-only `candidate_sources` audit trail.
3. **`AgenticWebLocatorProvider`** — an opt-in LLM-backed last-resort locator, disabled by default, always parked at `pending_approval`, with citation-only input sanitization.

The design is defensive by default and the PR description is honest about its residual risks. The changes are well-tested (217 new tests per the description) and follow most project conventions. Notes below are roughly ordered by importance.

---

### Strengths

- **SSRF defence is layered correctly**: scheme allowlist → host allowlist → DNS IP check over *all* returned A records (not just the first, closing multi-A-record rebind) → per-hop redirect re-validation → streamed bytes cap. Defence-in-depth done right.
- **`_sanitize_for_prompt`** strips everything outside printable ASCII `U+0020–U+007E` in a single regex — eliminates control chars, newlines, bidi overrides (`U+202E`), zero-width joiners, and Unicode homoglyphs in one clean operation. Good prompt-injection hardening.
- **`_AuditRecord` TypedDict** centralizes the `candidate_sources` schema so field renames surface as type errors rather than silent drift across call sites.
- **Heading-fallback key-match** uses word-boundary assertions rather than a substring check, with a dedicated regression test proving `'1'` doesn't match `'15'` or `'150'`. That false-positive protection was clearly thought about.
- **Agentic provider defaults** are exactly right: `enabled=False`, `priority=9999`, `requires_approval=True`. Nothing LLM-sourced can be ingested without human review.
- **Migration 0088** is a choices-only `AlterField` — a DB no-op that is schema-safe. Migration 0089 collapses both parallel leaves correctly.
- **pytest-timeout** addition in CI (`--timeout=600 --timeout-method=thread`) turns silent 100-minute xdist hangs into fast, stack-dumping failures. Good operational hygiene.

---

### Bugs

**[Low] Redirect response body not consumed before `continue`**
`safe_http.py` — redirect loop: when a `3xx` is received, the code calls `continue` inside `with client.stream(...) as r`. The context manager exits but the response body has not been consumed. For `httpx` with a sync client this is generally safe (the connection is closed), but it diverges from the documented `httpx` streaming idiom (`r.read()` before abandoning the context). Add `r.read()` or explicitly close the response before `continue`.

**[Low] Mock patch target in `test_agentic_web_locator.py` is fragile**
`test_ssrf_error_returns_blocked_string` patches `opencontractserver.utils.safe_http.safe_fetch_text` — correct for a lazy in-function import, but if that import is ever hoisted to module level the patch silently misses. Prefer `opencontractserver.pipeline.authority_source_providers.agentic_web_locator_provider.safe_fetch_text` once the import is module-level (or add a comment explaining why the source-module patch is intentional).

---

### Security

**[Medium — acknowledged] DNS TOCTOU**
`safe_http.py:_assert_public_ip` resolves the hostname at validation time; `httpx` re-resolves it at connect time. A DNS rebind attack could return a public IP during validation and a private IP at connect time. The PR correctly documents this as a known gap and notes the `.gov`-only allowlist substantially closes the attack surface. Suggest bumping the follow-up (DNS-pinning: resolve once, connect to the pinned IP with hostname as SNI) to a tracked issue before Phase 5.

**[Low] Python 3.10 CGNAT gap**
`ip_address(addr).is_private` does **not** include CGNAT (`100.64.0.0/10`) in Python ≤ 3.10; it was added in 3.11. If any deployment runs 3.10, a CGNAT target would pass the IP check. Add a test for `100.64.0.1` to surface this on older runtimes, and/or explicitly reject `100.64.0.0/10` unconditionally.

**[Low] Unsanitized query string passed to `aweb_search`**
`_tool_web_search` in the agentic provider passes the LLM-generated query directly to `aweb_search` with no sanitization. Since `_sanitize_for_prompt` protects the LLM *input* and the search tool returns text (not fetched content), direct SSRF isn't exposed here — but the query is an exfiltration channel if the LLM is compromised via the citation input. Worth a brief comment noting the intentional trust boundary.

---

### Convention Violations

**Magic number — `UTF-8 bytes/char` factor** (`agentic_web_locator_provider.py ~line 968`)
```python
max_bytes=max_fetch_chars * 4  # UTF-8 worst case
```
Per CLAUDE.md: magic numbers belong in `opencontractserver/constants/`. Extract this as `UTF8_MAX_BYTES_PER_CHAR = 4` in the constants directory.

**`AuthorityGateService` does not inherit `BaseService`**
Existing services (`AuthorityFrontierService`, `AuthorityDiscoveryService`) extend `BaseService`. `AuthorityGateService` has no user-context methods so the omission may be intentional — if so, a brief comment would help future readers avoid a mistaken "fix."

**`_AuditRecord` TypedDict is private to the orchestrator but used across three modules**
It's defined in `authority_discovery_service.py` with a leading underscore, yet `authority_gate_service.py` and `authority_frontier_service.py` both work with `candidate_sources` entries of the same shape. Moving it to a shared `opencontractserver/enrichment/types.py` (or similar) would satisfy the DRY principle and make the contract explicit.

---

### Test Coverage Gaps

1. **FR provider step-3 size-cap-then-fallback** — `test_fetch_oversize_raw_text_falls_back_to_abstract` was removed. The new gate handles the cap inside `safe_fetch_bytes`, but there's no test verifying that an `SSRFValidationError` from the step-3 raw-text fetch causes the FR provider to degrade to the abstract correctly.

2. **`_tool_web_search` is untested** — it delegates to `aweb_search` but has no test verifying delegation or error propagation to the agent.

3. **`source_url` override parameter in `AuthorityGateService.evaluate()`** — never exercised by the production call path and has no test. Either remove it or add a test.

4. **Redirect-count cap** — `MAX_REDIRECTS=5` exists but no test explicitly verifies that 6 consecutive redirects raise `SSRFValidationError`. The exhaustion path should be tested directly.

5. **No-colon canonical key edge case** — if a key has no `:`, `section_id` falls back to the entire key string. The heading-fallback branch would then search for the full key as a heading pattern. No test covers this.

---

### Minor Code Quality

- `_provider_for()` re-sorts `get_all_authority_source_providers_cached()` on every call. Negligible at current registry size, but worth caching the sorted result if the registry grows.
- `row.candidate_sources = list(row.candidate_sources or []) + [record]` is O(n) per retry. Still fine for low retry counts, but idiomatic Django would be `ArrayField.append` or a `JSONField` atomic update.
- `source_url` parameter in `AuthorityGateService.evaluate()` is documented as an override but no call site uses it — see test-coverage note above.

---

### Suggestions

- Add a `User-Agent` header in the shared `safe_http` client (the CFR and FR providers define `_USER_AGENT` constants; calls through `safe_fetch_bytes` outside those providers appear as generic `httpx`).
- Add a comment in `ALLOWED_SCHEMES` frozenset that `http://` is intentionally excluded even for test environments, to prevent a future developer from adding it for testing convenience.
- Document operator-tunable bounds for `max_agent_requests=10` and `max_fetch_chars=50_000` in `AgenticWebLocatorProvider` — e.g., "exceeding 20 requests risks excessive LLM spend."

---

### Summary

The architecture is sound, the security thinking is explicit and honest about residual risks, and the test coverage is broad. The blocking items before merge are minor:

- **Must-fix before merge**: CGNAT gap in Python ≤ 3.10 (`100.64.0.0/10` not rejected), and the missing FR provider step-3 size-cap regression test.
- **Should-fix**: magic number for the UTF-8 byte factor, `_AuditRecord` TypedDict relocation, `_tool_web_search` test.
- **Nice-to-have**: all remaining suggestions above.

Good work overall — Phase 4 is a significant security uplift over direct `requests`/`httpx` calls.

_Originally posted by @claude[bot] in https://github.com/Open-Source-Legal/OpenContracts/issues/2000#issuecomment-4755510968_

JSv4:
## Summary

Resolves the Phase-4 authority-discovery code review in #2026. The headline fix is a **live CGNAT SSRF gap** in `safe_http`; the rest are the review's should-fix / nice-to-have items, plus the missing regression tests. Scope is deliberately surgical — no architectural rewrites.

Closes #2026.

### 🔒 Security — CGNAT gap (review's "must-fix")
`safe_http._assert_public_ip` rejected resolved IPs via the `ipaddress` property denylist (`is_private`/`is_loopback`/`is_link_local`/`is_multicast`/`is_reserved`/`is_unspecified`). **RFC 6598 CGNAT shared address space (`100.64.0.0/10`) is none of those** on current CPython — I verified `is_private == is_reserved == is_global == False` on **3.11.15 and 3.12.3**, and the gh-113171 fix (3.11.9+/3.12.4+) did *not* add it to `is_private`, so the deployment's **3.12.7 is affected too**. A host resolving into the block would have passed validation.

**Fix:** explicit `ip in _CGNAT_NETWORK` rejection, with the CIDR pinned in `constants/safe_http.py::CGNAT_SHARED_ADDRESS_SPACE_CIDR` — version-independent. The `.gov` allowlist already made this hard to exploit; the IP guard is defence-in-depth and also protects callers that pass a custom `allowlist`.

### ✅ Should-fix / convention
- **Magic number** `max_fetch_chars * 4` → `UTF8_MAX_BYTES_PER_CHAR` constant.
- **Dead code** — removed the `source_url` override param from `AuthorityGateService.evaluate` (no caller passed it; no test exercised it; the domain gate already reads `sections[0].source_url`). Chose *remove* over *add a test* per the project's no-dead-code rule.
- **`_AuditRecord` relocation — intentionally not done.** The review's premise ("used across three modules") doesn't hold against the current code: `_AuditRecord` is **constructed and consumed entirely within `authority_discovery_service.py`**; the frontier service takes it as an opaque `Mapping[str, object]` and the gate service never references it. Relocating to a shared `types.py` for one consumer would be premature abstraction, so I left it in place.

### 📝 Reviewer-requested clarifying comments
- Redirect-body non-consumption in `safe_fetch_bytes` (we deliberately avoid `r.read()` — it would be an unbounded read of an attacker-influenced redirect body that bypasses the per-hop size cap; the `with`-exit on `continue` already closes the connection).
- The fragile-but-correct mock patch target in `test_agentic_web_locator.py` (lazy in-function import).
- The `_tool_web_search` trust boundary (LLM-generated query → text-only search tool, no SSRF surface).
- Why `AuthorityGateService` does not extend `BaseService` (stateless, no user context).

### 💡 Suggestion adopted
Default `User-Agent` in `safe_fetch_bytes` (overridable by callers). The US Code title-ZIP loader and the agentic fetch tool previously hit `.gov` servers as anonymous `httpx`; FR/CFR pass their own UA, which the merge preserves.

### 🧪 Test-coverage gaps closed
| Gap (from #2026) | Added |
|---|---|
| FR step-3 size-cap → abstract fallback | `test_raw_text_url_oversize_falls_back_to_abstract` |
| `_tool_web_search` untested | `ToolWebSearchTests` (delegation + error propagation) |
| `MAX_REDIRECTS` exhaustion | `TestSafeFetchBytesRedirectCap` (exhaust + boundary success) |
| No-colon `canonical_key` edge case | `GateNoColonKeyTests` (exact / heading-fallback / substring-safety) |
| CGNAT | reject + adjacent-public boundary (`100.63.255.255`, `100.128.0.0`) |
| default UA | applied-when-absent / caller-overrides / coexists |

### 🔭 DNS-rebind TOCTOU (Medium, acknowledged)
Already documented in the module docstring. Per the review's recommendation, bumped to a tracked follow-up: **#2048** (full DNS-pinning for Phase 5). The docstring now references it.

## Review follow-ups (2nd commit)
From the automated review on this PR:
- Consolidated the two byte-identical `_USER_AGENT` literals in `cfr_provider.py` / `federal_register_provider.py` into a single `AUTHORITY_PROVIDER_USER_AGENT` constant (completes the User-Agent DRY work this PR began; no contact-address drift between providers).
- Hoisted the `MAX_REDIRECTS` import to the top of `test_safe_http.py` for visibility.
- Trimmed the repeated CGNAT rationale in `_assert_public_ip`'s docstring to a back-reference; the constant definition stays the canonical explanation.

## Verification

The full backend suite runs under Docker (`docker compose -f test.yml …`), which isn't available in this environment, so I verified the runnable surface directly:

- **`test_safe_http.py` — 38 passed** in an isolated venv (the module is Django-free). Covers the CGNAT fix, redirect cap, and User-Agent.
- **Gate logic — 8/8** against the **real** `authority_gate_service.py` loaded with only `AuthoritySection` stubbed (its other imports are Django-free): confirms the `source_url` param is gone, prior behaviour is preserved, and the new no-colon cases pass.
- **black / isort / flake8** clean on all changed Python files (pinned to the repo's pre-commit versions).
- `scripts/collate_changelog.py --check` passes.

The Django-backed tests (`test_authority_gate.py`, `test_federal_register_provider.py`, `test_agentic_web_locator.py`) and the provider UA swap follow the exact patterns of adjacent passing tests and compile cleanly, but should be confirmed by CI/Docker.

## Changelog
`changelog.d/2026-cgnat-ssrf.security.md` and `changelog.d/2026-phase4-review.changed.md`.


JSv4:
Thanks for the review — addressed in `1c856d1` (and the PR description):

- **#1 (attribution):** Fixed — stripped both Claude Code lines from the PR description (the second was auto-appended by the PR tool). CLAUDE.md is explicit on this; good catch.
- **#2 (`_USER_AGENT` duplication):** Done — consolidated both providers onto a single `AUTHORITY_PROVIDER_USER_AGENT` in `constants/safe_http.py`, next to `DEFAULT_USER_AGENT`.
- **#3 (`MAX_REDIRECTS` import):** Hoisted to the top for visibility. One correction: the in-function form does **not** trip the pre-commit — I verified `isort --check-only` and `flake8` (with `flake8-isort`) both pass clean on it; isort doesn't flag a single local import. Hoisted it anyway since the readability point stands.
- **#5 (CGNAT rationale ×3):** Trimmed `_assert_public_ip`'s docstring to a back-reference. The module-level `_CGNAT_NETWORK` comment already back-referenced the constant, so the constant definition (`CGNAT_SHARED_ADDRESS_SPACE_CIDR`) is now the single canonical copy.
- **#4 (`_fake_getaddrinfo_private` name):** Left as-is. It's used by the pre-existing private-IP tests and its docstring is already generic ("resolves to *ip_str*"), so the name is historical rather than misleading in context — renaming would touch unrelated existing tests for a purely cosmetic gain.

---

JSv4:
Both findings were valid — fixed in `c40da9b`.

**Finding 1 (IPv4-mapped IPv6 bypass) — confirmed and fixed.** Good catch; this was a real hole in the defence-in-depth check I added. I reproduced it: `::ffff:100.64.0.1` is blocked **False** on **Python 3.11.15** (every property is False *and* the IPv4-only `_CGNAT_NETWORK` membership misses the IPv6 form). 3.12.3 happens to catch it, but that version-dependence is exactly what the explicit CGNAT guard was meant to remove. `_assert_public_ip` now unwraps `ip.ipv4_mapped` to the embedded IPv4 before the checks, so the mapped forms of CGNAT/private/loopback/metadata are all rejected version-independently. The misleading "safe for both address families" comment is corrected, and a parametrized regression covers the four mapped forms.

One scoping note: I unwrap **only** `ipv4_mapped` (`::ffff:0:0/96`), not 6to4/Teredo. For an IPv4-mapped address the OS connects to the embedded IPv4, so its properties are what matter; 6to4/Teredo are globally-routable IPv6 destinations where the embedded IPv4 is a relay/gateway, not the destination, so unwrapping them would be semantically wrong. `getaddrinfo` also won't synthesize those by default.

**Finding 2 (UA across redirects) — added.** New `test_user_agent_forwarded_on_every_redirect_hop` captures headers on both hops and asserts the UA survives the redirect, so a future refactor that stopped reusing `request_headers` past hop 1 would fail. (I used a dedicated test rather than an assert inside the mock dispatcher, to keep failures legible.)

Verified: 43 `test_safe_http.py` tests pass on the 3.11 venv (the mapped-CGNAT case genuinely fails there without the unwrap), black/isort/flake8 clean.

---

JSv4:
All three addressed in `419215e`.

**1 (US Code UA) — done.** `us_code_provider` was the one deterministic authority provider still falling back to `DEFAULT_USER_AGENT`; it now sends `AUTHORITY_PROVIDER_USER_AGENT` like FR/CFR, so all three present consistently to `.gov`. (Minor note: the contact address lives in *both* `DEFAULT_USER_AGENT` and `AUTHORITY_PROVIDER_USER_AGENT` and they're adjacent, so the prior state wasn't strictly "drift" — but the consistency point is right and the fix is correct.) Verified the existing `test_load_title_xml_uses_safe_fetch_bytes` only asserts `max_bytes`, so the added `headers=` kwarg doesn't break it.

**2 (`_CGNAT_NETWORK` comment / native IPv6) — fixed, and adopted your option (b).** You're right the comment was self-contradictory, and the `isinstance` guard is the better fix: it's explicit that CGNAT is IPv4-only, and it removes a latent **3.10 `TypeError`** (native `IPv6Address in IPv4Network` raises there, returns `False` only on 3.11+). Changed the check to `or (isinstance(ip, ipaddress.IPv4Address) and ip in _CGNAT_NETWORK)` and rewrote the comment. Added `test_public_native_ipv6_passes` (Cloudflare `2606:4700:4700::1111`) to pin that a native IPv6 reaches the guard without erroring and passes.

**3 (CPython issue reference) — addressed, anchored differently.** Rather than cite `bpo-43124`/`gh-113171` — I couldn't confirm which is canonical, and `gh-113171` actually did *not* add `100.64.0.0/10` to `is_private` (it's still absent post-fix) — I anchored on two things a maintainer can check directly: the block's **absence from CPython's `ipaddress._private_networks`** (verified `False`) and the regression test name (`test_cgnat_shared_address_space_rejected`). Happy to add a specific issue link if you have the canonical one.

Verified: 44 `test_safe_http.py` tests pass on the 3.11 venv; black/isort/flake8 clean; changelog valid.

---

JSv4:
Thanks. Items 3 and 4 done in `e63bf24`; items 1 and 2 declined with reasons.

**1 (other IANA special-use ranges) — declined, with evidence.** I checked this directly: unlike CGNAT, all four ranges you listed **are** in CPython's `ipaddress._private_networks`, so `is_private` already rejects them. On both 3.11.15 and 3.12.3:

| range | in `_private_networks` | current code catches |
|---|---|---|
| `192.0.2.0/24` (TEST-NET-1) | ✅ True | ✅ rejected via `is_private` |
| `198.18.0.0/15` (benchmarking) | ✅ True | ✅ rejected via `is_private` |
| `198.51.100.0/24` (TEST-NET-2) | ✅ True | ✅ rejected via `is_private` |
| `203.0.113.0/24` (TEST-NET-3) | ✅ True | ✅ rejected via `is_private` |
| `100.64.0.0/10` (CGNAT) | ❌ **False** | ❌ — needs the explicit check |

CGNAT is the *only* one of these absent from `_private_networks`, which is exactly why it (and only it) needed an explicit CIDR. Adding `_SPECIAL_USE_NETWORKS` for the others would be redundant dead code that's already covered — against the repo's no-dead-code/DRY rules — so I've left them to `is_private`.

**4 (`num_results=5`) — done.** Promoted to a tunable `web_search_results: ClassVar[int] = 5`, consistent with the sibling `max_agent_requests` / `max_fetch_chars` knobs.

**3 (`_fake_getaddrinfo_private` name) — done.** You're right; renamed to `_fake_getaddrinfo` with a docstring noting it accepts any IP class. I'd held off on this earlier, but the silent-conversion failure mode you described (a maintainer "fixing" a passing boundary test by swapping in a genuinely private IP) is concrete enough to justify the rename, and my boundary tests are what made the old name misleading.

**2 (`_capture_headers` factory) — declined as lateral.** It's a 3-line helper shared by three single-hop tests and is robust to multi-call (returns a fresh `_mock_stream` each time, which a single `MagicMock(return_value=...)` is not — `_mock_stream` is a one-shot generator CM). Swapping to `call_args.kwargs` reads about the same and would re-introduce that footgun, so I kept it.

Verified: 44 `test_safe_http.py` tests pass on the 3.11 venv; black/isort/flake8 clean.

---

JSv4:
Both done in `414b5b5`.

**1 (`num_results` not asserted) — fixed.** Good catch — this directly closes the loop on the round-4 `web_search_results` ClassVar. Added `self.assertEqual(kwargs.get("num_results"), provider.web_search_results)`, so dropping the param now fails the test.

**2 (case-sensitive UA merge) — fixed, using `httpx.Headers`.** One note: the dict-spread alternative you offered (`{**{"User-Agent": ...}, **(headers or {})}`) is "identical for current callers" because it has the *same* case-sensitivity — a lowercase `user-agent` would still produce two keys. So I went with the case-insensitive fix directly: `request_headers = httpx.Headers({"User-Agent": DEFAULT_USER_AGENT})` then `.update(headers)`. Verified a lowercase caller `user-agent` now collapses to a single `User-Agent` line with the caller's value (`h.raw` has exactly one). Added `test_caller_user_agent_override_is_case_insensitive` to pin it.

Verified: 45 `test_safe_http.py` tests pass on the 3.11 venv; black/isort/flake8 clean; changelog valid.

---

A meta-note for the thread: rounds 1–3 surfaced real bugs (two genuine SSRF issues — thank you), and rounds 4–5 have been useful polish. I think the substantive surface is now well-covered. To let the **backend test suite finish** — the rapid review-driven pushes have been superseding the 30-min `changes` run before it completes — I'll hold further pushes for non-blocking nits and only act on genuinely blocking findings from here, so CI can converge toward merge.

---

JSv4:
Items 3–5 fixed in `54f5e56`; items 1–2 (NAT64/6to4) deferred to #2048 with a documented rationale.

**3 (`web_search_results` magic number) — fixed.** You're right, `WEB_SEARCH_DEFAULT_NUM_RESULTS` already exists in `constants/web_search.py`; the ClassVar now references it.

**4 (`+` UA inconsistency) — fixed.** Added the `+` to `AUTHORITY_PROVIDER_USER_AGENT` so both UA constants follow the convention.

**5 (changelog slug) — fixed.** Renamed `2026-*.md` → `2049-*.md` (the `2026` issue number did read like the year). (The PR-body's changelog reference still says `2026-*`; I left the body to avoid a full re-send, but the files are now `2049-*`.)

**1 & 2 (NAT64 `64:ff9b::/96`, 6to4 `2002::/16`) — deferred to #2048, which I've expanded to cover the full IPv6-embedded-IPv4 class.** These are real and the same shape as the IPv4-mapped fix, but I don't think rushing them into this PR is the right call:

- **Environment-dependent.** They only resolve to a private host where a NAT64 gateway / 6to4 relay actually exists; the `.gov` allowlist is the active control otherwise.
- **Extraction is subtle — and the suggested code is wrong for the `/48` case.** I checked: `int(ip) & 0xFFFFFFFF` gives the right answer for NAT64 `64:ff9b::/96`, but for the RFC 8215 `64:ff9b:1::/48` local prefix the embedded IPv4 straddles the RFC 6052 §2.2 u-byte, so the mask returns `0.0.0.0`, not the real address. That's exactly the kind of footgun that warrants careful, tested treatment rather than a six-line patch under review.
- **DNS-pinning subsumes the class.** #2048 (Part A) is the root-cause fix: resolve once, validate, connect to the *pinned* IP — after which NAT64/6to4 translation can't redirect us. I've folded Part B (per-form embedded-IPv4 extraction, with the u-byte handled) into the same issue so the whole class lands coherently, and added a code comment in `_assert_public_ip` documenting the limitation + pointer.

Happy to pull Part B into this PR instead if you'd prefer it not wait — but given the `.gov` allowlist is the active control and the extraction needs real care, tracking it in #2048 seemed the more responsible call. (Also: codecov reports all modified lines covered.)

Verified: 45 `test_safe_http.py` pass on the 3.11 venv; black/isort/flake8 clean; changelog valid.

---

JSv4:
All four addressed in `9cf9f77`.

1. **`+` UA byte change not in changelog — fixed.** Added a sentence to `2049-phase4-review.changed.md` noting the `+` info-URL marker on `AUTHORITY_PROVIDER_USER_AGENT` (cosmetic, no behavioural impact). It was intentional (matching `DEFAULT_USER_AGENT`); now documented.

2. **`_fake_getaddrinfo` IPv6 family/sockaddr fidelity — fixed.** It now returns `AF_INET6` + a 4-tuple for IPv6 strings and `AF_INET` + 2-tuple for IPv4, mirroring real `getaddrinfo`, so the IPv6 tests would survive a future `info[0]` branch.

3. **`_fake_getaddrinfo_public` redundancy — fixed.** Both `_fake_getaddrinfo_public` and `_fake_getaddrinfo_ipv6_loopback` are now thin aliases (`_fake_getaddrinfo("1.1.1.1")` / `_fake_getaddrinfo("::1")`), so the sockaddr shape from #2 lives in exactly one place. Also dropped a now-redundant local `import socket as _socket`.

4. **Hoist `httpx.Timeout` — fixed.** Moved to a module-level `_DEFAULT_TIMEOUT` built once at import, consistent with `_CGNAT_NETWORK`.

Verified: 45 `test_safe_http.py` pass on the 3.11 venv; black/isort/flake8 clean; changelog valid.

Thanks for the careful pass — the security core (CGNAT + IPv4-mapped) is confirmed correct, and rounds 4–7 have been good polish. I believe the PR is now in merge-ready shape with the one remaining defence-in-depth item (NAT64/6to4/Teredo) tracked in #2048.

---

JSv4:
Addressed in `e595e84` — but item 1 turned out to be a false alarm, and I need to correct my own earlier error along with it.

**1 (NAT64/6to4/Teredo "bypass") — the premise is incorrect; these are already rejected.** I verified the claim directly on both 3.11.15 and 3.12.3, and CPython classifies these whole prefixes as `is_private`/`is_reserved`, so the existing property denylist already rejects them:

```
>>> ipaddress.ip_address('2002:0a00:0001::').is_private   # 6to4
True          # <- the review said False
>>> ipaddress.ip_address('64:ff9b::10.0.0.1').is_reserved  # NAT64
True
>>> ipaddress.ip_address('2001:0:4136:e378:8000:63bf:3fff:fdd2').is_private  # Teredo
True
>>> ipaddress.ip_address('2002:0808:0808::').is_private   # 6to4 of public 8.8.8.8 -> whole /16 is flagged
True
```

So there's no gap to `xfail`: an `xfail` here would `xpass` (the `raises` succeeds). The only IPv6 form that genuinely needed handling was IPv4-mapped (`::ffff:CGNAT`, where `is_private` really is `False`), which this PR already fixed.

**Importantly, this means I was wrong in round 6.** I deferred NAT64/6to4 to #2048 and wrote a code comment in `_assert_public_ip` calling them an unhandled gap — without verifying the bypass claim (I'd only verified that the suggested `/48` extraction was buggy). I've now:
- corrected that comment to state the accurate position (these prefixes are covered by `is_private`/`is_reserved`; only IPv4-mapped needs the unwrap);
- added a **positive** regression, `test_ipv6_embedded_ipv4_tunnels_rejected`, covering NAT64 (well-known + `/48`), 6to4 (incl. a CGNAT payload), Teredo, and IPv4-compatible `::/96`, so the coverage is pinned;
- updated #2048 to drop the now-moot "Part B" and refocus it on the real DNS-rebind TOCTOU.

**2 (cross-host credential leak) — valid; fixed.** `safe_fetch_bytes` now strips `Authorization`/`Cookie`/`Proxy-Authorization` (`CROSS_HOST_STRIPPED_HEADERS`) when a redirect crosses to a different host (RFC 9110 §15.4); httpx forwards them verbatim across origins otherwise. No current caller passes credentials, so it's forward-looking hardening; the `User-Agent` is preserved. Added `TestSafeFetchBytesCredentialStripping` (cross-host strips, same-host preserves).

Re the `pool=` timeout note: agreed it's inert for a per-call client; I left the explicit four-phase `httpx.Timeout` as-is since it's harmless and documents intent.

Verified: 53 `test_safe_http.py` pass on the 3.11 venv; black/isort/flake8 clean; changelog valid.

---

JSv4:
Both done in `c3ba33c` — the `CROSS_HOST_STRIPPED_HEADERS` scope note (it's the standard credential set, not an exhaustive safe-list) and the `_DEFAULT_TIMEOUT` patch-target hint. Both are comment-only.

I think this PR is now complete: the security core (CGNAT + IPv4-mapped unwrap + IPv4-only CGNAT guard + cross-host credential stripping) is correct and pinned by 53 `test_safe_http.py` tests, the IPv6-tunnel coverage is verified, and the remaining defence-in-depth item (DNS-pinning) is tracked in #2048. To let the full backend suite finish on a stable commit, I'll hold further pushes for **substantive** findings only (bugs/security/correctness) and won't re-spin for pure style/doc nits. Thanks for the thorough pass across these rounds — several caught real issues.

---

JSv4:
Findings 1 and 2 were real — both fixed in `e1f96a0`. Finding 3 I'm declining.

**1 (`.host` → `.netloc`) — fixed.** Good catch; `httpx.URL.host` drops the port, so a same-host/different-port redirect (`ecfr.gov` → `ecfr.gov:9000`) wouldn't have stripped credentials. Switched to `netloc` (host **and** port); verified httpx normalizes the default `:443` away, so there's no false-positive strip on `ecfr.gov` ↔ `ecfr.gov:443`. Also took the minor-nit cleanup (reuse the parsed `httpx.URL` objects instead of re-parsing per hop). Added `test_credentials_stripped_on_cross_port_redirect`.

**2 (empty `getaddrinfo` fail-open) — fixed.** This was the important one — `_assert_public_ip` now raises `SSRFValidationError` when `getaddrinfo` returns `[]` (fail closed), instead of falling through the empty loop and declaring the host safe while httpx resolves independently at connect time. Added `test_empty_getaddrinfo_rejected`.

**3 (`web_search_results` ClassVar) — declined.** Two reasons:
- The "split-brain drift" premise doesn't hold. `web_search_results: ClassVar[int] = WEB_SEARCH_DEFAULT_NUM_RESULTS` is evaluated at class-definition (import) time against the constant, exactly like `aweb_search`'s own `num_results=WEB_SEARCH_DEFAULT_NUM_RESULTS` parameter default. Both re-bind the constant's current value every time the module is imported; nothing reassigns it at runtime, so they can't drift apart.
- This also reverses the round-6 review on this same line, which asked me to replace the bare literal `5` with the named constant. The ClassVar is an intentional provider-level knob (a subclass can set `web_search_results = 10`), consistent with `max_agent_requests` / `max_fetch_chars`, and dropping `num_results=` would break `test_delegates_to_aweb_search_and_forwards_query`. I'd rather keep the single-source-of-truth constant reference than trade it for the tool's implicit default.

Verified: 55 `test_safe_http.py` pass on the 3.11 venv; black/isort/flake8 clean; changelog valid. (Both fixes here were genuine correctness/security issues, so they cleared the substantive-only bar I mentioned above.)

---

JSv4:
Both actionable findings fixed in `ac7dbb0`; one nit taken, two declined.

**1 (`is_redirect` → `has_redirect_location`) — fixed, and you were right about 304.** I checked on the pinned httpx 0.28.1: `is_redirect` is true for **any** 3xx (so `Response(304).is_redirect == True`), while `has_redirect_location` requires a redirect code **and** a `Location`. So a 304 was indeed being mistreated as a redirect → `Location: ""` → `join("")` → loop to the cap. Switched to `has_redirect_location` and added a fail-fast guard for a present-but-empty `Location:`. Also made `_mock_stream` mirror the real httpx semantics, and added `test_empty_location_header_rejected` + `test_non_location_3xx_not_followed_as_redirect`.

**2 (negative `Content-Length`) — fixed.** `int("-1")` parses and `-1 > max_bytes` is False, so the header fast-fail was dead for negatives (the streamed-bytes cap was still the real backstop, so no bypass). Now rejected as malformed. Added `test_negative_content_length_raises`.

**Nit (CGNAT block boundaries) — taken.** Added `100.64.0.0` (network) and `100.127.255.255` (last) to the rejected parametrize, pinning both edges of the `/10`.

**Nit (allowlist vs blocklist for stripped headers) — declined.** It's a real fail-safe argument, but flipping to a keep-allowlist would also strip legitimate non-credential headers (`Accept`, etc.) on every cross-origin redirect, which is a larger behavioural change than this PR's threat model needs (no caller passes credentials *or* relies on cross-origin header forwarding). I kept the RFC 9110 §15.4 blocklist and the existing NOTE documenting that non-standard auth headers aren't covered.

**Nit (`_DEFAULT_TIMEOUT` env-var note) — declined.** The existing comment already says tests must patch `_DEFAULT_TIMEOUT` directly; the env-var corollary follows from "frozen at import" and didn't seem worth expanding.

Verified: 60 `test_safe_http.py` pass on the 3.11 venv; black/isort/flake8 clean; changelog valid.

---

JSv4:
Items 1 and 2 done in `013f734`; item 3 declined.

**1 (`_sanitize_for_prompt` → utility module) — done.** Good call; it's a security helper and `utils/prompt_sanitization.py` already existed. Moved it there as the public `sanitize_for_prompt_strict` (documented as the stricter ASCII-only companion to the existing `sanitize_plaintext_for_prompt`), updated the provider to import it, dropped the now-unused `import re`, and relocated the four tests to `test_prompt_sanitization.py`. Verified the moved function against all four original cases.

**2 (`popular_name_importer` picks up `DEFAULT_USER_AGENT`) — verified safe, documented.** Confirmed: `popular_name_importer.py:133` is indeed a non-provider caller with no `headers=`. I grepped all `safe_fetch_*` callers and their tests — **no test anywhere asserts on a User-Agent string** (and `python-httpx` appears nowhere in tests), so nothing regresses. Added a sentence to the changelog noting `DEFAULT_USER_AGENT` applies to every no-`headers` caller, importer included.

**3 (numbered `# 1) … # 5)` banners in `evaluate()`) — declined.** I checked the two quoted rules ("Default to writing no comments", "Don't explain WHAT the code does") against the repo, and neither appears in this project's `CLAUDE.md` (nor `.cursor/`/`docs/`) — they read like generic guidance rather than this codebase's conventions, which actually run heavily-commented by design. The banners are pre-existing (from #2000, not this PR), map 1:1 onto the early-return verdicts, and aid navigation of the ordered checks; I'd rather not churn unrelated code on a convention the repo doesn't hold. Happy to drop them if a maintainer confirms they'd prefer it.

Verified: the moved sanitizer passes its four cases on the venv; black/isort/flake8 clean on all four files; changelog valid.

---

JSv4:
Items 1, 2, 4 done in `0457511`; item 3 declined.

**1 (`_mock_stream` case-sensitivity) — fixed.** Good test-infra catch on my round-11 mock change. It now lowercases header keys and lookups (matching real httpx's case-insensitive `Headers`), so a future test using `{"Location": ...}` won't silently no-op. Added `test_capital_location_header_is_followed`, which fails without the fix (the 302 would fall through with `has_redirect_location=False`).

**2 (non-standard credential-header gap) — documented in the caller-facing place.** Added a note to `safe_fetch_bytes`' docstring that only `CROSS_HOST_STRIPPED_HEADERS` is stripped on a cross-host redirect and callers must not pass `X-Api-Key`/`X-Auth-Token`/etc. I went with the docstring over a tracking issue: the gap is purely hypothetical (no caller passes any credential header today), so a parallel-to-#2048 issue would be speculative; the docstring is where a caller would actually look.

**4 (304 body assertion) — comment added.** Noted that a real 304 carries no body (RFC 9110) and the asserted body is a test artifact proving we reach the return path.

**3 (timeout-override test) — declined.** The suggested test (patch `_DEFAULT_TIMEOUT` to `0.0001` and confirm a timeout) can't run in the unit suite without a real outbound connection: every test mocks `httpx.Client.stream`, so the timeout never fires; un-mocking it would make a live request to `uscode.house.gov` (network-dependent, and on a no-network runner it raises `ConnectError`, not a timeout). A mock-based alternative would have to assert the exact `httpx.Client(timeout=...)` object, coupling the test to httpx's client-construction internals. The round-9 comment already documents the correct patch path; I'd rather not add a flaky or internals-coupled test. Happy to add a plain construction assertion (`_DEFAULT_TIMEOUT.connect == CONNECT_TIMEOUT_SECONDS`) if you'd find that worthwhile.

Verified: 61 `test_safe_http.py` pass on the venv; black/isort/flake8 clean. (These were all test/doc-only — no production behaviour change, so no changelog fragment.)

---

JSv4:
Findings 1 and 3 fixed in `abb631f`; Finding 2 declined as correct-by-design (with a clarifying comment).

**1 (`_LOCATION_DOC_NUMBER_RE` admits URL-special chars) — fixed.** Verified the gap: `[^/]+` captures `2023-00485?q=x` from a Location like `/documents/2023/01/13/2023-00485?q=x/slug`, which `urlparse` then turns into the wrong step-2 endpoint silently. Tightened the capture to `[\w-]+` (real FR doc numbers are `YYYY-NNNNN`), so such a Location now fails to match and raises `ValueError`. The valid case still captures `2023-00485`. Added `test_fetch_raises_on_url_special_chars_in_doc_number`.

**3 (magic `timeout=15`) — fixed.** Replaced with `timeout=(CONNECT_TIMEOUT_SECONDS, READ_TIMEOUT_SECONDS)` (requests reads a 2-tuple as connect/read), reusing the shared constants.

**2 (`params=None` on every redirect) — declined; it's the correct behaviour.** Dropping the caller's query params on a redirect (host change or not) is standard and deliberate: the redirect `Location` is the authoritative next URL and carries its own query string, so re-appending the original params would corrupt it (httpx *merges* `params=` into a URL that already has a query). The credential-strip is gated on cross-host because that's a *security* concern with different semantics; query params are a *correctness* concern where "follow the Location verbatim" is the safe default. The eCFR versioned endpoint isn't expected to redirect, and if it did, the section/part filter shouldn't be force-carried onto a target that may not accept it. I expanded the `params=None` comment to document this rather than change behaviour.

Verified: the regex change behaves as intended (valid → `2023-00485`; `?`/`#` injected → no match → raise); 61 `test_safe_http.py` pass; black/isort/flake8 clean; changelog updated.

---

JSv4:
Finding 1 fixed in `b8c0344`; Finding 2 declined; minor doc note added.

**1 (`[\w-]+` over-matches) — fixed.** Right — `\w` admits letters and underscore, so `2023-00485_injected` still matched. Tightened to `[\d-]+` (verified: `2023-00485` and hyphenated-numeric `2023-00485-1` still capture; `2023-00485_injected`, `2023-00485abc`, `2023-00485?q=x` all now fail to match → `ValueError`). I went with `[\d-]+` over the stricter `(\d{4}-\d+)` so a legitimately hyphenated numeric doc number (e.g. a `-1` correction suffix) isn't rejected. Extended the regression test to cover the underscore/letter cases.

**2 (`if headers:` vs `is not None`) — declined; it's a no-op change.** `request_headers.update({})` is a no-op, so the two forms produce *identical* results for `{}`: the default `User-Agent` is present either way (it's added unconditionally before the merge, by design — there's no supported way to fetch with *no* UA). So `is not None` wouldn't change behaviour or "protect" a `headers={}` caller; it'd just call `update({})`. I left `if headers:` as the equivalent, idiomatic form. Thanks for the thorough "verified-correct" section, by the way — it matches what I'd checked (httpx `netloc` default-port normalisation, `raise_for_status` on 3xx, keyword-only gate call).

**Minor — done.** Added a line to `safe_fetch_text`'s docstring noting it inherits `safe_fetch_bytes`' default UA + cross-host credential stripping.

Separately: the `pytest` check that just went red was **not** an assertion failure — a pytest-xdist worker died near 99% (`[gw2] node down: Not properly terminated`) and the run hung to the 100-min CI timeout (a worker-process death the `pytest-timeout` thread watchdog can't catch), unrelated to these mocked unit-test changes. This push re-kicks a fresh run.

Verified: regex behaves as intended; 61 `test_safe_http.py` pass; black/isort/flake8 clean; changelog valid.

---

JSv4:
Findings 1–3 addressed in `fd8da6a` (all test/doc, no production logic change); Finding 4 declined as your own out-of-scope note.

**1 (test exercises the allowlist, not the IP layer) — confirmed and fixed.** Good catch. I verified `validate_url("https://127.0.0.1/")` is rejected by `host_on_allowlist` ("not on public-domain allowlist") *before* `_assert_public_ip` runs, so that test never touched the IP layer. Renamed it to `..._rejected_by_allowlist` (with `match="allowlist"`) and added `test_redirect_to_allowlisted_host_resolving_private_rejected`: hop-1 (uscode.house.gov) → 302 → an **allowlisted** host (ecfr.gov) whose DNS resolves to `127.0.0.1`, asserting the rejection comes from `_assert_public_ip` (`match="non-public"`). That's the actual redirect-hop rebind surface.

**3 — the premise is incorrect, but it surfaced a real test-fidelity gap.** `httpx.raise_for_status()` *does* raise for any non-2xx including 3xx — I verified 301/304/302 all raise `HTTPStatusError` (this matches your own round-15 "verified-correct" note; the two passes contradicted each other here). So a 301-without-`Location` raises, it does **not** "silently return a body." My *test* was the misleading part: `_mock_stream` mocked `raise_for_status` to a no-op, so the old 304 test showed a body-return that can't happen in reality. Fixed the mock to be faithful (raise for non-2xx) and rewrote the test (now parametrized over 301/304) to assert it raises `HTTPStatusError`.

**2 — done.** Promoted the params-drop-on-redirect constraint from the inline comment into `safe_fetch_bytes`' docstring so it's visible to callers reading the signature.

**4 (strip-list vs keep-allowlist) — declined**, per your own framing that it's an out-of-scope Phase-5 architectural note. The limitation is documented in both the constant and the `safe_fetch_bytes` docstring; inverting to a keep-allowlist would strip legitimate non-credential headers and is a larger change than this PR's threat model needs.

Verified: 63 `test_safe_http.py` pass on the venv; black/isort/flake8 clean. (CI note: the prior red `pytest` was the xdist worker-death timeout, not an assertion failure; this push re-kicks again.)

---

JSv4:
I've verified both findings directly, and **neither holds** — so no code change (acting on either would introduce dead code or a false comment).

**🔴 Finding 1 — the IPv6-tunnel addresses are NOT a gap; the `is_private`/`is_reserved` table is incorrect.** Measured on both Python 3.11.15 and 3.12.3 (`<cgnat-unwrap applied>`):

| address | `is_private` | `is_reserved` | code rejects? |
|---|---|---|---|
| `64:ff9b::10.0.0.1` (NAT64) | False | **True** | ✅ |
| `64:ff9b:1::a00:1` (NAT64 /48) | **True** | **True** | ✅ |
| `2002:6440:1::` (6to4 CGNAT) | **True** | False | ✅ |
| `2002:c0a8:0101::` (6to4) | **True** | False | ✅ |
| `2001:0:4136:…` (Teredo) | **True** | False | ✅ |
| `::100.64.0.1` (IPv4-compat) | False | **True** | ✅ |

All six are rejected by the existing property denylist — exactly what `test_ipv6_embedded_ipv4_tunnels_rejected` pins, and matches my round-8 verification. The review's claim that these are `is_private=False, is_reserved=False` does not reproduce. Two corroborating points: (a) this is the third review pass to assert an IPv6-tunnel "bypass" that measurement refutes; (b) the red `pytest` check was an **xdist worker-death timeout at ~99%**, not an assertion failure — if this test failed 5/6 cases the log would show `F` and a traceback, but it shows only passes/skips before the worker died.

**🟡 Finding 2 — httpx `is_redirect` IS any 3xx; my comment is correct.** Straight from the installed httpx 0.28.1 source:

```python
@property
def is_redirect(self) -> bool:
    """A property which is `True` for 3xx status codes, `False` otherwise."""
    return codes.is_redirect(self.status_code)   # 300 <= code <= 399
```

I confirmed `is_redirect == True` for all of 300–399, including 304. So the rationale in the mock comment ("`is_redirect` is any 3xx, so a 304 would be looped") is accurate, and the mock's `300 <= status_code < 400` faithfully mirrors real httpx. The suggested change (`status_code in (301,302,303,307,308)`) would make the mock *less* faithful than the library it stands in for, so I've left it.

Thanks for the thorough "verified correct" section — those all match my understanding. (CI: the worker-death timeout is the only thing standing between this and green; the latest push re-kicks it.)

---

PR #2050, opened 2026-06-23
Address issue #2020 authority-frontier code review findings

### LINKED ISSUE #2020: ## Code Review — Phase 3: Authority Frontier + Source Provider Registry + USC↔Act Equivalence
## Code Review — Phase 3: Authority Frontier + Source Provider Registry + USC↔Act Equivalence

This is a well-architected, thoughtfully-documented PR that delivers a lot of value. The layered defense against SSRF, XPath injection, and URL construction attacks is solid, and the test coverage (176 tests, offline fixtures, no live API calls) is exemplary. Below are findings ranging from blocking bugs to minor style nits, in priority order.

---

### Bugs / Logic Issues

**1. Race condition in `AuthorityFrontierService.seed_from_wanted_authorities`**

The `get_or_create` + `save` pattern in `authority_frontier_service.py` has a TOCTOU gap: two concurrent seed runs can both pass `get_or_create` and then one silently clobbers the other's count update. Use `update_or_create` with the demand fields in `defaults` to collapse the two round-trips into one atomic operation:

```python
row, was_created = AuthorityFrontier.objects.update_or_create(
    canonical_key=root,
    defaults={
        "authority": authority,
        "jurisdiction": juris,
        "authority_type": atype,
        "mention_count": key_entry["mention_count"],
        "distinct_corpus_count": key_entry["corpus_count"],
    },
)
```

Note this always overwrites `jurisdiction`/`authority_type` on existing rows. If "set once, never overwrite" semantics are needed, wrap the original pair in a `select_for_update` block instead.

**2. `last_error` is never cleared on successful retry**

`authority_frontier_service.py:mark()` only writes `last_error` when `error is not None`. A row that transitions `failed (error="timeout") → ingested` will retain the stale error string in the final state. Downstream callers checking `row.last_error` to determine health will misread a healthy row as broken.

Consider explicitly clearing the field on terminal success states:

```python
if state in ("ingested", "resolved"):
    row.last_error = None
    update_fields.append("last_error")
```

---

### Security

**3. `_USC_PREFIX_RE` / `_CFR_PREFIX_RE` are private-named but cross-module**

`opencontractserver/enrichment/constants.py` defines these with an underscore prefix (signalling internal), yet `cfr_provider.py` and `us_code_provider.py` import them directly. If constants.py is ever refactored the imports break silently. Rename to `USC_PREFIX_RE` / `CFR_PREFIX_RE` (no underscore) or explicitly re-export them.

**4. SSRF domain-allowlist note for Phase 4 (non-blocking)**

The per-provider SSRF guard (`_FR_ALLOWED_HOST`) is correct and well-tested. When Phase 4's centralized client (#1993) lands, confirm the per-provider checks are *removed* (not just shadowed) to avoid two independently-maintained allowlists that can diverge.

---

### Performance

**5. Double `_flatten_element_text` call in CFR provider**

```python
p_parts: list[str] = [
    _flatten_element_text(p_el)
    for p_el in section_el.iter("P")
    if _flatten_element_text(p_el)   # called twice per element
]
```

Each `<P>` element is traversed twice. Fix with a walrus:

```python
p_parts = [t for p_el in section_el.iter("P") if (t := _flatten_element_text(p_el))]
```

**6. `_provider_for` instantiates every provider on every call**

`authority_discovery_service.py:_provider_for` creates an instance of every registered provider to test `can_handle`. Fine for three providers; worth caching or adding a `can_handle_key` classmethod before Phase 4 adds more.

---

### Code Quality / Conventions

**7. Discovery state strings violate the "no magic strings" rule**

`AuthorityFrontier.DISCOVERY_STATE_CHOICES` defines the authoritative state strings, but `mark()` and `discover_and_bootstrap()` reference them as bare literals (`"queued"`, `"in_progress"`, `"failed"`, etc.). CLAUDE.md item 4 applies: move these to `opencontractserver/enrichment/constants.py` so a rename is one-line.

**8. `AuthorityDiscoveryService` imports `ET` and `zipfile` at module level unnecessarily**

These modules are only used in the `except` clause to name provider-level exceptions. Moving them to local imports keeps the orchestrator's dependency surface clean and avoids implying it does XML/ZIP work itself.

**9. `dequeue_for_provider` use case is undocumented**

The normal seed path never sets `provider` until mid-flight inside `discover_and_bootstrap`, so `dequeue_for_provider` returns zero rows for freshly-seeded entries. The method is useful for crash recovery (where `provider` was recorded but the Celery task died), but without a docstring explaining this, future callers will assume it's the primary dispatch entrypoint and find it always returns empty.

**10. Equivalence table direction is implicit**

`AuthorityKeyEquivalence` has a "direction convention" documented only in the seed module's docstring, not in the model. Because all callers query bidirectionally, the direction is documentation-only — but a developer reading the model in isolation won't know that. A one-liner `Note: direction is for documentation only; queries always check both columns` in the model docstring would help.

---

### Test Coverage Gaps

**11. `dequeue_for_provider` missing the `provider=None` negative case**

All `_make_row` calls in `AuthorityFrontierServiceDequeueMarkTests` set a `provider`. There's no test asserting that a freshly-seeded row (`provider=None`) does NOT appear when dequeuing for a named provider — the most common production state.

**12. No concurrent-seed regression test**

Given the race condition in #1, a test using `threading.Thread` or two `atomic()` blocks to simulate simultaneous seeds would serve as a regression guard once the fix lands.

---

### Nitpicks

- `_collect_text` in `us_code_provider.py` silently drops tail text of **excluded** elements (e.g. the text immediately after `</sourceCredit>`). In USLM XML this is typically whitespace or a closing paren, so the behavior is probably correct — a one-line comment confirming the intent would help.
- `AuthorityRequest.params` and `extra` are mutable dicts mutated after creation. A note that they are provider-owned scratch (not aliased between callers) would prevent future bugs.
- `license: ClassVar[str] = "public-domain"  # noqa: A003` in two provider classes. `content_license` or `source_license` would avoid the noqa entirely.

---

### Summary

| Priority | Items |
|---|---|
| Fix before merge | Race condition in seed (#1); stale `last_error` on retry (#2) |
| Address in this PR or follow-up | Private-named public constants (#3); double flatten call (#5); magic state strings (#7) |
| Low / optional | #4, #6, #8–#12 + nitpicks |

The architecture is sound and the security hardening is thorough. Items #1 and #2 are the main asks before merge.

_Originally posted by @claude[bot] in https://github.com/Open-Source-Legal/OpenContracts/issues/1999#issuecomment-4752760847_

JSv4:
Closes #2020.

This addresses the Phase 3 authority-discovery code review (issue #2020). I evaluated all 12 findings + nitpicks against the current code (the PR has since merged and evolved) and fixed the ones still present.

## Status of each finding

| # | Finding | Status |
|---|---------|--------|
| 1 | Race in `seed_from_wanted_authorities` | **Fixed** — atomic `select_for_update` upsert |
| 2 | `last_error` not cleared on successful ingest | **Fixed** — `mark()` clears on success states |
| 3 | `_USC_PREFIX_RE`/`_CFR_PREFIX_RE` private but cross-module | **Fixed** — renamed (no underscore) |
| 4 | Per-provider SSRF allowlist (Phase 4 follow-up) | **Already resolved** — central `safe_http` |
| 5 | Double `_flatten_element_text` traversal | **Fixed** — walrus |
| 6 | `_provider_for` re-instantiates every provider per candidate | **Fixed** — instantiate once per call |
| 7 | Magic discovery-state strings | **Fixed** — single source of truth in `enrichment.constants` |
| 8 | `ET`/`zipfile` module-level imports | Kept — they're the orchestrator's caught-exception contract |
| 9 | `dequeue_for_provider` undocumented | **Fixed** — docstring (crash-recovery purpose) |
| 10 | Equivalence direction implicit in model | **Fixed** — model docstring note |
| 11 | Test gap: `provider=None` negative case | **Fixed** — added test |
| 12 | No concurrent-seed regression test | **Fixed** — deterministic `FOR UPDATE` lock test |

## Key changes

**Blocking bugs**
- `seed_from_wanted_authorities` now does a single `transaction.atomic()` + `select_for_update().get_or_create()` upsert, seeding the real demand counts in `defaults`. This closes the TOCTOU window where two concurrent seeds could drop a count update, and eliminates the transient `mention_count=0` state between the old insert and its follow-up save. The "prefer existing non-null" semantics for `jurisdiction`/`authority_type` are preserved (which is why this isn't a plain `update_or_create`).
- `AuthorityFrontierService.mark()` clears `last_error` when transitioning into a terminal success state (`DISCOVERY_SUCCESS_STATES`, i.e. `ingested`), so a `failed → ingested` retry no longer leaves a stale error that health checks misread. The failure history is retained in the append-only `candidate_sources` audit trail.

**Quality / DRY**
- The discovery-state vocabulary (`DISCOVERY_STATE_*`, `DISCOVERY_STATE_CHOICES`, `DISCOVERY_SUCCESS_STATES`) now lives once in `enrichment/constants.py`. The model's `DISCOVERY_STATE_CHOICES` is re-exported from it with **identical** values, so no migration is generated. The frontier service, discovery orchestrator, crawl driver, and the gate's `GATE_*` verdicts all reference the constants.
- Renamed the USC/CFR prefix regexes to drop the misleading leading underscore (they're imported by the providers).
- `_provider_for` instantiates each enabled provider once per call rather than once per candidate key.

## Verification
- `black`, `isort`/`flake8` (via `flake8-isort`), and `pyupgrade` all pass on the changed files (run with the versions pinned in `.pre-commit-config.yaml`); changelog fragments validate.
- Added regression tests in `test_authority_frontier.py` for findings 1, 2, and 11.
- **Note:** the full Django test suite and `manage.py makemigrations --check` could not be executed in this environment — the container build network intercepts PyPI with a self-signed cert, so the test image can't install dependencies. The no-migration claim rests on Django's autodetector comparing *resolved* choice/default values, which are byte-identical to migration `0100`. CI will run the suite.

---
_Generated by [Claude Code](https://claude.ai/code/session_01B7ka3WfadyJYhbrA1Qpude)_


JSv4:
Thanks for the review. Addressed in f0e2d1a:

1. **Magic strings in new tests** — converted the bare discovery-state literals in the three new tests to `C.DISCOVERY_STATE_*` (fixed the within-method `ingested` constant vs `failed`/`queued` literal mix). Left the surrounding pre-existing tests untouched to keep the change scoped.
2. **`can_handle` purity** — documented on `BaseAuthoritySourceProvider.can_handle` that overrides MUST be stateless, since `_provider_for` now probes one instance with multiple candidate keys and reuses it for `locate`/`fetch`.
3. **`mark()` `error=` + success guard** — added, mirroring the existing exclusivity guards, so the "an ingested row carries no error" invariant fails loudly instead of silently persisting a contradiction.

**Declined the minor nit** (`DISCOVERY_STATE_CHOICES` `list` → `tuple`): the field's `choices` currently match migration `0100`'s recorded list, and no code mutates the list, so I kept it as-is to avoid any migration-state churn for a marginal, theoretical immutability gain. Happy to revisit if you'd prefer the tuple form.

---

Issue #1738, opened 2026-05-21
Continue draining the mypy baseline established in #1331 — test suite + config tail


JSv4:
Picks up the `#1331 → #1335 → #1447` cadence. #1447 graduated the production-code packages (`config.graphql.*`, `llms.*`, `shared.*`, `tasks.*`, `pipeline.*`, `utils.*`, and `corpuses`/`annotations`/`documents`/`agents`/`analyzer`/`discovery`/`conversations.*`). What remains is the tail: a handful of production-code files plus the test suite.

`mypy.ini` currently has **220 `ignore_errors` blocks** left to drain (`docs/typing/mypy_baseline.txt` ≈ 5,600 lines, down from the original 7,208).

### Remaining surface

**Production code (5 files — highest value, do first):**
- `config.settings.{base,local,test_integration}`
- `config.websocket.consumers.{thread_updates,unified_agent_conversation}`

**Worst-offender single file:**
- `opencontractserver.mcp.tests.test_mcp` — ~319 errors, the single largest baseline file; explicitly deferred from #1486. Heavy `MagicMock` usage — graduating it is a chance to audit for real bugs hidden behind auto-attribute mocks (the same pattern that hid bug #1471).

**Test suite (214 modules):**
- `opencontractserver.tests.permissioning.*` — 18 modules
- `opencontractserver.tests.websocket.*` — 2 modules
- `opencontractserver.tests.performance_optimizations.*` — 2 modules
- `opencontractserver.tests.*` (leaf files) — ~192 modules

### Suggested batching

1. **Config tail** — 1 PR for the 5 production-code files. Real runtime code, small surface.
2. **`test_mcp`** — 1 dedicated PR (large, audit-worthy).
3. **`tests.permissioning.*`** — 1 PR for the cohesive 18-file subpackage.
4. **`tests.*` leaf files** — chunk into 4–6 PRs grouped by domain (corpus / document / annotation / agent / extract / misc). Most share the same `setUpTestData` class-attribute pattern; see #1479 for the recommended lightest-touch fix.

### Per-package workflow (unchanged from #1447)

1. Enumerate current errors for the package.
2. Fix or annotate (prefer real annotations over `# type: ignore`; where ignores are needed, scope them with an error code).
3. Remove the package's `[mypy-...]` block(s) from `mypy.ini` and prune the corresponding lines from `docs/typing/mypy_baseline.txt`.
4. Document the delta in `CHANGELOG.md`.

### Acceptance

[...]

Issue #1335, opened 2026-04-20
Mypy: type analyzer, shared, agents; introduce Protocols for pluggable interfaces


JSv4:
## Context

Several smaller packages are under 50% coverage but contain interfaces that other packages rely on — they're a good place to invest in `Protocol` and `TypedDict` definitions rather than just raw annotations.

| Package | Files | Defs | Annotated | Coverage |
|---|---:|---:|---:|---:|
| `analyzer/` | 13 | 16 | 2 | 12.5% |
| `shared/` | 11 | 84 | 32 | 38.1% |
| `agents/` (legacy) | 5 | 32 | 11 | 34.4% |
| `badges/` | 7 | 17 | 8 | 47.1% |
| `worker_uploads/` | 9 | 28 | 13 | 46.4% |

Project-wide, only 2 files use `typing.Protocol` (`llms/types.py`, `llms/agents/core_agents.py`). The pipeline has real protocol-shaped interfaces (`BaseParser`, `BaseEmbedder`, `BaseThumbnailer`) but they're expressed only via abstract base classes. A `Protocol` layer would let type-checkers verify dynamic plugin registration without forcing inheritance.

## Scope

1. **Annotate target packages** to ≥70% coverage using the patterns established in #1332 / #1333 / #1334.

2. **Introduce shared `Protocol` types** in `opencontractserver/types/protocols.py`:
   - `VectorStoreProtocol` — used by `llms/vector_stores/` and pipeline embedders
   - `PipelineComponentProtocol` — parser / embedder / thumbnailer discovery
   - `ToolProtocol` — framework-agnostic `CoreTool` interface (see `docs/architecture/llms/README.md`)
   - `PermissionedQueryManagerProtocol` — what `.visible_to_user()` is guaranteed to return

3. **Promote bare `dict` / `list` / `set` in public signatures** to the correct parametrised form. There are ~81 such occurrences, concentrated in `tasks/`, `utils/pdf_tools.py`, and `shared/`. A mechanical sweep with manual review of edge cases.

4. **Audit the 48 existing `# type: ignore` comments**. Most are in `llms/` (17) and split across the rest. For each:
   - Can it be fixed at the root (missing annotation upstream)?
   - If genuinely third-party, scope it narrowly (`# type: ignore[attr-defined]` not bare `# type: ignore`).
   - Document any that remain with a one-line comment explaining why.

## Target

[...]

PR #2051, opened 2026-06-23
typing: graduate the conversation & threading test domain out of the mypy baseline

JSv4:
## Summary

Next increment of **#1738** (the `#1331 → #1335 → #1447` mypy-baseline drain). After the config tail, `tests.permissioning.*`, and the annotation / document / corpus / agent / extract / pipeline leaf-file chunks, this graduates the next cohesive domain from the issue's `tests.*` batching plan: the **chat / threads** feature.

Removes the **7** `[mypy-opencontractserver.tests.test_*]` `ignore_errors` blocks:

- `test_conversation_mutations_graphql`
- `test_conversation_permissions`
- `test_conversation_query`
- `test_conversation_search`
- `test_long_conversation_api`
- `test_thread_corpus_actions`
- `test_threading`

Prunes the corresponding **433 lines** from `docs/typing/mypy_baseline.txt` (3375 → 2942) and fixes the **361** errors that actually surface. The baseline had drifted from reality: `test_conversation_search` fell 77 → 23 (its historical `set_permissions_for_obj_to_user` arg-type errors no longer reproduce under `django-stubs==6.0.5`), while `test_conversation_permissions` grew 86 → 159 after recent edits.

## How

Established patterns from prior chunks:
- **Class-level annotations** for `setUpClass`/`setUpTestData` attributes (the recommended fix from #1479) — the bulk of the errors.
- **Graphene `self.client` → `self.graphene_client` rename** (3 files): a `graphene.test.Client` assigned to `self.client` shadows `django.test.TestCase.client`, which has no `.execute()`.
- **`assert ... is not None`** narrowing of Optional ORM / embedding returns.
- The three `setUpClass`-heavy files swapped their module-level `User = get_user_model()` alias for the concrete `from opencontractserver.users.models import User` import, since mypy rejects a `get_user_model()` *variable* as a type annotation.

## One real cleanup surfaced: `_skip_signals`

`_skip_signals` is an out-of-band flag that tests/fixtures and one production path (`llms/tools/moderation_tools.py`) set on model instances so the signal handlers in `*/signals.py` skip their side effects (notifications, badge awards, corpus-action triggers). It was **undeclared**, which forced a `# type: ignore[attr-defined]` in `moderation_tools.py` and produced 26 `attr-defined` errors in `test_thread_corpus_actions`.

Declared once as `_skip_signals: bool` on **`InstanceUserCanMixin`** (`opencontractserver/shared/user_can_mixin.py`) — the shared instance-side base for *both* `BaseOCModel` and the `TreeNode`-rooted `Corpus`/`CorpusFolder`, so the flag is typed across both model lineages. It's a **type-only bare annotation** that Django's model metaclass ignores at runtime, so the presence-based `hasattr`/`getattr` guards behave identically. This removes the existing ignore, clears all 26 test errors, and future-proofs the other still-baselined tests that use the flag (on `Conversation`/`ChatMessage`: `test_badges`, `test_leaderboard`, `test_notifications`; on `Corpus`: `test_caml_intelligence_block`).

## Verification

- ✅ `mypy --config-file mypy.ini opencontractserver config` → **clean** under both the pre-commit pin (`mypy==2.0.0`) and CI's pin (`mypy==2.1.0`).
- ✅ All **7** graduated test modules pass (run against a pgvector Postgres + Redis).
- ✅ `black`, `isort`, `flake8` clean on the changed files.
- ✅ Changelog fragment added (`changelog.d/1738-conversation-tests-mypy.changed.md`).

Refs #1738


JSv4:
Thanks for the review — addressed in `ae8c23a`. Point-by-point:

**1. `Corpus` not covered — valid, fixed.** Moved the declaration from `BaseOCModel` to `InstanceUserCanMixin` (`shared/user_can_mixin.py`), the documented single instance-side base for *both* `BaseOCModel` and the `TreeNode`-rooted `Corpus`/`CorpusFolder`. One annotation now types the flag across both lineages, so a future graduation of `test_caml_intelligence_block` (which sets `corpus._skip_signals`) is covered too.

**2. Presence-based semantics — documented (declined the broader refactor).** The comment now spells out the protocol explicitly: set to `True` to suppress; most handlers use `hasattr`, so assigning `False` does **not** re-enable — `del instance._skip_signals` is the reliable re-enable. I deliberately did **not** normalize the ten signal sites to `getattr(..., False)` in this PR: that's a behavior change to production signal handlers (it would make `= False` actually re-enable), it's latent/harmless today (every caller sets `True`), and it's well outside a mypy-baseline-graduation PR. Worth its own issue if we want the value-based protocol.

**3. `ClassVar[bool]` — declined, it breaks the call sites.** `ClassVar` forbids assignment through an instance, and every use here is exactly that (`conversation._skip_signals = True`, etc.). Verified with mypy 2.1.0:
```
error: Cannot assign to class variable "_skip_signals" via instance  [misc]
```
The bare `bool` instance-attribute annotation is the correct form; I added the runtime-absence caveat to the comment to cover the "future refactor trusts the type" concern.

**4. Inline comment in `moderation_tools.py` — kept intentionally.** It documents the *local* intent (suppressing **re-triggering of additional corpus actions** from inside corpus-action handling — i.e. recursion avoidance), which the general mechanism docblock on the mixin doesn't convey. It's call-site rationale, not duplication.

**5. Double assertion — declined, `assertIsNotNone` does not narrow.** mypy does not treat `unittest`'s `assertIsNotNone` as a narrowing guard — that's *why* the original errors existed (`test_conversation_search.py:151` errored despite `assertIsNotNone(embedding)` on line 150). The standalone `assert` is required for narrowing, and the `assertIsNotNone(...)` + `assert ... is not None` pairing is the established pattern from the prior pipeline-chunk graduation (`test_pipeline_registry.py`). Keeping both preserves the descriptive test-failure message *and* the type narrowing.

mypy stays clean under both pins (`2.0.0`/`2.1.0`) and all 7 modules still pass after the relocation.

---

JSv4:
Thanks — second pass addressed in `3f12d6e` + a PR-body edit.

**1. `notifications/signals.py:146` `getattr` vs `hasattr` — declined (out of scope).** This is pre-existing production code I don't touch in this PR. Worth noting the *first* automated review on this PR asked for the **opposite** normalization (move all sites to `getattr(..., False)`); the two reviews contradict each other, which is a good sign this is a maintainer design call rather than a clear-cut fix. It's also latent today — no caller ever assigns `_skip_signals = False`. My mixin comment documents the current reality accurately ("most handlers use `hasattr`, so `False` does not re-enable"). Happy to file a separate issue to make the protocol uniform if you want it converged.

**2. Bare `assert` in the two registry-loop tests — fixed.** Added `self.assertIsNotNone(tool, f"Tool {tool_name} should be registered")` before the narrowing `assert` in both `test_moderation_action_tools_require_approval` and `test_read_only_tools_no_approval`, matching the sibling `test_moderation_tools_registered`. Now a missing tool fails with the tool name rather than a bare `AssertionError`.

**3. `mypy.ini` narrative comment — declined (matches established convention).** This file already carries **12** prior `# Issue #N graduated …` narrative blocks (the #1332/#1485/#1484/#1486/#1477/#1482 graduations and the earlier #1738 chunks). I matched that convention deliberately; making this one block terse while the other twelve stay verbose would be the inconsistency. (FWIW the "Default to writing no comments" line quoted isn't in this repo's `CLAUDE.md`.) If the convention should change, that's a separate sweep across all blocks rather than this PR.

**4. PR body credited Claude — fixed.** Removed both footer lines per `CLAUDE.md` rule #3. Same edit also corrected the stale `BaseOCModel`/`Models.py` reference to `InstanceUserCanMixin`/`user_can_mixin.py` (your first minor note).

**Minor — `test_conversation_search` `retrieved_vector`:** keeping the bare `assert` there. The original test already had `assertIsNotNone(embedding)` (so I kept it + added the narrowing assert = double-guard), but it had **no** None-check for `retrieved_vector` — the real assertion is `assertEqual(len(retrieved_vector), 384)`. Since it's a single non-looped call, the failing line is self-identifying, so the added `assert` is purely for narrowing without changing the test's original check.

mypy stays clean under both pins and `test_thread_corpus_actions` passes after the change.

---

JSv4:
Thanks — addressed in `bfea061`.

**1. Incomplete protocol doc — fixed.** The `_skip_signals` comment now names the exception explicitly: `notifications/signals.py`'s `emit_analysis_status_notification` reads `getattr(instance, "_skip_signals", False)` and so is value-sensitive, unlike the eight `hasattr` handlers. (Took the documentation option rather than standardizing the handlers, since that's a production behavior change out of scope here — and an earlier review on this PR asked for the *opposite* standardization direction.)

**2. PR body / `mypy.ini` name the wrong class — already fixed (stale).** Both were corrected in `ae8c23a` + the PR-body edit last round; they currently read `InstanceUserCanMixin` (`shared/user_can_mixin.py`). This review looks to have run against an earlier commit. No further change needed.

**3. Inconsistent `assertIsNotNone` + `assert` — keeping the current rule (declined).** The application is principled, not arbitrary: where the original test already had a descriptive `assertIsNotNone` (or the call is in a loop where the message aids diagnosis), I keep it **and** add the bare `assert` for narrowing; where there was no prior None-check, the real assertion is the following line (`len(...)`, subscript) and the bare `assert` is purely for mypy. I'd avoid the suggested `assert x is not None, "msg"` one-liner specifically because `assert` is stripped under `python -O` — that's the exact -O fragility the *previous* review flagged, and `assertIsNotNone` (not stripped) is the more robust runtime guard. Converting the existing `assertIsNotNone` calls would also rewrite pre-existing test assertions. The double-guard matches the established pipeline-chunk graduation pattern.

**4. Latent `TypeError` catch — thanks, noted.** Agreed it's a pre-existing latent bug the mypy work surfaced; the `assertIsNotNone(tool, …)` guard (now in both loops, `3f12d6e`) turns it into a clear assertion failure.

mypy stays clean under both pins; the change is documentation-only.

---

PR #2075, opened 2026-06-27
Add LiteParse local PDF ingestion engine (#2074)

### LINKED ISSUE #2074: Add Liteparse Ingestion Engine
Add Liteparse as ingestion engine

JSv4:
Closes #2074

## What

Adds **`LiteParseParser`** — a third, fully-local PDF ingestion engine built on [LiteParse](https://github.com/run-llama/liteparse) (run-llama, PDFium-based with optional Tesseract OCR) — alongside the existing Docling and LlamaParse engines. It auto-registers via the pipeline registry and is selected with `PDF_PARSER=liteparse`.

## Following liteparse's actual API

Verified against liteparse's Python source (`packages/python/liteparse/types.py` + `parser.py`):

- `from liteparse import LiteParse`; `LiteParse(**opts).parse(bytes|path) -> ParseResult`
- `ParseResult.pages` → `ParsedPage(page_num` 1-indexed`, width, height, text, text_items)`; `ParseResult.text`
- `TextItem(text, x, y, width, height, font_name, font_size, confidence)` — **absolute PDF points, top-left origin** (same convention as pdfplumber), line-level, with **no element-type labels or hierarchy** in the structured output.

## Matching the core outputs of the other engines

Because liteparse's JSON gives line bboxes + text only, the parser augments it to produce the same surface (bboxes, tokens, feature labels, parent-child) as Docling/LlamaParse:

| Output | How |
|--------|-----|
| **Bounding boxes** | LiteParse `TextItem` coords (clamped; no fractional conversion needed) |
| **Tokens** | Word-level PAWLs tokens via pdfplumber (`extract_pawls_tokens_from_pdf`, same path as LlamaParse); each line bbox mapped to enclosed word tokens with a shapely spatial index |
| **Feature labels** | Derived from `font_size`: modal size = body (`Text Block`); larger sizes ranked into heading levels → `Title` / `Section Header` |
| **Parent-child** | A heading stack walked in reading order sets `parent_id`, fed to `import_annotations` (self-FK) and `build_subtree_groups_for_document` (`OC_SUBTREE_GROUP` relationships) |
| **Images** | Embedded images via `extract_images_from_pdf`, added to the unified token array + emitted as `Image` annotations |

Heading detection is heuristic and can be disabled (`detect_headings=False`), in which case output is flat `Text Block`s with no parent (matching LlamaParse's flat output).

## Changes

- `opencontractserver/pipeline/parsers/liteparse_parser.py` — the new parser (lazy `liteparse` import so registry discovery/startup don't require it).
- `config/settings/base.py` — `LITEPARSE_*` settings, `_PDF_PARSER_MAP["liteparse"]`, `PARSER_KWARGS` entry, and a `_SELECTED_OFFICE_PARSER` fallback so DOCX/PPTX route to Docling when LiteParse (PDF-only) is selected. Default/test config is unchanged (the fallback only diverges when `PDF_PARSER=liteparse`).
- `requirements/ingestors/liteparse.txt` — optional dependency.
- `opencontractserver/tests/test_doc_parser_liteparse.py` — covers token mapping, font-size feature labels, the parent-child hierarchy chain, headings-disabled, image extraction, error paths, configuration, and registry discovery.
- `docs/pipelines/liteparse_parser.md` + `docs/pipelines/pipeline_overview.md` — documentation.
- `changelog.d/2074-liteparse.added.md` — changelog fragment.

## Testing

`black`, `isort`, `flake8`, and the changelog-fragment checker pass on all changed files. The Docker/Postgres-backed Django suite could not be run in this environment (no Docker/Django available); the new tests mock `liteparse.LiteParse` and the pdfplumber-backed extraction utilities, and follow the existing `test_doc_parser_llamaparse.py` patterns.

---
_Generated by [Claude Code](https://claude.ai/code/session_01KAbSFKaSYKEd5bG2vPmKtc)_


JSv4:
Thanks for the review — addressed in 29c0254 (fixes) and 18aea12 (coverage). Disposition of each item:

**Fixed**
- **1 (heading kwargs dropped)** — `detect_headings` / `heading_size_ratio` are now resolved in `_parse_document_impl` and threaded through `_convert_result_to_opencontracts` → `_classify_heading_sizes` (falling back to instance settings when `None`). Added a regression test asserting a `detect_headings=False` call-time kwarg overrides an instance configured with detection on.
- **2 (missing image env vars)** — added `LITEPARSE_IMAGE_FORMAT` / `LITEPARSE_IMAGE_QUALITY` / `LITEPARSE_MIN_IMAGE_WIDTH` / `LITEPARSE_MIN_IMAGE_HEIGHT` to `base.py` so the `env_var` declarations are honored by `migrate_pipeline_settings`.
- **3 (page_count vs target_pages)** — export `page_count` now uses `len(pawls_file_content)`, the value `save_parsed_data` persists, so the two agree.
- **4 (degenerate bbox at page edge)** — `_bounds_from_item` now expands inward when an item sits exactly on the page edge; added a unit test.
- **6 (magic numbers)** — moved the fallback US-Letter dimensions to `constants/document_processing.py` (`DEFAULT_PDF_PAGE_WIDTH/HEIGHT`) and import them. (Left the pre-existing copies in `docling_parser_rest.py` / `compact_pawls.py` untouched as out of scope for this PR; noted in the constant's docstring.)

**Not changed (intentional)**
- **5 (field-by-field copy in `__init__`)** — kept, to match the established `LlamaParseParser` / `DoclingParser` pattern; `self.settings` can be `None` (it falls back to `Settings()`), so the explicit copy is deliberate rather than incidental.
- **7 (unbounded `liteparse>=0.1.0`)** — agree the `_attr()` helper degrades quietly on an upstream rename. I left the version unpinned for now because I can't confirm liteparse's current published version from here, and a wrong `<0.2` ceiling would make it uninstallable. Flagging for a maintainer to set the right upper bound against the actual release.

Also added tests covering the token-/image-extraction failure fallbacks and invalid-page-metadata paths to bring patch coverage back over the gate.

---

JSv4:
Thanks for the second pass — addressed in d3304eb.

**Fixed**
- **image_format not runtime-validated** — clamped to `{jpeg,png}` at use-time in `_append_image_tokens` (`cast` is type-checker-only, as you noted); a `webp` value now falls back to `jpeg` instead of writing PNG bytes tagged `.webp`. Added a test asserting the clamp.
- **`PARSER_KWARGS` holes** — expanded the entry to every setting `_parse_document_impl` actually resolves per-call (`dpi`, `num_workers`, `target_pages`, `max_pages`, `password`, `ocr_language`, `ocr_server_url`, `image_mode`, `heading_size_ratio`, plus the original four). I intentionally left `image_format`/`image_quality`/`min_image_*` out: those are read from the resolved settings inside `_append_image_tokens`, *not* per-call, so listing them in `PARSER_KWARGS` would imply an override that doesn't fire — this matches `LlamaParseParser`, whose image settings are likewise absent from its `PARSER_KWARGS` entry (so the "mirrors every `LLAMAPARSE_*`" comparison isn't quite accurate). Operators still configure those via env → `PipelineSettings`.
- **`"documents/"` magic string ×3** — added `DOCUMENT_IMAGE_STORAGE_PREFIX` to `constants/document_processing.py` and used it in the liteparse, llamaparse, and docling parsers (byte-identical output).
- **O(n) scans in `_classify_item`** — `_classify_heading_sizes` now also yields a `size → level` map (built once in the converter) so per-item classification is an O(1) dict lookup.
- **intermediate `sizes` list** — `_classify_heading_sizes` accumulates into a `Counter` directly; memory is now proportional to distinct font sizes.

**Declined (out of scope)**
- **Extract `_create_annotation` / fallback-page / image-token blocks to `pipeline/base`** — a real duplication, but doing it means refactoring the well-tested `LlamaParseParser` path (and its image/figure handling, which is a functional superset in places). That's a cross-cutting change I'd rather not fold into a "add an engine" PR; better as a dedicated follow-up so the shared helper can be reviewed against both parsers' full behavior.

All changes are `black`/`isort`/`flake8`-clean.

---

JSv4:
Thanks — addressing the last two review passes together (they overlap) in cb3c0bb.

**Fixed**
- **Fallback page misindexing with `target_pages`** (flagged in both passes) — confirmed real. The fallback page synthesis (used only when pdfplumber token extraction raises) compacted the list by `sorted(page_dimensions)`, breaking the "list position == absolute page index" invariant the success path guarantees, so `_append_image_tokens` (which indexes by absolute page number) wrote images into wrong entries or dropped them once `target_pages` started at a non-zero / non-contiguous index. Fixed at the root: the fallback now builds over `range(max_idx + 1)`, filling gaps, so positions always align in both paths (this also covers the `'3,5'` → keys `{2,4}` case). Added a regression test with a single parsed page at absolute index 1.
- **llamaparse local `612`/`792`** — now imports `DEFAULT_PDF_PAGE_WIDTH`/`HEIGHT`, completing the centralization this PR introduced.

**Clarified (no behavior change)**
- **`LITEPARSE_DPI` not forwarded to `extract_images_from_pdf`** — by design: `LITEPARSE_DPI` is LiteParse's own page-render DPI; embedded-image extraction uses the pipeline-wide `IMAGE_EXTRACTION_DPI` default, exactly as Docling and LlamaParse do (neither couples its parser DPI to image extraction). Added a comment documenting the decoupling rather than changing behavior and diverging from the other two parsers.

**Declined (cross-cutting; better as a dedicated follow-up)**
- **Centralize the structural-label strings** (`"Title"`, `"Section Header"`, `"Text Block"`, `"Image"`) — these are already named module constants here, not inline literals. A true single-source-of-truth means defining the *canonical structural-label vocabulary* and migrating all parsers — LlamaParse's `ELEMENT_TYPE_MAPPING` has ~17 such labels and Docling's come from the microservice. Centralizing only the 4 that happen to overlap would leave LlamaParse's mapping half-literal/half-constant (worse, not better). Worth doing as its own change against all parsers.
- **Shared `patch_parser_settings` test helper** — extracting it would mean editing the pre-existing `test_doc_parser_llamaparse.py` (different field set); per the repo's "don't touch old tests without permission" guidance I'd rather leave it and revisit when a third parser test lands.

All changes are `black`/`isort`/`flake8`-clean.

---

JSv4:
Thanks — addressing the last two review rounds together (they overlap) in e9c8eee.

**Fixed**
- **🔒 PDF password / API key leaking into logs** (good catch) — `BaseParser.parse_document` logged the merged kwargs unredacted at INFO before the per-parser impl could redact, exposing SECRET settings (the new LiteParse `password`, and the *pre-existing* LlamaParse `api_key`). Now redacted via `redact_sensitive_kwargs` in the base log line, which fixes it for every parser at once.
- **Heading detection on heading-heavy / footnote-heavy docs** (raised in both rounds) — switched body-size detection from line **frequency** to **character mass**. Body prose dominates a document's characters, so this correctly picks body even when heading-style lines outnumber it (slide decks, TOCs, repeated exhibit labels) or when small footnotes are as frequent as body — the failure modes of the old most-frequent + `min()` tiebreak. Updated the hierarchy test to realistic line lengths and added a unit test for the heading-heavy/footnote case.
- **Image-token exception safety** — `_append_image_tokens` now skips image dicts missing required keys, so a malformed entry can't raise mid-loop and strand partial tokens the `except` can't roll back (the append mutates `pawls_pages` in place).
- **Office-format fallback future-proofing** — replaced the inline `== "liteparse"` check with a named `_PDF_ONLY_PARSER_NAMES` set, so adding another PDF-only engine is a one-line change rather than a silent DOCX-misrouting bug. (Introspecting each parser's `supported_file_types` here isn't viable — settings load before the app registry/models are importable.)
- **Dead `else` branch** in the import-error test cleanup — removed.

**Clarified, no behavior change**
- **Image tokens on pages outside `target_pages`** — documented in a comment: like the word-token pass, image extraction spans the whole PDF, so with `target_pages` the full token layer (text *and* image) is present but only parsed pages are annotated. Unreferenced tokens are harmless (token indices are per-page, so nothing shifts). Restricting images but not text would make the two inconsistent.
- **`image_*` absent from `PARSER_KWARGS`** — already has an explanatory comment above that block (added in an earlier commit): those are read from resolved settings in `_append_image_tokens`, not per-call.

**Declined (dedicated follow-up)**
- **Extract `_create_annotation` / `_build_image_token` to a shared base/util** — real duplication, but unifying them correctly means reconciling per-parser variants (e.g. Docling's `image_type="cropped"` default, LlamaParse's figure handling) and re-running all three engines' suites. Better as a focused refactor PR than as collateral in a feature PR.
- **Centralize structural-label strings** and the **dataclass/env/test default trio** — the labels need a canonical cross-parser vocabulary (LlamaParse's mapping has ~17), and the default-value trio is the intended settings-schema shape (dataclass default is canonical; env/test mirror it) shared by every parser. Both are cross-cutting and out of scope here.

All changes `black`/`isort`/`flake8`-clean.

---

JSv4:
Good catches — addressed in 63dc0cc.

**Fixed**
- **Image `tokenIndex` misalignment when an image is skipped** — correct, and it was a regression from the malformed-image guard I added in the previous commit: I skipped malformed images when appending tokens but still stored the *full* raw list in `images_by_page`, so `token_offset + img_idx` pointed past the real slot for every image after a skipped one. Now I store only the successfully-appended images (in append order), so annotation indices stay aligned. Added a regression test (malformed first image must not shift the second image's `tokenIndex`).
- **Micro-font watermark** — excluded sub-point sizes (`< MIN_CONTENT_FONT_SIZE = 1.0`) from heading-size detection so a 0.1pt watermark with many characters can't be picked as body and drag the threshold to ~0. (Note: the suggested `max(body_size, 1.0)` floor on the *threshold* wouldn't fix it — real 11pt text would still exceed `1.2` and be classed a heading; filtering the watermark out of body candidacy is the actual fix.) Added a unit test.
- **docling `612`/`792`** — now imports `DEFAULT_PDF_PAGE_WIDTH/HEIGHT` for all three inline literals, completing what the constant's docstring described; docstring updated.

**Declined (dedicated follow-up)**
- Extracting `_build_image_token` / `_create_annotation` to a shared base/util, and the shared `create_mock_token_extraction_result` / `patch_parser_settings` test helpers — all real duplication, but each refactors the well-tested LlamaParse/Docling paths (or their test files), so they belong in a focused refactor PR that can reconcile the per-parser variants and re-run all three engines' suites, rather than as collateral churn here.

All changes `black`/`isort`/`flake8`-clean.

---

JSv4:
Addressed in 57d1725:
- **Conversion vs parse failures** — `_convert_result_to_opencontracts` now runs in its own `try/except` so a pdfplumber/shapely/assembly failure is logged distinctly instead of as "LiteParse parsing failed". Added a test for the conversion-failure path.
- **Double INFO log** — removed the redundant per-impl log (and its unused import); `BaseParser.parse_document` already logs the redacted merged kwargs.
- Commented the image-page out-of-range guard.

Declined (rationale in the commit): a test enforcing `_PDF_ONLY_PARSER_NAMES` vs `supported_file_types` (those private settings vars aren't exposed via `django.conf.settings`; the in-code comment already flags the sync requirement), and converting test-fixture `612`/`792` to constants (concrete fixture dimensions, not references to the production default).

---

JSv4:
Addressed in de867a5:
- **Image annotation bounds not clamped** (real bug) — image annotations now build bounds via `_bounds_from_item`, so an image whose pdfplumber stream bbox bleeds past the page edge is clamped to the page (same as text lines) instead of storing out-of-page coordinates. Also dedups the bounds-building.
- **Version pin** — ⚠️ correcting the record: `liteparse`'s latest published release is **2.2.1**, not 0.1.x (it shipped breaking `0.x → 1.x → 2.x` bumps). The suggested `<0.2.0` cap would be **uninstallable**. More to the point, my own placeholder `>=0.1.0` was wrong — this parser targets the **2.x** Python API (`LiteParse` / `ParseResult.pages` / `ParsedPage.text_items` / `TextItem` fields), so I've pinned `>=2.2.1,<3.0.0`.

Declined (unchanged rationale — these refactor the well-tested LlamaParse/Docling paths or are parser-specific): backporting the malformed-image guard to LlamaParse, extracting `_create_annotation` / `_build_image_token` to a shared base, and moving `MIN_CONTENT_FONT_SIZE` (a liteparse-only heading-heuristic constant, not a cross-parser value) to the shared module. The duplications are tracked for a dedicated follow-up PR.

---

JSv4:
On #1 — you're right that the tiebreak should favor the larger size; switched `min()` → `max()` in 28c8dd3 (reversing the choice I'd defended earlier). The precise reason is the failure mode: on an exact char-mass tie, picking the smaller size lowers the threshold and risks **over-detecting** headings (real body paragraphs → Section Headers), whereas picking the larger size raises it and fails toward **under-detection** (flat output, same as headings-off) — the benign error. It only affects exact ties (degenerate/short inputs); normal docs have a unique body maximum. Updated/renamed the tie test.

The other items I'm leaving as-is, with the same rationale given earlier: test-fixture `612`/`792` are concrete fixture dimensions (not refs to the prod default); the `DEFAULT_WIDTH` alias is a deliberate local-brevity binding (and LlamaParse uses the same one, so it's not inconsistent); `STRtree` under `TYPE_CHECKING` saves nothing since shapely is imported regardless; and the `password` logging risk is already handled by `redact_sensitive_kwargs`. The remaining DRY extractions stay tracked for a dedicated follow-up.

---

JSv4:
No code change for this pass — evaluated all six, none are actionable:

- **#1 (numpy import):** non-issue. numpy is a core transitive dependency (pdfplumber/shapely/pandas/opencv all require it), and this module imports `shapely.strtree.STRtree` at top level — which itself needs numpy — so the import can't fail in any working install. The `np.ndarray` annotation is also accurate (those values are numpy arrays from `extract_pawls_tokens_from_pdf`).
- **#2 (image loss with `target_pages` + extraction failure):** the skipped pages are *outside* the requested range, so skipping is correct, not loss. For `target_pages='5-7'` the fallback `range(7)` covers exactly 0-indexed pages 4–6 (the requested ones); only pages 8–10 get skipped. A warning there would fire on every `target_pages` run for expected behavior. Already documented in a code comment.
- **#3/#4/#5 (DRY: shared `_create_annotation`/`_build_image_token`, the `DEFAULT_WIDTH` alias, `_attr`→utils):** the standing declines — the cross-parser extractions are tracked for a dedicated follow-up; the alias is a deliberate local-brevity binding LlamaParse also uses.

Thanks for the confirmations on the non-issues (heading-stack ordering, redaction coverage, bounds clamping, float-vs-int). The PR is functionally complete; remaining suggestions are all cross-parser refactors better done in their own change.

---

PR #2088, opened 2026-06-29
Secure structural annotation document resolution

JSv4:
### Motivation
- Structural annotations with `document_id=NULL` were resolving their `document` via shared structural sets without applying `Document.visible_to_user`, which could leak private document metadata and signed file URLs through GraphQL nested fields.
- The change aims to ensure any `Document` returned for an annotation is permission-checked for the requesting user and to keep corpus/document scoping deterministic.

### Description
- Make `AnnotationType.resolve_document` intersect candidate documents with `Document.visible_to_user` (via `BaseService.filter_visible`) for direct `document_id`, prefetched structural-set documents, and fallback lookups so only accessible documents are returned.
- Extend `AnnotationService.structural_document_prefetch` to accept `user` and `context` and, when provided, prefetch only documents visible to that user by using `BaseService.filter_visible` while preserving existing corpus/document scoping and ordering.
- Update GraphQL annotation and search query paths to pass `info.context.user` and `info.context` into `structural_document_prefetch` so prefetches are permission-scoped.
- When a structural-set prefetch exists but contains no in-context visible documents, the resolver now returns `None` instead of falling back to a different structural-set member.

### Testing
- Compiled the modified modules with `python -m compileall config/graphql/annotation_types.py config/graphql/annotation_queries.py config/graphql/search_queries.py opencontractserver/annotations/services/annotation_service.py` which completed successfully.
- Attempted to run `pytest opencontractserver/tests/test_corpus_cards_structural_document_resolution.py` but the test run could not execute in this environment due to a missing dependency: `ModuleNotFoundError: No module named 'django'`.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a41c74a0be8832898c4b70d54d29b09)


JSv4:
Closing in favor of #2089, which fixes the same structural-annotation document-visibility IDOR in resolve_document. The SemanticSearchResultType.resolve_document surface flagged in this PR's review will be folded into #2089.

PR #2112, opened 2026-07-05
Add dedicated bulk embeddings pool for ingest

JSv4:
## Summary

Ingest embedding tasks and search queries currently share a single embeddings
microservice URL (`EMBEDDINGS_MICROSERVICE_URL`). That forces a compromise: a
pool sized for latency-sensitive **search queries** (needs to stay warm) also
absorbs the load of **batch ingest** (thousands of embeddings, happy to hit an
autoscaled / scale-to-zero pool), and vice versa.

This PR lets operators point ingest at a separate **bulk** pool while query call
sites stay on the always-warm pod, fully isolating search latency from ingest
load — with **no change to the embedder client, base class, or any search
resolver**.

## Changes

1. **New setting** — `EMBEDDINGS_MICROSERVICE_URL_BULK` in
   `config/settings/base.py` (next to `EMBEDDINGS_MICROSERVICE_URL`), defaulting
   to `EMBEDDINGS_MICROSERVICE_URL` so single-pool deployments need no config.
   Query call sites are untouched and stay warm automatically.

2. **Ingest routing** — the Celery tasks in
   `opencontractserver/tasks/embeddings_task.py` resolve the bulk URL via a new
   `_bulk_embeddings_service_url()` helper and thread it through an optional
   `service_url_override` parameter added to the four ingest leaves —
   `_create_text_embedding`, `_create_embedding_for_annotation`,
   `_batch_embed_text_annotations`, `_apply_dual_embedding_strategy` — plus
   `_embed_relationship`. The override is translated into the embedder's
   **existing** call-time `embeddings_microservice_url` kwarg (read by
   `MicroserviceEmbedder._get_service_config`), so no embedder/base-class change
   is required. Embedders that don't read that kwarg (hosted providers, the
   multimodal image pool) simply ignore it.

3. **Backwards compatible by construction** — when the override is `None`
   (the default, and any direct/legacy caller), **no** URL kwarg is passed and
   behavior is byte-identical to before. The dual-embedding strategy only
   forwards the keyword to `embed_func` when a bulk URL is set, so the original
   three-argument `embed_func` contract keeps working.

4. **Multimodal image pool left alone** — the bulk setting is text-only; image
   embedding keeps its own `CLIP_EMBEDDER_URL` / `QWEN_EMBEDDER_URL` pool.

### Deployment note

Rebuild/redeploy the Django image so Celery workers pick up the new code and
setting. Set `EMBEDDINGS_MICROSERVICE_URL_BULK` in the environment to point
ingest at the bulk pool; leaving it unset keeps the current single-pool
behavior.

## Tests

`opencontractserver/tests/test_batch_embedding.py` gains regression tests
verifying:
- `_batch_embed_text_annotations` forwards `service_url_override` as
  `embeddings_microservice_url`, and passes **no** URL kwarg when unset.
- The full `calculate_embeddings_for_annotation_batch` task reads
  `EMBEDDINGS_MICROSERVICE_URL_BULK` (via `override_settings`) and routes ingest
  to it end-to-end.

## Docs / changelog

- `docs/deployment/performance_tuning.md` — new "Separate bulk embeddings pool
  for ingest" section.
- Sample env files updated with the new (commented, optional) variable.
- Changelog fragment `changelog.d/bulk-embeddings-pool.added.md`.

---
_Generated by [Claude Code](https://claude.ai/code/session_01RSJewSJz3p4sqancvmoQ8E)_

### DIFF
--- a/changelog.d/bulk-embeddings-pool.added.md
+++ b/changelog.d/bulk-embeddings-pool.added.md
@@ -1,0 +1,12 @@
+- **Separate "bulk" embeddings microservice pool for ingest.** Added the
+  `EMBEDDINGS_MICROSERVICE_URL_BULK` setting (`config/settings/base.py`, next to
+  `EMBEDDINGS_MICROSERVICE_URL`), defaulting to `EMBEDDINGS_MICROSERVICE_URL` so
+  single-pool deployments are unaffected. The Celery ingest tasks in
+  `opencontractserver/tasks/embeddings_task.py` now thread this bulk URL into
+  the embedder's existing call-time `embeddings_microservice_url` override kwarg
+  via a new optional `service_url_override` parameter on `_create_text_embedding`,
+  `_create_embedding_for_annotation`, `_batch_embed_text_annotations`, and
+  `_apply_dual_embedding_strategy` (plus `_embed_relationship`). Query/search
+  call sites are unchanged — they keep reading `EMBEDDINGS_MICROSERVICE_URL` (the
+  always-warm pod), so search latency is fully isolated from batch ingest load.
+  No change to the embedder client, base class, or any search resolver.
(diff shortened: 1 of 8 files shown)


JSv4:
This introduces separate configuration pathway besides our singletons... take another pass

PR #2113, opened 2026-07-05
Add Warp-Ingest PDF parser (REST microservice)

JSv4:
## Summary

Adds **`WarpIngestParser`**, an alternative PDF parser based on [Warp-Ingest](https://github.com/Open-Source-Legal/Warp-Ingest) — a *deterministic, rule-based* PDF parser (pdfplumber + optional RapidOCR, no GPU) that renders **straight to the OpenContracts export format** (PAWLS tokens, per-block structural annotations, heading hierarchy as `parent_id` links + relationships).

It follows the existing **Docling REST-microservice pattern**: heavy parsing deps stay in an isolated container (the official `ghcr.io/open-source-legal/warp-ingest` image) and OpenContracts talks to it over HTTP, keeping the Django image slim. Docling remains the default PDF parser; Warp-Ingest is opt-in (selected per corpus/file-type in the admin System Settings UI).

### Design notes
- **Non-chunked** (unlike Docling): Warp-Ingest is CPU-only, its `/api/parse` API takes a whole file, and it joins structure across page boundaries — page-range chunking would fragment that. The whole PDF is sent in one multipart request.
- **Auth via `X-API-Key`** (not `Authorization: Bearer`, which Warp-Ingest also accepts) so `Authorization` stays free for a Cloud Run IAM id_token.
- **No key normalization** — Warp-Ingest's export already uses OpenContracts field names, so the `result` payload flows straight into `save_parsed_data` (verified: its annotation/relationship keys match `import_annotations`/`import_relationships` exactly).
- Error classification mirrors Docling: timeout/connection/5xx → transient, 4xx → permanent `DocumentParsingError`; an `apply_ocr`+`disable_ocr` conflict fails fast client-side.

## Changes
- `opencontractserver/pipeline/parsers/warp_ingest_parser.py` — the parser (auto-discovered by the pipeline registry; no manual registration).
- `config/settings/base.py` + `opencontractserver/constants/document_processing.py` — `WARP_INGEST_PARSER_SERVICE_URL` / `WARP_INGEST_API_KEY` / `WARP_INGEST_PARSER_TIMEOUT` settings backed by a shared timeout constant; documented in the sample `.django` env files.
- `local.yml` / `production.yml` — a `warp-ingest` service (official image) behind an opt-in `warp-ingest` compose profile (~2.5 GB image, so not started by default).
- Docs: `docs/pipelines/warp_ingest_parser.md` (+ pipeline overview + mkdocs nav) and an end-to-end smoke-test script `docs/test_scripts/warp_ingest_parser_smoke_test.md`.
- `changelog.d/warp-ingest-parser.added.md`.

## Testing
- **Unit tests** (mocked HTTP): `opencontractserver/tests/test_doc_parser_warp_ingest.py` — request construction (params, multipart `file`, `X-API-Key`), response extraction, error classification, settings.
- **Live smoke test** against the real `ghcr.io/open-source-legal/warp-ingest` container: the exact parser-shaped request is accepted and returns the expected envelope (`{page_dim, num_pages, result}` → **9 pages / 213 annotations / 64 relationships** for `doc_1_pdf_file.pdf`), and the export's keys match the importer verbatim.
- `black` + `flake8` clean.

---
_Generated by [Claude Code](https://claude.ai/code/session_01R8d98EKmTs2vuYzHVVVpq9)_


JSv4:
Thanks for the thorough review — addressed both pre-merge items in `0f06415`:

1. **File-size guard** — added a configurable `max_file_size_mb` (default 200, env `WARP_INGEST_MAX_FILE_SIZE_MB`) that rejects oversized PDFs with a permanent `DocumentParsingError` before buffering them in memory, mirroring `DocxodusServiceParser`. Documented the per-parse memory footprint in the parser doc.
2. **Multipart filename sanitization** — `document.title` now flows through `_safe_pdf_filename`, which strips CR/LF, control chars, quotes and path separators before it becomes the `Content-Disposition` filename. Verified end-to-end against the live `ghcr.io/open-source-legal/warp-ingest` container that a CRLF-laden title is sanitized and still accepted (200, 9 pages / 213 annotations / 64 relationships).

On the non-blocking items: leaving the no-catch-all style and the `abc123` env fallback as-is for consistency with `DocxodusServiceParser` and the existing `VECTOR_EMBEDDER_API_KEY` convention (the Celery task's top-level `except` already wraps unexpected errors).

Note on the earlier red CI: the root cause was a single `mypy` error in the test (passing a `list` to `_extract_export`), which **gated the `pytest` job** (it was skipped, not failed) — that's also why codecov backend showed only carryforward coverage. Fixed by typing `_extract_export`'s parameter as `Any` (it validates the shape at runtime); the added tests for the size-guard and filename paths further raise patch coverage.

---

JSv4:
Good catches — all three addressed in `5463967`:

1. **Size guard read-then-check** — you're right, the old ordering read the whole file before checking, so the comment/doc "before buffering" claim was false. Now checks `default_storage.size(document.pdf_file.name)` **before** opening/reading, so an over-limit PDF is never buffered. The test asserts `default_storage.open` is never called on the oversized path.
2. **Doc over-claim on env vars** — corrected. Confirmed `migrate_pipeline_settings` resolves each `env_var` via `getattr(django_settings, env_var, None)` (lines 252-253), so only the 3 fields with matching `base.py` attributes are env-seeded. The config section now splits into (a) the 3 base.py-wired settings and (b) the 6 component defaults that are metadata-only and changed via the admin System Settings UI (with an explicit note that setting those 6 in `.env` does nothing through the command).
3. **`response.json()` outside the try** — now wrapped: a truncated/non-JSON 200 body raises a classified transient `DocumentParsingError` (with a test) instead of a raw `JSONDecodeError`.

Also took the regex nit (`\r\n` is redundant with `\x00-\x1f`). Left the classification-block and `MockResponse` duplication as-is since they're the established per-parser pattern (Docling/Docxodus each carry their own copies) — a shared test double / helper feels like a separate cleanup across all three.

---

JSv4:
Thanks — acted on two of these in `5949f60`:

- **Secret in logs (item 3).** Good find, and it turned out to be a genuine (if pre-existing) leak worth fixing here since this PR routes a live `api_key` through it: `BaseParser.parse_document` logged the merged component settings at INFO **without** redaction, while `embedder`/`post_processor`/`enricher` base classes already run `redact_sensitive_kwargs()` on the identical line — `parser.py` was the lone holdout. Fixed it there (masks `api_key` as `***`), with a regression test asserting the secret doesn't reach the log. Changelog fragment added under `security`.
- **Test gaps (item 4).** Added coverage for the `disable_ocr`/`include_images` call-time overrides and the `use_cloud_run_iam_auth` Authorization-header merge path.

On the others: (1) the missing blanket `except` is intentional — as you traced, `ingest_doc` reclassifies any uncaught exception, and unlike Docling there's no chord path here that needs a load-bearing local handler; (2) agreed the pre-read size check is stricter than Docxodus's read-then-check — backporting it to `DocxodusServiceParser` is a reasonable separate follow-up. Live-container smoke test (real fixture PDF end-to-end) passed: 9 pages / 213 annotations / 64 relationships.

---

JSv4:
Addressed the two actionable points in `84c6934`:

1. **Compose `depends_on` race** — added `warp-ingest` to `django`/`celeryworker` `depends_on` (`condition: service_started`, `required: false`) in both `local.yml` and `production.yml` (via the `&django` anchor, so celeryworker/celerybeat inherit it). I verified with `docker compose config` that `required: false` makes the *profiled* dependency optional — config validates cleanly with the `warp-ingest` profile **off** (no error, service absent) and **on** (all three of django/celeryworker/celerybeat depend on it). So Django now waits for the service when the profile is active, without breaking the default stack.
2. **`process_document` kwargs log** — ran it through `redact_sensitive_kwargs` too, for consistency with the `parse_document` line (same file). Agreed it's unlikely to carry secrets today, but cheap defense-in-depth.

On (3): agreed the shape check is intentionally a cheap "obviously-wrong-shape" guard, not full nested validation — deeper structural issues are `save_parsed_data`'s domain, so I'll leave it as the light sanity check.

FYI the earlier codecov 34%→now-91% jump was the mypy-gated `pytest` job finally running; the linter failure on the prior push was a `[method-assign]` in the new redaction test (assigning to a bound method), now fixed via `patch.object`.

---

JSv4:
Thanks — took the two cheap ones in `88b5989`:

1. **`_extract_export` docstring** — softened it to describe the check honestly: it's a lightweight "resembles an export" guard (≥1 recognized key), not full field validation. You're right that `save_parsed_data` `.get()`s every field with defaults, so it tolerates a partial export — the guard only separates "an export" from "an error envelope with just `page_dim`/`num_pages`". Kept the lenient behavior (deliberately resilient to future API drift) and just made the comment match.
2. **Timeout assertion** — `test_request_is_built_correctly` now asserts `requests.post` receives `timeout=self.request_timeout`.

On the rest: (3) TOCTOU — agreed, immutable uploaded PDFs make it theoretical, leaving as-is; (4) no `warp-ingest` healthcheck — intentional, matches the `docling-parser`/`gotenberg` `condition: service_started` convention. No behavior change in this commit, so it's docs+test only.

---

JSv4:
Both actionable items done in `0e06a04`:

1. **API-key doc/behavior mismatch** — good catch. Warp-Ingest always requires a key (its own default is `abc123`), so the "leave blank if unauthenticated" wording was wrong. Reconciled the `api_key` field description and the sample `.django` comments to state that a blank value yields a 401 and must match the service's `WARP_API_KEY`.
2. **No-normalization contract now covered in CI** — added `TestWarpIngestSaveParsedDataIntegration`, which runs a Warp-Ingest-shaped export (camelCase `annotationLabel`/`rawText`, page-keyed `annotation_json`, `parent_id` link, `relationshipLabel` + `*_annotation_ids`) through `save_parsed_data` → `import_annotations`/`import_relationships` and asserts the `Annotation`/`Relationship` rows and the parent link are created. So a future upstream schema drift is caught in CI, not just by the manual smoke test.

On (3): agreed `_safe_pdf_filename` isn't a general-purpose sanitizer — it deliberately targets only the `Content-Disposition`-relevant chars (CR/LF/quotes/path separators). Leaving Windows-reserved chars alone so it doesn't over-mangle legitimate titles.

---

JSv4:
Thanks for the pass. Evaluated all three and I'm intentionally **not** pushing a commit for them — each is non-blocking and a change would be scope creep or speculative:

1. **`Document.DoesNotExist` unclassified** — as you noted, this matches the majority pattern (`docxodus`, `oc_markdown`, `oc_text`, and `BaseParser.save_parsed_data` itself), and `ingest_doc`'s top-level `except` reclassifies it as a retryable error anyway. Making it consistent is a codebase-wide follow-up, not this PR.
2. **`WARP_API_KEY` default `abc123`** — opt-in, documented, and consistent with the existing `VECTOR_EMBEDDER_API_KEY` convention (flagged as consistent in an earlier round). A startup assertion would be a reasonable enhancement once/if the profile sees real production use, but it's orthogonal to adding the parser.
3. **`_extract_export` fallback vs an error-shaped 200 body** — genuinely speculative: Warp-Ingest returns error bodies with 4xx/5xx status (caught by `raise_for_status` before `_extract_export`), and success is `{page_dim, num_pages, result}`. A 200 body shaped like `{"error": ..., "labelled_text": []}` isn't in the current contract, so hardening the presence check against it would be guarding a shape the API doesn't produce. The fallback exists purely as defense against a *future* unwrapping change; I'd rather not over-fit it to a hypothetical error shape.

Net: leaving these as noted follow-ups. The substantive review items across the earlier rounds (pre-read size guard, secret redaction, compose `depends_on`, malformed-JSON classification, and the `save_parsed_data` integration test) are all landed and tested.

---

JSv4:
Pinned the image (point 1): `warp-ingest` now uses `ghcr.io/open-source-legal/warp-ingest:1.0.0` in `local.yml`/`production.yml` and the docs, instead of `:latest`. Agreed the deterministic framing makes a floating tag a real risk of silent output drift; `1.0.0` is the exact image the live smoke test ran against (its digest matches what `:latest` currently points to), and this mirrors `docxodus-parser`'s exact-version pin.

Point 2 (component-settings vs call-time precedence) is pre-existing base-class behavior — and now that `main` (with #2115) has been merged into this branch, that precedence is explicitly the documented, intentional behavior (component settings win, with the falsy-guard), so nothing to change here.

Also FYI: I rebased this branch onto the latest `main` merge — #2115 independently added the same `redact_sensitive_kwargs` call to `parse_document` that this PR did, and the two converged cleanly (my `process_document` redaction is the remaining unique bit). Verified the merged tree compiles, lints, and keeps all the warp-ingest wiring intact.

---

JSv4:
Thanks — evaluated all four; not pushing a change, with reasoning:

1. **Stream the file instead of `f.read()`** — the premise doesn't quite hold for plain `requests`: passing a file object to `files=` still builds the full multipart body in memory (urllib3's `encode_multipart_formdata` reads it whole); true streaming needs `requests-toolbelt`'s `MultipartEncoder`, i.e. a new dependency. So swapping `f.read()` for `f` wouldn't meaningfully lower peak memory here. Given the `default_storage.size()` pre-read cap already bounds this and it mirrors `DocxodusServiceParser`, I'd rather not add a dep for a marginal win — noting it as a possible future pass if huge PDFs become common.
4. **`headers: dict[str, str | bytes] = {}` annotation is redundant** — it's actually load-bearing: `maybe_add_cloud_run_auth` is typed `dict[str, str | bytes]`, and without the annotation the empty literal infers `dict[str, str]`, which mypy rejects at that call (invariant dict value type). It also isn't "immediately reassigned" — the `X-API-Key` is inserted into it first. It mirrors `docling_parser_rest.py` exactly, so I'll leave it.

2. **Field `description=` strings don't repeat the "admin-UI-only, env var not wired" caveat** — fair point on code/doc drift. Leaving it for now since those descriptions render in the admin System Settings GUI (where you're already editing the value, so the env-var confusion doesn't arise) and the doc table covers the distinction explicitly — but agreed it's a reasonable optional polish.
3. **Cloud-Run test mocks the helper** — as you noted, that's the intended unit-level scope (`maybe_add_cloud_run_auth` has its own coverage); no change.

All non-blocking, and I'm deliberately not re-pushing for cosmetics to avoid churning the branch/CI. The substantive items across the earlier rounds are all landed.

---

PR #2133, opened 2026-07-07
Authority packs: baseline-origin collision guard + multi-YAML taxonomy merge (#2057)

### LINKED ISSUE #2057: First-class authority packs: data-driven source-host allowlist + out-of-tree provider discovery
**Pack-enabler infra for the Authority Packs proposal** (PR #2053). Makes the authority-pack format **fully drop-in (zero core edits)** for any jurisdiction.

## Problem

Two things currently force an authority pack to touch core's source tree (documented as the pack's residual gaps):

1. **Host allowlist is a hardcoded frozenset.** `PUBLIC_DOMAIN_SOURCE_HOSTS` in `opencontractserver/constants/safe_http.py` is the SSRF allowlist every fetch is gated against. A pack for a new jurisdiction cannot open its government host(s) as *data* — it needs a same-PR constants edit. This is the single binding the pack format cannot self-declare (an unlisted host → `SSRFValidationError` → `GATE_BLOCKED_DOMAIN`).
2. **Provider discovery scans one hardcoded package.** `PipelineComponentRegistry` auto-discovers providers only under `opencontractserver/pipeline/authority_source_providers/`. A pack's provider module must be physically copied into core; there is no entry-point / out-of-tree plugin path, so a pack cannot live self-contained in its own package.

## Proposed direction

- **(a) Data-driven source-host allowlist** — a superuser-curated, license-gated DB allowlist that *augments* the static frozenset (still SSRF-validated; still requires `license="public-domain"` at the gate). A pack/provider declares its hosts as data.
- **(b) Out-of-tree provider discovery** — entry-point (or configured-extra-package) discovery so a provider outside the core package registers, letting a pack ship its provider in its own folder.
- **(c, minor)** Loader multi-YAML merge + a baseline-vs-baseline collision guard (`update_or_create` currently has no source guard between two baseline writers on the same prefix — last loader run wins).

## Acceptance criteria

- A provider/pack can declare its fetch host(s) as data; SSRF + license gating preserved; superuser-curated.
- A provider outside `pipeline/authority_source_providers/` is discovered and routable.
- Re-loading two packs that touch distinct prefixes never clobbers each other.

## Context

Closes the one **blocker** and the out-of-tree gap from `docs/architecture/proposals/0002-authority-packs.md` §7. Independent of Phases 2–4; with this, a pack becomes pure data + a self-contained provider package with no edits to core.

JSv4:
Closes #2057.

## Scope note (what was already done)

Issue #2057 lists three items. Two of them **already shipped** in PR #2064 (merged the day after the issue was filed), with a deliberately different design than the issue text sketches:

- **(a) Host allowlist as data** — shipped as `pack.yaml` `source_hosts:` unioned into the SSRF allowlist at runtime (`enrichment/services/authority_source_hosts.py`; install-is-trust, not a DB table).
- **(b) Out-of-tree provider discovery** — shipped as `<pack>/providers/*.py` scanning + the `AUTHORITY_PACK_PATHS` setting (the issue's "configured-extra-package" alternative).

The proposal doc already declares gaps 1 & 6 closed. This PR therefore implements the **remaining item (c)** — loader multi-YAML merge + the baseline-vs-baseline collision guard (gap 7 in proposal 0002 §7) — which is also acceptance criterion 3 ("re-loading two packs that touch distinct prefixes never clobbers each other").

> **Flagging one open design question:** if you still want the literal part (a) — a superuser-curated, runtime-editable **DB** host allowlist on top of the pack.yaml mechanism (value: no filesystem access / no restart needed) — say the word and I'll add it as a follow-up. I kept it out here since it would be a second, overlapping allowlist-widening mechanism.

### Baseline-origin collision guard (fixes silent last-write-wins)

Two `source="baseline"` writers on the same prefix — core `authority_mappings.yaml` vs. a pack, or two packs — previously clobbered each other silently (`AuthorityMappingLoader.load_namespaces`' `update_or_create` had no writer partition; the last loader run won).

- New nullable `AuthorityNamespace.baseline_origin` column (migration `annotations/0101`): `"core"` for the shipped YAML / `post_migrate` seed, else the pack's manifest `name`.
- `load_namespaces` (`enrichment/services/authority_mapping_loader.py`) now **skips + warns** (counted as `skipped_foreign_baseline`) when a different origin owns the prefix — first writer wins; curator `manual` rows and corpus-linked namespaces still trump everything, unchanged.
- The `post_migrate` convergence (`enrichment/_namespace_seed.py`) honours the same guard (field-presence-guarded for historical migration states), so `migrate`/flush can no longer revert a pack-owned prefix to the constants baseline — and it stamps its own rows `"core"`.
- Legacy null-origin baseline rows are **adopted** (stamped) by the next owning load, so fleets converge without a data migration.
- The pack name `core` is **reserved**; `load_authority_pack` rejects it fail-fast before any DB write.

### Multi-YAML merge (one command converges everything)

- `AuthorityMappingLoader.load_installed()` loads the core baseline **plus every installed pack's mappings YAML** (in-tree `authority_packs/` + `AUTHORITY_PACK_PATHS`), each stamped with its origin. Core loads first, so on a same-prefix collision the shipped baseline wins — mirroring `authority_pack_config`'s "baseline always wins" merge rule for shape rules/abbreviations. Per-pack fault isolation: one malformed pack YAML is logged + reported as an error, the rest still load.
- `manage.py load_authority_mappings --include-packs` exposes it; both commands now print per-origin summaries including `skipped_foreign_baseline`.
- DRY: pack discovery reuses `authority_pack_config.iter_pack_mapping_files` (made public; now yields the parsed manifest so callers don't re-read `pack.yaml`).

### Docs & changelog

- `docs/guides/authoring-authority-packs.md`: new "Prefix ownership" section (what a re-load can/cannot clobber, how to resolve a collision, the one-command converge).
- `docs/architecture/proposals/0002-authority-packs.md`: gap-7 update callout; §7 table rows 1/6/7 marked shipped.
- Changelog fragments: `changelog.d/2057-baseline-origin-guard.fixed.md`, `changelog.d/2057-loader-multi-yaml.added.md`.

## Tests

- `test_authority_mapping_loader.py` — `BaselineOriginGuardTests` (distinct-prefix packs never clobber; foreign origin skipped + warned; unattributed run can't steal; legacy adoption; manual still wins; core stamp), `LoadInstalledTests` (core+bolivia merge, idempotency, malformed-pack isolation), reseed-ownership additions, `--include-packs` command coverage.
- `test_authority_pack.py` — pack-origin stamping via `load_authority_pack`; reserved `core` name rejected.
- `test_authority_pack_config.py` — updated for the public 3-tuple `iter_pack_mapping_files` (mechanical rename; logic unchanged) + manifest-passthrough assertion.

### DIFF
--- a/changelog.d/2057-baseline-origin-guard.fixed.md
+++ b/changelog.d/2057-baseline-origin-guard.fixed.md
@@ -1,0 +1,19 @@
+- **Baseline-vs-baseline namespace clobbering fixed (issue #2057).** Two
+  `source="baseline"` writers on the same authority prefix — the core
+  `authority_mappings.yaml` vs. a pack's mappings YAML, or two packs — previously
+  last-write-wins'd each other silently
+  (`AuthorityMappingLoader.load_namespaces`' `update_or_create` had no writer
+  partition). Every baseline `AuthorityNamespace` row is now stamped with its
+  writer origin (new nullable `baseline_origin` column, migration
+  `annotations/0101`; `"core"` for the shipped YAML / `post_migrate` seed, else
+  the pack's manifest `name`), and a load skips + warns (counted as
+  `skipped_foreign_baseline`) instead of overwriting a prefix a different origin
+  owns — first writer wins; curator `manual` rows still trump everything. The
+  `post_migrate` convergence (`opencontractserver/enrichment/_namespace_seed.py`)
+  honours the same guard, so a `migrate`/flush can no longer revert a pack-owned
+  prefix to the constants baseline. Legacy baseline rows (null origin) are
+  adopted — stamped — by the next owning load. The pack name `core` is reserved
+  (`load_authority_pack` rejects it fail-fast). Re-loading two packs that touch
+  distinct prefixes can never clobber each other. Tests:
+  `opencontractserver/tests/test_authority_mapping_loader.py::BaselineOriginGuardTests`
+  and the reseed-ownership additions in the same file.
--- a/changelog.d/2057-loader-multi-yaml.added.md
+++ b/changelog.d/2057-loader-multi-yaml.added.md
@@ -1,0 +1,15 @@
+- **One-command multi-YAML taxonomy merge (issue #2057).**
+  `manage.py load_authority_mappings --include-packs` now converges the shipped
+  core baseline **plus every installed authority pack's mappings YAML** (in-tree
+  `authority_packs/` + `AUTHORITY_PACK_PATHS`) in one idempotent run, each file
+  stamped with its writer origin (`AuthorityMappingLoader.load_installed`,
+  reusing `authority_pack_config.iter_pack_mapping_files` — now public and
+  yielding the parsed manifest — for pack discovery). Core loads first, so on a
+  same-prefix collision the shipped baseline wins and the pack's claim is
+  skipped + warned, mirroring the shape-rules/abbreviations merge rule.
+  Per-origin summaries (including `skipped_foreign_baseline`) are printed for
+  both commands (`load_authority_mappings`, `load_authority_pack`). Docs:
+  "Prefix ownership" section in `docs/guides/authoring-authority-packs.md`;
+  gap 7 closed in `docs/architecture/proposals/0002-authority-packs.md` §7.
+  Tests: `test_authority_mapping_loader.py::LoadInstalledTests` +
+  command tests.
--- a/config/graphql/annotation_types.py
+++ b/config/graphql/annotation_types.py
@@ -457,6 +457,11 @@ class AuthorityNamespaceNode(DjangoObjectType):
             "source_root_url",
             "license",
             "is_global",
+            # Which baseline writer owns the row ("core" or a pack name; null on
+            # manual/corpus-linked and pre-#2057 rows) — lets the console show
+            # who owns a prefix when a load reports a skipped foreign-baseline
+            # collision. Plain CharField (no choices) → auto-resolves as String.
+            "baseline_origin",
             "authority_corpus",
             "created",
             "modified",
--- a/docs/architecture/proposals/0002-authority-packs.md
+++ b/docs/architecture/proposals/0002-authority-packs.md
@@ -83,6 +83,22 @@ rather than against a `scraping/` app that was never built.
 > `test_discover_authority_candidates_command.py`, plus the pack-discovery
 > additions in `test_authority_pack_providers.py`.
 
+> **Update — gap 7 closed (issue #2057).** The mappings loader now guards
+> baseline-vs-baseline collisions and merge-loads multiple YAMLs. Every
+> `source="baseline"` `AuthorityNamespace` row is stamped with its writer origin
+> (`baseline_origin`: `"core"` for the shipped YAML, else the pack's manifest
+> `name`), and a load **skips + warns instead of clobbering** a prefix a
+> different origin owns — first writer wins; curator `manual` rows still trump
+> everything (`AuthorityMappingLoader.load_namespaces`; the `post_migrate`
+> convergence in `enrichment/_namespace_seed.py` honours the same guard). So
+> re-loading two packs that touch distinct prefixes can never clobber each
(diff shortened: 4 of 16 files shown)


JSv4:
Addressed both review rounds in 176a72b:

- **`yaml.YAMLError` fault-isolation bug (both reviews)** — confirmed real: `load_mappings_file` calls `yaml.safe_load` unguarded and `yaml.YAMLError` is not a `ValueError`. `load_installed` now catches `(ValueError, OSError, yaml.YAMLError)`, with a new test using genuinely unparsable YAML (`test_load_installed_isolates_an_unparsable_pack`).
- **Broken-manifest packs invisible in the report (review 2)** — `iter_pack_mapping_files` grew an optional `errors` sink; `load_installed` now reports a pack whose `pack.yaml` won't parse as an `{"error": …}` entry keyed by directory name (`test_load_installed_reports_an_unparsable_manifest`). Runtime vocab scans keep their log-and-skip contract.
- **`baseline_origin` not visible to curators (review 1)** — added to `AuthorityNamespaceNode.Meta.fields` (plain CharField, auto-resolves as String). A console facet/column is left as frontend follow-up.
- **Race window in `load_namespaces` (review 1)** — accepted, now documented in-code: namespace loads are operator-run commands/migrations, not concurrent runtime writers (unlike equivalences, whose `upsert_equivalence` locks because ingest tasks call it concurrently). Will revisit if namespace loading ever moves into task fan-out.
- **Duplicate pack names sharing one origin (review 1 nit)** — intentional: two dirs declaring the same manifest `name` are by declaration the same pack (e.g. in-tree + `AUTHORITY_PACK_PATHS` copy); they co-own their prefixes, the load stays idempotent, and the duplicate is warned. Unique names are the authoring contract (documented in the guide).
- **Defensive delete in `test_reserved_core_pack_name_rejected` (review 1 nit)** — kept but re-commented: it baselines the *nothing-was-written* assertion against possible cross-module leakage on reused `--keepdb` databases, not because `TestCase` isolation is suspect.

---

JSv4:
Round-3 review addressed in 4f2e511:

- **`load_all` non-atomicity** — confirmed and fixed: both steps now run inside `transaction.atomic()`, so a YAML with valid `prefixes:` but malformed `equivalences:` rolls back entirely — an "errored" pack in the `load_installed` report now genuinely means nothing took effect. Regression test `test_load_all_error_leaves_no_partial_namespace_rows`. This also restores `load_authority_pack`'s "validate the whole pack before any DB write" contract for the taxonomy step.
- **Duplicate pack names** — documented in the guide's Prefix ownership section: `name:` must be unique across all installed pack dirs; same name = same pack, no guard between the directories.
- **Origin-derivation duplication** — extracted `authority_pack_config.pack_origin_name(pack_dir, manifest)`; both call sites use it.
- **Case-sensitive `core` check** — now case-insensitive in both the command guard and `load_installed`'s reserved-name skip. (A pack named `Core` never actually bypassed the guard — its origin string differs from `"core"` — but it would read as the reserved name to a human, so it's rejected too.)

---

JSv4:
Round-5 observations handled in b1ba0bf:

- **#3 (over-length pack name)** — this was actually sharper than flagged: the resulting `DataError` is neither a `ValueError` nor a `yaml.YAMLError`, so one such pack would have aborted the whole `--include-packs` converge. Fixed twice over: `load_authority_pack` now fail-fasts a name longer than the column (new `BASELINE_ORIGIN_MAX_LENGTH` constant, shared with the model field), and `load_installed`'s per-pack catch is broadened to `Exception` — any per-pack failure now rolls back atomically and lands in the report as `{"error": "<ExcClass>: …"}` instead of killing the run. Tests: `test_over_length_pack_name_rejected`, `test_load_installed_isolates_a_db_level_failure`.
- **#1 (per-prefix N+1)** and **#2 (no index on `baseline_origin`)** — agreed both are follow-ups, not this PR: the per-prefix loop is pre-existing and file sizes are tens of prefixes; the index belongs with whatever console feature first filters on the column (adding it now is a speculative migration).

---

PR #2137, opened 2026-07-07
One-step desktop setup: python oc-desktop.py self-bootstraps everything

JSv4:
## Goal

Make the single-user desktop packaging usable by a **minimally technical end user**: from a source download to the app in a browser in as close to one step as possible, cross-platform (Windows/macOS/Linux).

Validated by two clean-room usability tests with an agent role-playing a non-technical user: one against the baseline flow (to find the real cliffs), one against the new flow (to confirm they're gone). Results are posted as a PR comment.

## What the baseline test found

The old flow (`yarn build` + `pip install` + `OC_DESKTOP_PASSWORD=... python oc-desktop.py`) failed a non-technical user four ways:

1. **BLOCKER — every fresh desktop install died mid-migration**: the embedded `pgserver` Postgres bundles no contrib extensions, so `annotations/0074`'s unconditional `TrigramExtension()` raised `NotSupportedError: extension "pg_trgm" is not available`. Escaping it required compiling the extension from PostgreSQL source.
2. **MAJOR** — step 1 (`yarn build`) fails as written (`yarn install` never mentioned) and assumes a Node toolchain.
3. **MAJOR** — desktop docs unreachable: not linked from README, quick start, or the mkdocs nav; login username `desktop` documented nowhere (README's `admin` fails).
4. **MINOR** — no venv guidance, random port each launch, no stop instructions.

## What changes

**One command now does everything: `python3 oc-desktop.py`** (Windows: `py oc-desktop.py`).

- **`opencontractserver/desktop/bootstrap.py`** (new, stdlib-only): Python version gate 3.10–3.12 (`pgserver` has no cp313 wheels) with a plain-English error; asks for the login password **up front** (the one interactive question, then setup runs unattended); creates a private venv under the per-user app-data dir; installs `requirements/desktop.txt` (auto-reinstalls when the requirement files change, via content fingerprint) under a cross-process install lock; re-execs the launcher inside the venv. Loop-guarded, with actionable failure messages (missing python3-venv, disk full/read-only, no network).
- **`opencontractserver/desktop/spa_dist.py`** (new): removes the Node requirement. SPA resolution: repo `frontend/dist` → version-stamped cached download in app-data → **`opencontracts-frontend-dist.zip` from the GitHub release** (version-matched tag, then latest; SHA-256-verified, size-capped, staged-then-swapped so a failed refresh never destroys a working cache) → `yarn install && yarn build` fallback for developers. Zip-slip and symlink-member guarded extraction; honors `SSL_CERT_FILE`, falls back to certifi.
- **`.github/workflows/docker-build-release.yml`**: new `build-frontend-dist` job attaches the built SPA + its `.sha256` to every release (bundle is environment-agnostic — the launcher rewrites `env-config.js` at startup).
- **`annotations/0074` migration**: probes `pg_available_extensions` via `RunPython`; skips extension + GIN index when `pg_trgm` is absent (the queries it accelerates are plain `icontains` — correct unindexed, fine at single-user scale). Real Postgres deployments get the identical objects. Reverse now drops only the index, removing the documented `DROP EXTENSION` reversal hazard. **Note on the in-place edit (considered deliberately):** Django tracks applied migrations by `(app, name)`, never content, so no already-migrated database re-runs it; 0074 only landed on main days ago via #2133, and the new body is behavior-identical wherever `pg_trgm` exists — a superseding migration would not have fixed fresh desktop installs, which crash *inside* 0074 itself.
- **`desktop_bootstrap`**: password from `OC_DESKTOP_PASSWORD` *or the interactive first-run prompt* (min-length floor enforced on **both** paths, confirmed twice, EOF/Ctrl+C-safe); a password-less account left over from a headless run self-heals on the next launch; never re-prompts once a usable password exists. The password is scrubbed from every long-lived process's environment (wrapper, launcher, children) after the single hand-off.
- **Launcher**: stable default port 8406 (ephemeral fallback); startup banner with URL, the `desktop` username, and Ctrl+C stop instructions; crash-aware health check (no success banner over a dead Daphne — points at the per-child logs, which rotate at 10 MB); clean-teardown farewell stating where data lives; user-private (0o700) app-data tree enforced even for pre-existing dirs.
- **Discoverability**: README "Desktop — run it on your computer (no Docker)" Quick Start section with the app URL; `docs/quick_start.md` cross-link; `docs/deployment/desktop_packaging.md` rewritten around the one-command flow; page added to the mkdocs nav.

## Merge with main

`origin/main` merged in. The one conflict was add/add on `warp_ingest_parser.py`: main's version is the REST client to the Warp-Ingest microservice; this branch's is the in-process library parser the desktop build needs. Resolution keeps main's REST client at the original path and moves the in-process variant to **`warp_ingest_local_parser.py` / `WarpIngestLocalParser`** ("Warp-Ingest Parser (Local)", `WARP_INGEST_LOCAL_*` env vars, same `max_file_size_mb` guard and mutual-exclusion error convention as the REST sibling) so both components coexist unambiguously.

## Tests

- `opencontractserver/tests/test_desktop_packaging.py` (64 tests): path resolution per-OS, private-dir permission enforcement (incl. pre-existing permissive trees), SPA resolution order/zip-slip/symlink/checksum edge cases, trigram migration guard against a real Postgres, password prompt/self-heal/no-reprompt/floor/cross-process-scrub flows, launcher teardown and keyring-fallback behavior.
- Full desktop + local-parser test files pass against a real embedded Postgres with the complete migration run.
- Live end-to-end (three runs this branch): fresh app-data → password prompt within seconds → unattended venv+deps+Postgres+migrate (0074 skips cleanly without pg_trgm) → healthy app on :8406 → GraphQL login with the prompted password → clean Ctrl+C teardown → ~20 s warm relaunch.

https://claude.ai/code/session_01E4ahQjPm1McibG65spkxLR


JSv4:
Consolidated response to the review rounds above — adopted items are in commits `4e0aae1` and `9452e06`:

**Adopted**
- **Stale SPA cache**: downloads are now stamped with the checkout version (`spa/.version`); `ensure_spa` refreshes on mismatch and only falls back to the stale copy (with a warning) when the refresh fails offline. Covered by `test_ensure_spa_uses_version_matched_cache` / `test_ensure_spa_refreshes_stale_cache` / `test_ensure_spa_falls_back_to_stale_cache_offline`.
- **Bundle integrity**: the release job now publishes `opencontracts-frontend-dist.zip.sha256` and `download_spa` verifies it before extracting, refusing a mismatch (older releases without the sibling asset fall back to TLS-only trust). Covered by `test_verify_checksum`.
- **venv-creation failure message**: now distinct from pip failure, with the Debian/Ubuntu `apt install python3-venv` hint.
- Third-Ctrl+C force-kill escape hatch in the bootstrap wait loop; `0o700` on created app-data dirs; OCR-flags-conflict warning + transient-error cross-reference comment in `WarpIngestLocalParser`; `env-config.js` side-effect comment; corrected `MIN_PASSWORD_LENGTH` comment; rate-limit hint in the bundle-not-found message; double-launch warning in the docs; state-vs-DB index invariant note in the 0074 docstring.

**Acknowledged, no change**
- **Editing 0074 in place**: intentional. 0074 first landed on `main` via #2133 immediately before this PR, so the population of databases that ran the old body is tiny and — because the new body produces identical objects wherever `pg_trgm` exists — behaviorally indistinguishable from having run the new one. A superseding migration can't fix the actual failure (fresh desktop installs crash *inside* 0074 before any later migration runs).
- **`sqlalchemy_result_backend_url` and driver-qualified schemes**: `db+postgresql+psycopg2://…` is a valid Celery SQLAlchemy backend URL (the `db+` prefix is stripped and the remainder is a legal SQLAlchemy URL), and a driver-qualified scheme wouldn't get through django-environ's `env.db()` as `DATABASE_URL` in the first place — so the passthrough branch is correct as written.
- **`from .base import *` in `config/settings/desktop.py`** (code-quality bot): this is the settings-inheritance idiom every other profile in the repo uses (`local.py`, `production.py`, `test.py`); diverging in one profile would be worse than the wildcard.
- **`safe_join` private API**: mirrors `django.views.static.serve`'s own usage; if Django relocates it the SPA tests fail loudly.
- Launcher unit-testing and a headless CI smoke test (`/api/health/` + SIGTERM) are good Phase-1 follow-ups; the launcher's supervision loop is currently exercised by the live end-to-end usability runs documented in the PR.

---

JSv4:
## Usability test results (before → after)

Both tests were clean-room runs by an agent role-playing a **minimally technical end user** (knows how to open a terminal and paste commands; does not know git/pip/venv/Node/env vars), starting from nothing but the README on this machine.

### Baseline (old flow) — FAILED
- Desktop instructions unreachable from the README/docs nav (found only by filename search); written as an internal architecture plan.
- Step 1 (`yarn build`) fails instantly as documented (`yarn install` never mentioned; Node assumed).
- `pip install` into system Python, no venv guidance.
- **Hard blocker**: first launch died mid-migration — `pg_trgm is not available` on the embedded Postgres. The only escape was compiling the extension from PostgreSQL 16.2 source.
- Login username (`desktop`) documented nowhere; README's `admin` fails.
- Verdict: *"No — she'd never find the instructions, would trip on the missing yarn install, and would then hit a hard, un-googleable migration failure."*

### Re-test (this PR's flow) — FULL SUCCESS
- Discovery: first Quick Start subsection in the README, found in under a minute; steps genuinely copy-paste self-contained.
- **1 command + 1 password choice + Ctrl+C — the entire journey.** `python3 oc-desktop.py` → private venv + deps (~3 min) → embedded Postgres → full migrate (**0074 skipped cleanly without pg_trgm**) → password prompt → UI acquisition → app at `http://127.0.0.1:8406/` with a banner stating the URL, the `desktop` username, and how to stop.
- Verified end-to-end: `/` 200, `/api/health/` ok, GraphQL `tokenAuth` as `desktop` with the chosen password, authenticated `me` query.
- Clean Ctrl+C teardown (~10 s, zero orphan processes, ports closed). **Second launch: ~20 s to responding, no re-prompt, no reinstall — exactly as the README promises.**
- Verdict: *"Yes — a non-engineer colleague could do this unaided … compared to a README demanding yarn/pip/env-var setup, this is a different universe."*

### Fixes applied from the re-test's remaining friction (commits `7dfc22f`, `38163aa`)
1. Password prompt moved to the **start** of the first run (it landed ~15 min in, mid-install, after the user walked away) — setup is now fully unattended after one question.
2. The scary-but-harmless `ERROR: permission_corpus Permission does not exist!` that scrolls past every fresh install (`corpuses/0028`) is now silent on fresh DBs and a real WARNING only when existing corpuses were actually skipped.
3. Silent phases got progress lines (nltk download), shutdown got a farewell message (where data lives, how to restart), the keyring warning got plain-language wording, and the README states the app URL and honest first-run timing.
4. The "no Node" promise currently depends on the next release publishing `opencontracts-frontend-dist.zip` (the workflow in this PR does that); on this pre-release checkout the launcher fell back to the yarn build with a clear message.

---

JSv4:
Re: the latest review — implemented 2 and 3, acknowledging 1, 4, 5 (commit `888f173`):

1. **Orphaned children on SIGKILL/OOM — yes, known, tracked for Phase 1** (`docs/deployment/desktop_packaging.md` Phase-1 list: "harden embedded-Postgres supervision (stale-lock recovery …)"; process-group/Job-Object teardown belongs there with it). Two mitigations already soften the cliff: `_free_port()` falls back to an ephemeral port if 8406 is squatted by a leftover Daphne, and `pgserver.get_server()` reuses/recovers a running instance for the same pgdata rather than failing on the lock in the common case.
2. **Launcher teardown now has a behavioral test**: `LauncherShutdownTests` spawns a real child via `_spawn`, asserts `_shutdown()` actually stops it and that the durable per-child log exists — a regression net for the riskiest logic despite the coverage omit.
3. **Keyring fallback contract pinned**: `StableSecretKeyFallbackTests` covers hung-keyring (ephemeral within the timeout), broken-backend (ephemeral), and healthy (persisted key) paths, so a refactor can't silently overwrite a good persisted key with the ephemeral one without a test going red.
4. **`deps_ready()` importability-over-version check — intentional**, now documented in its docstring: it's what lets the venv child, dev environments, and future installer payloads run directly; the ambient-interpreter drift case is the accepted trade-off.
5. **`OC_DESKTOP_PASSWORD` in the bootstrap child's `/proc/<pid>/environ`** — accepted for the single-user threat model (same-UID-only visibility, seconds-long window); it's the transport between the early prompt and the DB-owning command, and it's scrubbed from the supervisor and all long-lived children immediately after.

---

JSv4:
Not doing this for now

PR #2138, opened 2026-07-07
Add optional turbopuffer-style object-storage vector search backend

JSv4:
# Overview

Adds an **opt-in, reversible** vector search backend that keeps its index entirely in object storage, following [turbopuffer's architecture](https://turbopuffer.com/architecture): object storage as index source of truth, per-namespace WAL with strongly consistent reads (unindexed tail scanned as an overlay), async compaction into immutable centroid-ANN segment generations, and tiered caching for warm queries. **Default behavior is unchanged** — `VECTOR_SEARCH_BACKEND=pgvector` remains the default, and the object-storage path falls back to pgvector on unindexed namespaces, engine errors, or post-filter shortfalls.

Design doc: `docs/architecture/object_storage_vector_search.md` (includes the build-on-turbopuffer vs from-scratch analysis and prior-art comparison — OpenSearch searchable snapshots, Quickwit — from the referenced HN discussion).

**Flag on feasibility (as requested):** turbopuffer is closed-source SaaS, so there is nothing to self-host; this is a from-scratch implementation of the same architecture (~500 lines of numpy over blob-store primitives). A hosted-turbopuffer API driver could later slot into the same `search_via_object_index()` / `enqueue_embedding_index_sync()` seam.

## How it works

```
Embedding.objects.store_embedding()  ── on_commit ──▶ Celery WAL append (one PUT)
  (+ worker_uploads bulk writes, hooked explicitly)
<prefix>/<ns>/wal/*.jsonl  ── auto-compaction (cache-lock, k-means, k=√n) ──▶
<prefix>/<ns>/index/segments/<gen>/{centroids.npy, cluster_*.npz} + manifest.json
VectorSearchViaEmbeddingMixin.search_by_embedding()  ◀── router probes nprobe
clusters + WAL tail, then RE-FILTERS candidate ids through the caller's queryset
```

- **Integration seam**: `VectorSearchViaEmbeddingMixin.search_by_embedding` (`opencontractserver/shared/mixins.py`). **Scope: document, annotation, and note search** — the parent kinds whose querysets inherit the mixin — so GraphQL semantic search, Discover, MCP `search_corpus` passages, agent tools, and hybrid RRF over those kinds inherit the toggle with zero caller changes. Conversation/ChatMessage/Relationship vector search deliberately stays on pgvector for now (their reads use separate inline implementations with a different result contract; documented follow-up — see `EMBEDDABLE_PARENT_KINDS`), and their writes are correspondingly not fanned out to the index.
- **Permissions preserved by construction**: the engine stores only `(parent_pk, vector)`; candidate ids are re-filtered through the caller's already-scoped queryset (`visible_to_user`, MIN(document, corpus), structural rules), with capped oversampling and a shortfall rule that falls back to pgvector rather than under-filling. Deleted rows drop out at the same re-filter. The index blobs carry no ACL — system check `opencontracts.W004` warns when the backend is enabled over storage with public-read signals (AWS + GCS).
- **Write path**: fan-out from `Embedding.objects.store_embedding` (the in-app pipeline chokepoint) plus explicit hooks in `worker_uploads/tasks.py`'s bulk writes, via `opencontractserver/tasks/vector_index_tasks.py`, with autoretry and pending-marker-gated auto-compaction (cache-lock serialised, deferred GC by one generation); fire-and-forget — an index sync failure never breaks the Postgres write.
- **Ops**: `manage.py rebuild_object_vector_index` (batched, `--dry-run`, progress, lock-aware) replays Postgres embeddings into the index; system checks `opencontracts.E002` (invalid backend value) and `W003` (process-local cache vs compaction lock); tuning constants in `opencontractserver/constants/search.py`.

## Testing

- **38 tests** (`opencontractserver/tests/test_object_storage_vector_backend.py`): engine-level (WAL-only strong consistency, compaction parity with numpy brute force, tombstones, overwrite ordering, determinism, recall, manifest mid-overwrite retry, WAL/manifest read-ordering race, deferred/partial GC, put_bytes overwrite races) and Django integration (toggle both ways, queryset scoping, all fallback paths incl. filter shortfall and fetch-cap, end-to-end `CoreAnnotationVectorStore` caller, worker-upload direct-write fan-out, rebuild command incl. dry-run + lock contention, auto-compaction gating, read/write kind symmetry, E002/W003/W004 checks).
- **Regression**: existing search/embedding suites (incl. conversation, subtree-group, and worker-upload suites) and the 92 architecture-invariant tests — all pass.
- **Real object-storage verification**: engine exercised end-to-end against **MinIO via the S3 API** (`docs/test_scripts/object_storage_vector_backend_minio.md`) and **fake-gcs-server via the GCS JSON API** through django-storages `GoogleCloudStorage` (`docs/test_scripts/object_storage_vector_backend_gcs.md`) — all three supported storage backends (LOCAL/AWS/GCP) verified.

## Limitations / future work (documented in the design doc)

Vector arm only (FTS/hybrid RRF unchanged in Postgres); document/annotation/note kinds only (conversation/message/relationship read paths are the named follow-up); async write-visibility window (Celery lag vs turbopuffer's synchronous ack); no group commit (~3 object-store roundtrips per online write — bulk jobs should use the rebuild command); lazy deletes; wall-clock WAL ordering under worker clock skew; non-atomic manifest overwrite (retry-mitigated; S3 conditional PUT named as the permanent fix); full recluster per compaction; per-process cache only; deployment-wide (not corpus-sharded) namespaces.

### DIFF
--- a/changelog.d/object-storage-vector-backend.added.md
+++ b/changelog.d/object-storage-vector-backend.added.md
@@ -1,0 +1,17 @@
+- **Optional object-storage vector search backend (turbopuffer-style), disabled by default.**
+  New `VECTOR_SEARCH_BACKEND` setting (`pgvector` default | `object_storage`) routes the
+  single similarity call site (`opencontractserver/shared/mixins.py::VectorSearchViaEmbeddingMixin.search_by_embedding`)
+  through a WAL + centroid-ANN index kept in the default file storage (local disk, S3/MinIO, or GCS
+  under `VECTOR_INDEX_STORAGE_PREFIX`), with Postgres `Embedding` rows remaining the source of truth.
+  Engine (`opencontractserver/vector_search/engine.py`) provides strongly consistent reads
+  (WAL-tail overlay), deterministic k-means compaction (Celery, cache-lock serialised,
+  auto-triggered at `OBJECT_INDEX_COMPACT_MIN_WAL_FILES`), tombstones, and a per-process LRU for
+  warm queries. Write path fans out from `Embedding.objects.store_embedding`
+  (`opencontractserver/shared/Managers.py`) via `opencontractserver/tasks/vector_index_tasks.py`;
+  candidate ids are re-filtered through the caller's queryset so all permission/visibility scoping
+  is preserved; unindexed namespaces and engine errors fall back to pgvector automatically.
+  Includes `manage.py rebuild_object_vector_index`, system check `opencontracts.E002` for invalid
+  backend values, constants in `opencontractserver/constants/search.py`, design doc
+  `docs/architecture/object_storage_vector_search.md`, MinIO manual test script
+  `docs/test_scripts/object_storage_vector_backend_minio.md`, and 18 tests in
+  `opencontractserver/tests/test_object_storage_vector_backend.py`.
--- a/changelog.d/object-storage-vector-backend.added.md
+++ b/changelog.d/object-storage-vector-backend.added.md
@@ -12,6 +12,7 @@
   is preserved; unindexed namespaces and engine errors fall back to pgvector automatically.
   Includes `manage.py rebuild_object_vector_index`, system check `opencontracts.E002` for invalid
   backend values, constants in `opencontractserver/constants/search.py`, design doc
-  `docs/architecture/object_storage_vector_search.md`, MinIO manual test script
-  `docs/test_scripts/object_storage_vector_backend_minio.md`, and 18 tests in
-  `opencontractserver/tests/test_object_storage_vector_backend.py`.
+  `docs/architecture/object_storage_vector_search.md`, manual verification scripts
+  against real S3 (MinIO — `docs/test_scripts/object_storage_vector_backend_minio.md`)
+  and the GCS API (fake-gcs-server — `docs/test_scripts/object_storage_vector_backend_gcs.md`),
+  and the test suite in `opencontractserver/tests/test_object_storage_vector_backend.py`.
--- a/config/settings/base.py
+++ b/config/settings/base.py
@@ -1462,6 +1462,22 @@ DEFAULT_EMBEDDER = "opencontractserver.pipeline.embedders.sent_transformer_micro
 # Default embedding dimension to use if no dimension is specified
 DEFAULT_EMBEDDING_DIMENSION = 768
 
+# Vector search backend
+# ------------------------------------------------------------------------------
+# "pgvector" (default): similarity search runs in Postgres via HNSW indexes.
+# "object_storage": similarity ranking is served from a turbopuffer-style
+# WAL + centroid index kept in the default file storage (S3/GCS/local disk),
+# with Postgres remaining the source of truth (automatic fallback to pgvector
+# for unindexed namespaces or engine errors). See
+# docs/architecture/object_storage_vector_search.md.
+VECTOR_SEARCH_BACKEND = env.str("VECTOR_SEARCH_BACKEND", default="pgvector")
+
+# Key prefix under which the object-storage vector index lives inside the
+# default storage backend.
+VECTOR_INDEX_STORAGE_PREFIX = env.str(
+    "VECTOR_INDEX_STORAGE_PREFIX", default="vector-index"
+)
+
 
 # Default runner
 TEST_RUNNER = "opencontractserver.tests.runner.TerminateConnectionsTestRunner"
--- a/docs/architecture/embeddings_creation_and_retrieval.md
+++ b/docs/architecture/embeddings_creation_and_retrieval.md
@@ -208,6 +208,9 @@ When working in async contexts (such as Django Channels WebSocket consumers), al
    - Core store builds Django queryset with filters (corpus, document, user, metadata)
    - Applies structural annotation logic (structural annotations bypass corpus filtering)
    - Uses `.search_by_embedding(...)` from `VectorSearchViaEmbeddingMixin` for similarity search
+     (backend-pluggable: pgvector HNSW by default, or the object-storage index
+     when `VECTOR_SEARCH_BACKEND=object_storage` — see
+     `docs/architecture/object_storage_vector_search.md`)
    - Returns `VectorSearchResult` objects with annotations and similarity scores
 
 4. **Result Conversion**
--- a/docs/architecture/object_storage_vector_search.md
+++ b/docs/architecture/object_storage_vector_search.md
@@ -92,16 +92,28 @@ list[Model instance] with .similarity_score  ← identical contract to pgvector
 - **Namespace** = `(parent kind, embedder_path, dimension)` — mirroring the
   filters the pgvector path applies (`embedder_path` + `vector_<dim>`
   column). See `router.build_namespace`.
-- **Consistency**: reads always list the WAL tail and overlay it on segment
-  results (newest upsert/tombstone per id wins), so a write is searchable the
-  moment its WAL PUT returns — same "strongly consistent by default" contract
-  turbopuffer advertises. The manifest GET per query is the equivalent of
-  their consistency roundtrip.
+- **Consistency**: reads list the WAL tail **before** reading the manifest —
+  this ordering is load-bearing. Compaction commits the new manifest strictly
+  before GC'ing folded WAL files, so a WAL-first reader sees either the
+  folded files (served from the overlay) or a manifest that already includes
+  them; a manifest-first reader could pair an old manifest with an
+  already-GC'd tail and silently drop writes. The manifest records the WAL
+  files folded into its generation (`folded_wals`); readers skip those when
+  building the overlay so lingering folded files are never replayed over
+  newer segment state. Net effect: a write is searchable the moment its WAL
+  PUT returns (the same "strongly consistent by default" contract turbopuffer
+  advertises), with the manifest GET per query as the consistency roundtrip.
 - **Compaction** folds segments + WAL into generation N+1; the manifest PUT
-  is the atomic commit point; folded WAL files and the old generation are
-  garbage-collected best-effort afterwards (orphans are harmless: upserts are
-  idempotent and replay order is preserved by time-ordered WAL names).
-  A Django-cache lock serialises compactors per namespace.
+  is the atomic commit point. GC is **deferred by one compaction cycle**:
+  committing generation N+1 deletes the WAL files and segment blobs that
+  generation *N* superseded, never its own — so in-flight readers holding
+  manifest N can still fetch every blob it references instead of hitting
+  `ObjectNotFound` and falling back to pgvector mid-race. Deletion failures
+  are harmless (readers skip folded files via `folded_wals`; dead-generation
+  segment blobs are never referenced again). A Django-cache lock serialises
+  compactors per namespace — which requires a **shared cache backend**
+  (e.g. Redis) in multi-worker deployments; system check `opencontracts.W003`
+  warns when the backend is enabled over `LocMemCache`.
 - **Caching**: centroid and cluster blobs are immutable per generation and
   held in a per-process LRU (`OBJECT_INDEX_CACHE_MAX_ENTRIES`), giving the
   warm-query tier. Cold queries pay object-storage GETs, exactly as designed.
@@ -167,9 +179,26 @@ Parameters*.
   structure; real embedding corpora have it, uniform random data does not
   (recall test in the suite uses clustered fixtures for this reason). Raise
   `OBJECT_INDEX_NPROBE_RATIO` / `OBJECT_INDEX_NPROBE_MIN` for higher recall.
+- **Write visibility lags the Postgres commit.** The WAL PUT happens in a
+  Celery task dispatched on transaction commit, not synchronously inside
+  `store_embedding` — so "searchable the moment the WAL PUT returns" is
+  measured from the task's PUT, and there is a queue-depth-dependent window
+  after the Postgres write during which an already-indexed namespace won't
+  return the new vector. (pgvector fallback still covers *never-indexed*
+  namespaces, not this per-row lag.) turbopuffer's client-synchronous write
+  ack does not have this window.
+- **WAL ordering across workers uses wall-clock names** (`time.time_ns()`),
+  so last-writer-wins between two re-embeds of the same parent on different
+  Celery workers is subject to clock skew. Low impact in practice —
+  embeddings are regenerated infrequently and deterministically per
+  `(parent, embedder)` — and any skew is healed by the next rebuild.
 - **No group commit.** Every `store_embedding` is one WAL PUT (batch
   producers already batch upstream). turbopuffer batches concurrent writers
   per namespace; we could buffer through Redis if PUT volume matters.
+- **Sequential WAL-tail GETs.** A query fetches unfolded WAL files one at a
+  time (tail is bounded by `OBJECT_INDEX_COMPACT_MIN_WAL_FILES`); fetching
+  them concurrently would cut cold/tail latency if p50 matters before
+  compaction catches up.
 - **Deletes are lazy.** Parent deletions are not tombstoned automatically;
   stale ids are dropped by the ORM refilter at query time and purged on the
   next `rebuild_object_vector_index`/compaction cycle.
--- a/docs/architecture/object_storage_vector_search.md
+++ b/docs/architecture/object_storage_vector_search.md
@@ -135,9 +135,15 @@ truth).
 
 Enabling the flag can never make search worse than pgvector: unsupported
 models, never-indexed namespaces, and any engine exception all fall back to
-the pgvector path inside `search_by_embedding` (logged). An invalid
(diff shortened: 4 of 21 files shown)


JSv4:
Both reviews addressed in the latest commits. The blocking finding — the manifest/WAL read-order race — was real and is fixed, along with the adjacent GC races:

**Fixed**
- `search()`/`compact()` now list the WAL **before** reading the manifest (the dangerous interleave was old-manifest + already-GC'd tail = silent data loss). Locked in by a regression test that runs two full compaction cycles between the two reads (`test_search_survives_compaction_between_wal_list_and_manifest_read`).
- The manifest now records `folded_wals`; readers/compactors skip those files when building the overlay, so a lingering folded file (failed or late GC) can never replay stale entries over newer segments or resurrect tombstoned ids (`test_partial_gc_failure_cannot_resurrect_tombstoned_ids`).
- **Deferred GC by one compaction cycle** (both reviews' old-generation-GC concern): committing generation N+1 deletes only what generation N superseded, so in-flight readers holding manifest N can still fetch every blob it references — no more `ObjectNotFound` → spurious pgvector fallback bursts per compaction. `wal_tail_count` now counts only unfolded files so deferred GC can't re-trigger auto-compaction forever (`test_deferred_gc_reclaims_prior_cycle`, `test_wal_tail_count_ignores_folded_lingering_files`).
- `put_bytes` collision recovery now tolerates a third writer deleting the uniquified blob mid-recovery (falls back to the payload in hand).
- New system check `opencontracts.W003` warns when `VECTOR_SEARCH_BACKEND=object_storage` runs over a process-local cache (LocMemCache/dummy), where the compaction cache-lock can't serialise compactors across workers; E002 and W003 both now have tests.

**Documented as accepted limitations** (design doc "Current limitations"): the async write-visibility window (WAL PUT happens in an on-commit Celery task, unlike turbopuffer's synchronous ack), wall-clock WAL ordering under worker clock skew, and sequential WAL-tail GETs (concurrent fetch listed as follow-up).

**Not taken**: consolidating the parent-kind literals across `router`/`vector_index_tasks` (different key spaces, as noted — the values are pinned to `Embedding`'s FK names and covered by tests); locking `get_default_engine` (documented why the unlocked lazy init is safe today). The `base.py` blank-line nit didn't materialize — pre-commit passes.

Suite is now 25 tests; all 154 pre-existing search/embedding regression tests and the MinIO real-S3 smoke run still pass.

---

JSv4:
Third-round findings addressed:

- **`put_bytes` recovery** — rewritten as a bounded delete-then-save retry loop: each attempt discards our stray uniquified blob and retries until the save lands at the exact key (terminal state is always exactly one blob at the key); after `OBJECT_STORE_PUT_OVERWRITE_MAX_ATTEMPTS` we concede last-writer-wins to the racing writer and log — contention on the manifest means the compaction lock was breached, and their blob is a valid manifest. The read-back GET is gone (you were right that it bought nothing). Direct test added (`test_put_bytes_retries_when_racing_writer_uniquifies_save`) forcing a `save()` uniquification and asserting final content + no strays.
- **"Never worse than pgvector" for heavily-filtered querysets** — now actually true rather than aspirational: if re-filtering consumes a *truncated* candidate set (engine returned a full `fetch_n`) without filling `top_k`, the router falls back to pgvector; a short result from an *exhausted* namespace is genuinely complete and returned as-is. Regression test swamps the candidate set with 10 other-user docs and asserts the caller still gets their 2 docs via the fallback. Docs updated to state the rule instead of hand-waving the oversample.
- **Lock TTL ↔ recluster scaling** — comment on `OBJECT_INDEX_COMPACT_LOCK_TIMEOUT_SECONDS` now ties the 600s ceiling to the ~1M-vectors-per-namespace note and names the failure mode if breached.
- Redundant `max()` in `fetch_n` dropped.

Suite is now 28 tests; all regression suites remain green.

---

JSv4:
Fourth-round findings addressed:

- **Storage exposure (the confirmation you asked for)** — the requirement is now explicit and enforced-as-warning: the design doc's Permissions section states the index prefix **must live in non-publicly-readable storage** and why (no per-blob ACL; a readable bucket leaks raw vectors + parent ids outside Django auth, a larger blast radius than pgvector). New system check `opencontracts.W004` warns at boot when `VECTOR_SEARCH_BACKEND=object_storage` runs over an AWS config showing public-read signals (`AWS_DEFAULT_ACL=public-read*`, `AWS_QUERYSTRING_AUTH=False`, or `AWS_S3_CUSTOM_DOMAIN`), with tests covering each signal plus the private/LOCAL/pgvector negative cases. Note the repo's default AWS config (`AWS_QUERYSTRING_AUTH=True`, no default ACL) passes clean.
- **Embedding-row deletion trap** — flagged in both places a future author would look: `EmbeddingManager`'s docstring now warns that deleting an `Embedding` row independent of its parent must tombstone the namespace entry (`ObjectStorageVectorEngine.delete`) or trigger a rebuild, and the design doc's lazy-deletes limitation spells out the same.
- **Per-write cost** — the "no group commit" limitation now quantifies it (~3 object-store roundtrips per `store_embedding`) and directs bulk backfills to `rebuild_object_vector_index`, which batches hundreds of vectors per PUT.
- **k-means reseed nit** — empty clusters now reseed from *successive* worst-fit points, so multiple empty clusters in one iteration can't collapse onto the same seed.

Suite is now 29 tests; regression suites remain green.

---

JSv4:
Rounds 5–6 addressed in the latest commit:

**Fixed**
- **Compaction storm** — `sync_embedding_to_object_index` now claims a cache pending-marker before enqueueing, so a sustained burst past the WAL threshold enqueues exactly one compaction task instead of one per write; the marker clears when compaction finishes (or by TTL if the queued task is lost, letting the next crossing re-trigger). Test: `test_compaction_enqueued_once_per_threshold_crossing`.
- **W004 GCS gap** — the exposure check now also covers `STORAGE_BACKEND=GCP` (`GS_DEFAULT_ACL` public values, `GS_QUERYSTRING_AUTH=False`), with tests for both signals plus the private-config negative case.
- **Rebuild operability** — `rebuild_object_vector_index` gained `--dry-run` (namespaces + counts, no writes), a progress line every 10k embeddings, and help text setting the runtime expectation (full WAL replay + full recluster per namespace; scope with `--embedder-path`).

**Documented**
- GCS is explicitly marked **unverified** in the design doc (`listdir` semantics are load-bearing for the WAL overlay) — run a smoke test before enabling on GCP.
- Limitations now note that deployment-wide namespaces can push narrow corpus-scoped queries onto the shortfall fallback at large multi-tenant scale, with corpus-sharded namespaces named as the follow-up.

**Accepted as documented trade-offs** (round 6): the 600s lock TTL vs. compaction duration coupling (a portable compare-and-swap doesn't exist in the abstract Django Storage API — S3 conditional PUTs would be a backend-specific enhancement; the constant's comment ties the two limits together), and `put_bytes` reliance on concrete-backend uniquify semantics (verified on FileSystemStorage + real MinIO).

Suite is now 30 tests; regression suites remain green.

---

JSv4:
Round 7's manifest read/write race is confirmed and fixed. You were right that it was worse than a doc-note case: a reader catching the delete-then-save window got `manifest=None` with a non-empty WAL, served only the WAL tail, and — as you traced — the shortfall rule read the small hit count as "namespace exhausted" and returned the truncated results as final, with no fallback.

**Fix**: `search()` and `wal_tail_count()` now read the manifest through a retry guard whenever the namespace demonstrably holds data (non-empty WAL): a `None` result is retried once after 50ms (`OBJECT_INDEX_MANIFEST_RETRY_DELAY_SECONDS`) before being believed. A genuinely pre-first-compaction namespace pays one extra GET; empty namespaces keep the fast single-read path. Regression test `test_search_retries_manifest_read_hit_mid_overwrite` simulates the mid-overwrite `ObjectNotFound` and asserts full results.

**Documented**: the design doc's limitations now describe the non-atomic manifest commit, the retry mitigation, the residual risk (retry also landing in the window degrades one query to WAL-tail-only and self-heals), and names S3 conditional PUTs as the backend-native way to remove the window entirely.

**Also taken**: W004's runtime hint now states it's best-effort (Django settings signals only — bucket policies set outside Django are invisible to it), matching the docstring's honesty.

The `compact()` full-materialization memory profile is intentional and documented as the ~1M-vectors-per-namespace scaling bound, as you suspected.

Suite is now 31 tests; regression suites remain green.

---

JSv4:
Round 8 addressed — all items taken:

- **Rebuild-command lock gap (the one you flagged as worth fixing)** — `rebuild_object_vector_index` now acquires the same per-namespace cache lock as the auto-compaction task (shared `compact_lock_key` helper) before compacting. If a compaction is already running, it replays the WAL and defers folding to the next cycle with a warning instead of racing `engine.compact()`. Test: `test_rebuild_command_respects_compaction_lock`.
- **`put_bytes` silent drop** — now raises `ObjectStoreWriteError` after retries are exhausted, so a compaction whose manifest never persisted fails visibly in Celery / the command rather than reporting success. Test: `test_put_bytes_raises_when_overwrite_retries_exhausted`.
- **Sync task retries** — `sync_embedding_to_object_index` gained `autoretry_for` with backoff (3 tries, idempotent upsert), so transient object-store errors no longer silently drop an embedding until the next rebuild.
- **`fetch_n` cap** — bounded by new `OBJECT_INDEX_MAX_FETCH_CANDIDATES` (2048) independent of caller `top_k`; cap-truncated shortfalls take the pgvector fallback, so correctness is preserved.
- **DRY / boundary comments** — both parent-kind maps now cross-reference each other with an explicit "kind values must match exactly or reads and writes land in different namespaces" warning, and the shortfall rule documents the accepted exactly-`fetch_n` false positive.
- **End-to-end caller test** — `test_toggle_serves_core_annotation_vector_store_callers` drives a real `CoreAnnotationVectorStore` vector search (the store behind GraphQL semantic search / MCP / agents) with the toggle on and proves the results come from the object index with pgvector's `CosineDistance` patched to raise — the "zero caller changes" claim now holds by test, not just inspection.

Suite is now 34 tests; regression suites (142 in this round's run) remain green.

---

JSv4:
Round 9 addressed:

- **Single-source parent-kind taxonomy (your suggestion #1, taken as proposed)** — new `EMBEDDABLE_PARENT_KINDS` constant in `constants/search.py` maps `model_name -> (fk_attr, kind)`; both `PARENT_KIND_BY_MODEL_NAME` (router) and `PARENT_FK_TO_KIND` (tasks) are now derived from it, so the silent-divergence → different-namespaces failure mode is structurally impossible rather than comment-guarded.
- **Integer-PK assumption** — documented on that same constant: the engine stores ids as int64 (`int(doc_id)`), every listed parent uses an integer AutoField pk, and adding a UUID-pk parent requires changing the engine's id representation first.
- **Request-thread sleep** — noted on `OBJECT_INDEX_MANIFEST_RETRY_DELAY_SECONDS` that the retry sleep can run on a request-serving thread (rare mid-overwrite window only; keep it small).
- **`--dry-run` coverage** — `test_rebuild_command_dry_run_reports_without_writing` asserts the preview output and that nothing lands in the object store.

On the flagged-not-blocking items: the lock-TTL-expiry double-compactor scenario stays a documented trade-off (the constants comment ties the TTL to the namespace-size ceiling; same-generation cluster interleaving resolves at the manifest PUT via last-writer-wins + `ObjectStoreWriteError` on the loser since round 8). The `np.load` `allow_pickle=False` observation is correct — that default is relied upon deliberately. Corpus-sharded namespaces stay in the design doc's limitations as the named follow-up; happy to open a tracking issue if maintainers want one.

Suite is now 35 tests; regression suites remain green.

---

JSv4:
Round 10 addressed:

- **#1 dead code — confirmed and removed.** `namespace_exists()` had no callers (`search()` returning `None` already serves the never-written signal); dropped per the no-dead-code rule.
- **#4 `top_k <= 0` guard — added.** `search_via_object_index` now returns `None` for degenerate `top_k`, deferring to pgvector so behavior stays byte-identical to the default backend for any caller quirk.
- **#3 clock-skew doc note — already present.** The design doc's limitations section has carried "**WAL ordering across workers uses wall-clock names** (`time.time_ns()`) … subject to clock skew … healed by the next rebuild" since an earlier round — no change needed.
- **#2 embedding-deletion tombstone** — stays a documented gap (EmbeddingManager docstring + design doc limitations, both pointing at `ObjectStorageVectorEngine.delete`/rebuild). A grep-based lint felt heavier than warranted for a path with zero call sites today; happy to add one if maintainers disagree.

Suite remains 35 tests, all green.

---

JSv4:
Round 11's cap-bypass finding is confirmed and fixed — good catch, and correct in every detail: `max(top_k, CAP)` neutralised the cap precisely when `top_k > CAP`, and `top_k` does reach this path unvalidated from GraphQL (`searchConversations`). It's now `fetch_n = min(top_k * OBJECT_INDEX_FILTER_OVERSAMPLE, OBJECT_INDEX_MAX_FETCH_CANDIDATES)`, exactly as the round-8 commit message claimed it should have been; when the cap truncates below `top_k`, the existing shortfall rule falls back to pgvector (whose SQL `LIMIT` has no equivalent exposure). Regression test `test_fetch_cap_binds_for_abusive_top_k` spies on the engine call with `top_k = 10× cap` and asserts the bound holds.

The minor pending-marker interaction was also closed while fresh: `rebuild_object_vector_index` now clears the auto-compaction pending marker after a manual compaction (which just serviced it), so a stale marker can no longer delay the next threshold-crossing enqueue by up to the TTL.

Suite is now 36 tests; regression suites remain green.

---

JSv4:
Round 12's read-path coverage finding is confirmed and addressed — you were right, and the original "single `CosineDistance` call site" premise was wrong: `ConversationQuerySet.search_by_embedding`, `ChatMessageQuerySet.search_by_embedding`, and `CoreRelationshipVectorStore.search` all carry their own inline implementations that never see the toggle (and notably with a *different result contract* — QuerySet + distance-valued `similarity_score`, ascending — so routing them through `search_via_object_index` is not a drop-in change).

Taken option (b), scoped down for symmetry:

- `EMBEDDABLE_PARENT_KINDS` now lists only **document/annotation/note** — the kinds whose querysets actually inherit the mixin — with the exclusion rationale and the migration recipe (route the read path through `search_via_object_index` first, then add the kind) documented on the constant.
- `enqueue_embedding_index_sync` skips non-indexed parent kinds before queuing any Celery task, so conversation/message/relationship writes no longer produce index writes nobody reads.
- Design doc: indexed kinds stated in the architecture section; new limitations bullet names the conversation/message/relationship read-path migration as the follow-up. The PR description's caller-coverage claim is corrected accordingly.
- New test `test_non_indexed_parent_kinds_do_not_fan_out` locks in the symmetry both ways (conversation embedding → no fan-out; document embedding → fan-out). Conversation and subtree-group (relationship) search suites verified green against the change.

Separately (from the session's ongoing goal): **GCS is no longer unverified** — the full engine lifecycle now runs green against the GCS JSON API via fake-gcs-server + django-storages `GoogleCloudStorage` (WAL over prefix-emulated `listdir`, manifest delete-then-save overwrite, multi-generation deferred GC, parity, tombstones), with zero adapter changes needed; steps recorded in `docs/test_scripts/object_storage_vector_backend_gcs.md`. All three storage backends (LOCAL/AWS via MinIO/GCP via fake-gcs-server) are now verified.

Suite is now 37 tests; regression suites remain green.

---

JSv4:
Round 13:

**Finding 1 (worker-upload bypass) — confirmed and fixed.** Genuinely new and the more serious of the two: `worker_uploads/tasks.py` writes `Embedding` rows via `bulk_create`/`update_or_create`, bypassing `store_embedding`, so externally-ingested documents/annotations would silently never reach the object index (non-empty namespace ⇒ no fallback ⇒ a per-row gap only a manual rebuild repairs). `_store_embeddings` and `_store_single_embedding` now call `enqueue_embedding_index_sync` explicitly per row (no-op when the backend is off), the `hooks.py` docstring and design-doc diagram no longer overclaim a "single chokepoint" (in-app pipeline chokepoint + the worker-upload bulk path hooked explicitly, with a warning for future direct-write producers), and `test_worker_upload_direct_writes_fan_out` covers both paths end-to-end into the index. The 132 worker-upload regression tests stay green.

**Finding 2 (six-kind taxonomy mismatch) — already fixed in round 12**, which the checked-out branch appears to predate: `EMBEDDABLE_PARENT_KINDS` is now scoped to document/annotation/note only, the write hook skips non-indexed kinds (so relationship embeddings from `embeddings_task.py` no longer fan out), and `test_non_indexed_parent_kinds_do_not_fan_out` locks in the symmetry. The conversation/message/relationship read-path migration is the documented follow-up.

The multi-dimension-per-row observation is correct and pre-existing; the rebuild command indeed iterates all `DIM_TO_FIELD_MAP` entries per row deliberately. And agreed on the human-review note — the automated rounds are no substitute for a maintainer reading `engine.py` end-to-end before merge.

Suite is now 38 tests; all regression suites remain green.

---

PR #2139, opened 2026-07-08
Migrate GraphQL layer from graphene to strawberry (query shapes pinned by golden SDL)

JSv4:
# Graphene → Strawberry migration

**Status: complete and green — ready to merge.**

## What this does

Replaces the graphene / graphene-django GraphQL layer (schema types, resolvers, mutations) and the graphene-level auth middlewares with strawberry-graphql, with a machine-verified guarantee of **zero query-shape changes**:

- **Golden SDL contract**: the full graphene SDL was captured at migration time as `config/graphql/schema.graphql` (10.6k lines). `opencontractserver/tests/test_schema_parity.py` structurally compares the served strawberry schema against it — every type, field, argument name/type, nullability wrapper, interface, enum member, and printed default must match exactly. The parity test is green.
- **Shared runtime** (`config/graphql/core/`): relay global IDs + `Node` interface in graphene wire format (`base64("TypeName:pk")`), countable/PDF-page-aware connection factories, a faithful port of graphene-django's connection resolution (arrayconnection cursors, `RELAY_CONNECTION_MAX_LIMIT=100`, 1-based `offset`→`after` conversion), django-filter FilterSet argument mapping incl. `GlobalIDFilter` decoding, `GenericScalar`/`JSONString`/`BigInt` scalars, permission-annotation resolvers, DRF-serializer mutation bases, auth decorators with graphql_jwt-compatible error messages.
- **Auth middlewares replaced**: `graphql_jwt.middleware.JSONWebTokenMiddleware` + the API-key graphene middleware are gone. Per-request authentication happens once in `config/graphql/views.py::GraphQLView.get_context` via the standard `AUTHENTICATION_BACKENDS` chain (JWT / Auth0 / API-key / session). `tokenAuth` / `verifyToken` / `refreshToken` are strawberry-native ports preserving long-running refresh-token laziness, and are now rate-limited (see Security hardening below).
- **Security hardening preserved and extended**: `DepthLimitValidationRule` + `DisableIntrospection` attach via strawberry's `AddValidationRules` (which appends to the full spec rule set — the graphene replace-the-rules trap is structurally impossible). The GCS file-URL pre-warm middleware became a strawberry `SchemaExtension`. Two review rounds caught and fixed IDOR regressions introduced by the port itself: singular `xxx(id:)` / relay `node(id:)` lookups for 14+ model-backed types (`test_singular_node_idor.py`), and FK/relay-FK traversal fields leaking private targets across a permission boundary (`test_fk_visibility_traversal.py`, `config/graphql/core/relay.py::resolve_visible_fk`). `tokenAuth`/`updateMe` — the only two mutations with no rate limiting — are now covered (`RateLimits.AUTH_LOGIN` IP-keyed for login, `WRITE_LIGHT` user-tier for profile updates).
- **Tests keep their substantive cases**: `graphene.test.Client` → drop-in `config/graphql/testing.py::Client` (same result dict shape); `GraphQLTestCase` ported for endpoint-level tests; `schema.execute(...)` → `schema.execute_sync(...)`.
- `graphene-django` removed from requirements/`INSTALLED_APPS`; `strawberry-graphql==0.320.3` added; `django-graphql-jwt` retained only as a JWT signing/backend utility (its graphene middleware/mutations are no longer used).

## How the port was done

The strawberry schema skeleton (394 types, all field/argument shapes) was generated by introspecting the live graphene schema, guaranteeing shape parity by construction. Custom resolver bodies (465 stubs) were ported verbatim module-by-module — roots are Django model instances in both frameworks, so bodies transfer nearly unchanged, keeping the service-layer permission patterns intact (the `opencontracts.E001` architecture check stays green).

## Verification

- `test_schema_parity` — green, zero SDL drift
- Full backend suite — **10380 passed, 0 failed, 26 skipped** (skips are pre-existing/environment-gated, unrelated to this PR)
- Codecov patch coverage — **96.34%** (target 89.40% backend / 83.19% overall)
- IDOR hardening: singular-node lookups (`test_singular_node_idor.py`) and FK-traversal visibility (`test_fk_visibility_traversal.py`) both pinned with structural tests that fail if any future registered type lacks its permission hook
- A CorpusType FK-traversal caching regression (recursive CTE storm on `annotation.corpus` access, `test_doc_annotations_prefetch_n_plus_one.py`) was found and fixed post-merge-with-main without reintroducing the stale-context leak the original uncached design guarded against
- Three pre-existing (pre-dating this migration, confirmed against graphene `main`) issues surfaced by the coverage pass were fixed rather than left as follow-ups: `tokenAuth`/`updateMe` had no rate limiting; `resolve_my_permissions` read a permission-dict key (`can_publish_model_type`) that the helper never wrote (always `can_publish`), permanently dead-ending the `publish_{model}` grant; and a duplicate, unregistered standalone customs-enrichment service riding along on this branch (superseded by main's grammar-tier implementation) was removed


JSv4:
Thanks — the 🔴 IDOR finding was spot-on and is now fixed in `99202ae`, and I extended the audit as you suggested.

**Method:** I grepped every `get_node_from_global_id(info, id, only_type_name="X")` call site and cross-checked each target `X` against its `register_type` hooks. That surfaced the 7 you flagged **plus 7 more** the same bug affected:

- Your 7: `relationship`, `annotationLabel`, `labelset`, `chatMessage`, `fieldset`, `column`, `datacell`
- Also fixed: `analyzer`, `gremlinEngine` (its `only_type_name` is `GremlinEngineType_READ` — the underscore made it easy to miss), `agent` (`AgentConfigurationType`), `badge`, `userexport`, `userimport`, and `assignment`

Each now registers a `get_node` hook mirroring the graphene resolver — `BaseService.get_or_none` / `filter_visible`, `AgentConfigurationService.get_agent_by_id` for `agent`, and `assignment` keeps its deprecated superuser-or-participant gate. Of the types you named as unverified: `moderationAction` was already safe (custom gated resolver via `conversation.can_moderate`, not the fallback path); `corpusReference`/`noteRevision`/`corpusCategory`/`corpusAction` aren't exposed as singular by-ID queries.

**🟠 medium:** also fixed — `GraphQLView.dispatch` now catches DRF `AuthenticationFailed` alongside `JSONWebTokenError`, so a bad API key returns a 200 GraphQL error instead of an unhandled 500.

**Regression insurance:** added `test_singular_node_idor.py` — a structural guard that fails if *any* type resolved via `get_node_from_global_id` lacks a permission hook (so this class can't silently return), plus a behavioral non-owner-denied check.

The two 🟡 pre-existing items (`GlobalIDFilter` type-name discard, `resolve_object_shared_with` `KeyError`) are faithful ports and out of scope for this migration — leaving them as-is for a separate follow-up as you suggested.

---

PR #2190, opened 2026-07-24
Fix local login race condition + WebSocket URL fallback (closes #2104)

### LINKED ISSUE #2104: Local version: Could not get details from local server Error
I'm running a local version of cite on a RHEL VM. When I spin up the containers, everything looks fine but when I login, I get a "Could not get details from local server" error. When I look at the developer tools in Chrome, I see that there is a error fetching backend user caused by an invariant violation. I can't find anything in the docker compose logs about what's causing the problem.  I can create users fine but I can't upload any documents except through the admin website.

JSv4:
## Summary

Fixes #2104 — a local (non-Auth0) deployment where login fails with a "Could not get user details from server" toast, an Apollo `InvariantViolation` (`clearStore()`) in the console, and a WebSocket connection to `ws://localhost:<port>/ws/notification-updates/` failing repeatedly.

Two independent root causes reproduce every symptom in the report:

**1. Login/`clearStore()` race (`frontend/src/views/Login.tsx`)**

`LOGIN_MUTATION`'s `onCompleted` set `authToken()` *before* awaiting `resetOnAuthChange()` (which calls Apollo's `client.clearStore()`). `App.tsx`'s `GET_ME` query is skipped via `skip: !auth_token || !auth_init_complete` — but in non-Auth0 deployments, `AuthGate` sets `authInitCompleteVar(true)` unconditionally at mount (there's no Auth0 SDK to wait on), so by the time a user submits the login form `auth_init_complete` is already `true`. Setting `authToken()` first therefore let `GET_ME` start fetching immediately, racing `clearStore()`, which cancels in-flight queries and throws an Apollo "store reset while query was in flight" invariant violation — delivered to `GET_ME`'s error handler as "Error fetching backend user" / "Could not get user details from server".

`AuthGate.tsx` already avoids this exact race for the Auth0 login path (see its inline comments) by deferring `authInitCompleteVar(true)` until *after* `resetOnAuthChange()` resolves. `Login.tsx` never got the equivalent fix. This PR reorders `Login.tsx`'s `onCompleted` to await `resetOnAuthChange()` before flipping `authToken`/`userObj`/`authStatusVar`, matching that pattern (with the same best-effort try/catch). `frontend/src/graphql/cache.ts`'s `authInitCompleteVar` doc comment now notes this deliberate exception so a future reader doesn't "fix" `Login.tsx` back to the documented flow and reintroduce the bug.

**2. WebSocket URL never resolves to the backend in the Docker/nginx image (`frontend/src/components/chat/get_websockets.ts`)**

`resolveWsBaseUrl()` only checked `VITE_WS_URL` / `REACT_APP_WS_URL` / `VITE_API_URL` / `REACT_APP_API_URL` via `import.meta.env` / `process.env`. None of these are ever set by this codebase's deployments — every sample env (`docs/sample_env_files/frontend/**`) only sets `REACT_APP_API_ROOT_URL`, and for the Docker/nginx production-style image (`frontend/Dockerfile` + `frontend/conf/conf.d/default.conf`, which is what `local.yml`'s `frontend` service under the `fullstack` profile builds) that value is injected at container start into `window._env_` (`frontend/env.sh`), not into `import.meta.env`. So the resolver always fell through to `window.location.host` — but that nginx config has no `/ws` location block at all, so the WebSocket handshake fails immediately, repeating on every reconnect attempt. This matches the user's second reported symptom exactly (confirmed by a live repro — see Test plan).

Fixed by having `getEnvVar()` also check `window._env_` (using the existing typed `Window._env_` global from `frontend/src/env.d.ts`, no `any` cast needed), and by adding `REACT_APP_API_ROOT_URL` as a fallback key in `resolveWsBaseUrl()` (after the existing explicit-override vars, before the `window.location.host` last resort) — this is the same variable `index.tsx` already uses to build the GraphQL endpoint, so the WS origin now stays in sync with the configured backend.

## Changes

- `frontend/src/views/Login.tsx` — await `resetOnAuthChange()` before setting auth state in `onCompleted`.
- `frontend/src/graphql/cache.ts` — doc-comment addendum on `authInitCompleteVar` explaining `Login.tsx`'s divergent ordering.
- `frontend/src/components/chat/get_websockets.ts` — `getEnvVar()` reads `window._env_`; `resolveWsBaseUrl()` falls back to `REACT_APP_API_ROOT_URL`.
- `frontend/src/views/Login.test.tsx` (new) — regression test asserting `resetOnAuthChange` is called while `authToken()` is still unset, and that login still succeeds if the cache reset rejects.
- `frontend/src/components/chat/__tests__/get_websockets.test.ts` — new cases for `window._env_.REACT_APP_API_ROOT_URL` resolution, priority vs. an explicit override, and the blank-value fallback.
- `changelog.d/2104-local-login-race.fixed.md` — changelog fragment.

## Test plan

- [x] `yarn vitest run src/views/Login.test.tsx` — new tests pass; verified they *fail* against the pre-fix `Login.tsx` (confirms they pin the regression).
- [x] `yarn vitest run src/components/chat/__tests__/get_websockets.test.ts` — all cases (existing + new) pass.
- [x] `yarn vitest run src/components/auth/AuthGate.test.tsx src/hooks/__tests__/useCacheManager.test.tsx src/services/__tests__/cacheManager.test.ts src/hooks/__tests__/useRefetchOnAuthChange.test.tsx src/hooks/__tests__/useNotificationWebSocket.test.ts` — 112 passed, no regressions in related auth/cache/WS suites.
- [x] `yarn tsc --noEmit` — clean.
- [x] `yarn lint` (prettier --check) — clean.
- [x] `yarn run any:check:strict` — clean (433=433, no baseline regression).
- [x] Live repro of root cause #2 against the real production artifacts: built the frontend (`yarn build`), served `dist/` under the actual `frontend/conf/conf.d/default.conf` in an `nginx:1.27.5-alpine` container, and sent a WebSocket-upgrade request to `/ws/notification-updates/` — nginx returns `200 OK` with `index.html` (no `101 Switching Protocols`, no `/ws` proxy exists), reproducing the exact failure mode from the report. Separately ran the real `frontend/env.sh` entrypoint with `OPEN_CONTRACTS_REACT_APP_API_ROOT_URL=http://localhost:8000` (matching every sample env under `docs/sample_env_files/frontend/**`) and confirmed it emits `window._env_ = { REACT_APP_API_ROOT_URL: "http://localhost:8000", ... }` — the exact value my fix's `getEnvVar()` now reads.


JSv4:
Thanks for the independent verification. Addressed both nits:

**Doc-comment addendum (cache.ts):** Added a paragraph to `authInitCompleteVar`'s doc comment (`frontend/src/graphql/cache.ts`) explaining why `Login.tsx` delays `authStatusVar`/`authToken` instead of following the canonical "`authStatusVar` before, `authInitCompleteVar` after" flow — pointing back to `Login.tsx`'s `onCompleted` comment for the full race. Pushed in a8e31bc.

**`getRuntimeEnv()` reuse for the `REACT_APP_API_ROOT_URL` fallback:** Looked into this, and it's not a safe drop-in. `getRuntimeEnv()` (`frontend/src/utils/env.ts:63-66`) bakes in a hardcoded `"http://localhost:8000"` default for `REACT_APP_API_ROOT_URL` when neither `window._env_` nor `import.meta.env` set it. Calling it directly from `resolveWsBaseUrl()` would mean *any* deployment that leaves `REACT_APP_API_ROOT_URL` unset — no `window._env_` and no build-time env — always resolves to `ws://localhost:8000` instead of falling through to the `window.location.host` last resort. That silently breaks same-origin WS resolution for anything not talking to a backend at `localhost:8000` (and it's exactly what the existing `"uses window.location protocol/host fallback"` test in `get_websockets.test.ts` pins). It happens to produce the same URL for the default local `yarn start` case (both land on `localhost:8000`), which is probably why it reads like a safe substitution, but the default-value semantics don't actually match what `getEnvVar`'s "return `undefined` if truly unset" contract needs here. Left it as its own lookup rather than introducing that behavior change — happy to revisit if there's a `getRuntimeEnv()` variant without the baked-in default, but that felt like more surface area than this fix warranted.

Login-loading double-click nit: agreed it's harmless/idempotent as noted, no change made.

---

JSv4:
Good catch on the footer — removed it from the PR description. (Noting for the record: this repo's `CLAUDE.md` explicitly forbids crediting Claude/Claude Code in commit messages, PR messages, comments, or any other artifact, so that applies to my own comments here too — this and future replies from me on this PR won't carry one, overriding the generic attribution convention.)

On the rest:

- **Doc-comment addendum on `authInitCompleteVar`**: already landed in a8e31bc (before this review ran) — `cache.ts` now explains the `Login.tsx` exception inline.
- **DRY / `getRuntimeEnv()` reuse for `get_websockets.ts`**: evaluated this in detail in my earlier reply on this thread — `getRuntimeEnv().REACT_APP_API_ROOT_URL` bakes in a hardcoded `http://localhost:8000` default, which would eliminate the "fall through to `window.location.host` when truly unconfigured" behavior the existing tests pin. Not applying the direct substitution for that reason; happy to factor a shared no-default `window._env_` lookup into `utils/env.ts` as a follow-up if that's wanted, but didn't want to expand this fix's surface area further.
- **Test coverage boundary** (mocked `useCacheManager` vs. real `clearStore()`/`GET_ME` interaction): agreed, that's the trade-off — mocking `resetOnAuthChange` keeps the test fast and deterministic but means it wouldn't catch some *other* future query being gated only on `auth_token`. No change made; flagging accepted as a known limit of this regression test's scope.
- **`login_loading` double-click UX nit**: agreed it's harmless (idempotent re-fire, `CacheManager`'s debounce absorbs the second `resetOnAuthChange()` call) and pre-existing in shape, just shifted by this reorder. No change made — out of scope for this bug fix.

---

JSv4:
Update on the manual Docker verification flagged as outstanding: done. Docker daemon came up in this sandbox after all.

- `yarn build`'d the frontend, then ran the actual `frontend/conf/conf.d/default.conf` in `nginx:1.27.5-alpine` (the same base image `frontend/Dockerfile` uses) and sent a WebSocket-upgrade request to `/ws/notification-updates/`. Result: `200 OK` with `index.html` — no `101 Switching Protocols`, confirming there is genuinely no `/ws` proxy in that image and this is exactly why the reported connection kept failing.
- Ran the real `frontend/env.sh` entrypoint script with `OPEN_CONTRACTS_REACT_APP_API_ROOT_URL=http://localhost:8000` (the value every sample env under `docs/sample_env_files/frontend/**` uses) and confirmed it emits `window._env_ = { REACT_APP_API_ROOT_URL: "http://localhost:8000", ... }` — the exact runtime value `get_websockets.ts`'s `getEnvVar()` now reads to resolve the WS origin correctly.

Both pieces of the root-cause chain for symptom #2 are now confirmed against the real artifacts, not just unit tests against a stubbed `window`. Updated the PR description's test plan accordingly. Full end-to-end verification (an actual browser connecting through the fixed bundle) would need the full `local.yml --profile fullstack` stack including Django/Postgres/Redis, which felt like more infrastructure than this fix warrants standing up — the two checks above isolate and confirm the specific mechanism the fix changes.

---

PR #2192, opened 2026-07-24
Tighten deep-research citation discipline (#2180–#2183)

### LINKED ISSUE #2180: Deep research cites section headers instead of supporting passages
## Problem

Deep-research reports anchor footnotes to **section headers** instead of the passages that actually support the sentence.

Observed in a live run (corpus of 192 SEC filings, ResearchReport 4): claims about a specific fixed-price cost-overrun risk factor and about the absence of quantified input-cost sensitivity were cited to structural anchors whose `raw_text` is literally `"ITEM 1A. RISK FACTORS"` and `"ITEM 3. QUANTITATIVE AND QUALITATIVE DISCLOSURES ABOUT MARKET RISK."` — the header annotations, not the operative language. The supporting sentences exist in the documents; the anchors just point at the section top.

## Why it matters

Citation credibility is the product's core promise. A reviewer clicking a footnote should land on the sentence that proves the claim, not the top of a 20-page section. Header-level anchors also make claim verification (and any future automated citation-checking) much weaker.

## Likely mechanics

The agent cites from `retrieved_annotation_ids` (see the `record_finding` / `finalize_report` closures in `opencontractserver/tasks/research_tasks.py`), and section headers are heavily represented in retrieval results (structural annotations from the corpus's OC_SECTION layer rank well for section-topic queries). Nothing in the research instructions or the citation-validation closure requires the anchor's text to support the sentence.

## Suggested direction

- Instruct the agent (system prompt) that a citation must anchor the passage whose own words support the sentence, never a bare header, and to prefer `search_exact_text_as_sources` for pinpoint anchors once it knows the language it wants to cite.
- Optionally: a finalize-time lint that flags footnotes whose anchor `raw_text` is < N chars or matches a structural/header annotation, prompting the agent to re-anchor.

From a live citation audit of an end-to-end run on real EDGAR data; happy to share the full audit table.

### LINKED ISSUE #2181: Deep research citations don't traverse incorporation-by-reference
## Problem

When a filing incorporates content by reference, deep-research footnotes cite the **referring** document rather than the one that contains the language.

Observed (ResearchReport 4, corpus of 192 SEC filings): the report's fixed-price cost-overrun claim is footnoted to a 10-Q's Part II Item 1A — whose entire body is a cross-reference ("We have described under ... Item 1A of our Annual Report on Form 10-K ..."). The operative language actually lives in the 10-K (and the S-1). The citation only "works" through an incorporation-by-reference hop the report never surfaces.

## Why it matters

A reader who opens the cited 10-Q section finds no supporting text — the citation looks wrong even though the claim is right. SEC-style corpora do this constantly (10-Q risk factors, proxy incorporations, exhibit IBR chains), and this corpus even carries explicit `INCORPORATED_BY_REFERENCE` relationships from enrichment that the researcher ignores.

## Suggested direction

- Agent instruction: "cite the document that contains the language; if you reached it via a cross-reference, follow the reference before anchoring."
- Longer term: let the research tools expose the corpus's `INCORPORATED_BY_REFERENCE` / `HAS_EXHIBIT` relationship edges so the agent can traverse them deliberately.

### LINKED ISSUE #2182: Deep research forces corpus anchors onto prompt-derived claims
## Problem

Claims derived from the research **prompt/task description** (not from corpus documents) get decorated with weak corpus anchors instead of being left uncited.

Observed (ResearchReport 4): the "Event Overview" section — facts about a July 20, 2026 tariff proclamation, supplied entirely by the research prompt — is footnoted to a `"Sanctions"` representation inside an unrelated underwriting-agreement exhibit dated *before* the event. The anchor cannot support the sentence; it appears the agent felt obliged to attach some retrieved annotation to every finding.

## Why it matters

One unsupportive footnote poisons trust in the other six. Uncited context is honest; miscited context is not.

## Suggested direction

- Agent instruction: findings that restate the task/event context should carry **no** citation; citations are only for claims grounded in retrieved passages.
- Optionally, `record_finding` could accept an explicit `source: task_context` marker so the composer renders these as uncited background rather than forcing a footnote.

### LINKED ISSUE #2183: Deep research report duplicates every finding sentence
## Problem

Deep-research report prose duplicates nearly every substantive sentence: each bullet renders the finding text immediately followed by a restatement of the same sentence (the citation-carrying variant), e.g.:

> ...vulnerable to tariff-driven price increases.TCFIII SPACECO HOLDINGS LLC's S-1 filing discloses that its primary raw materials include metals and alloys such as aluminum and copper, which directly connects...[^2][^3]

Observed throughout ResearchReport 4 (all five sections affected; note also the missing space at the join). Looks like the composer concatenates `finding.summary` + `finding.cited_text` (or the salvage/finalize path stitches both fields) without deduplication.

## Why it matters

Doubles report length, reads as a bug, and buries the citations.

## Suggested direction

Inspect the report-composition path in `opencontractserver/tasks/research_tasks.py` (finalize/salvage composition from recorded findings) and render one sentence per finding — the cited variant — or dedupe near-identical adjacent sentences at composition time.

JSv4:
## Summary

Four issues from a live citation audit of a deep-research run (ResearchReport 4, 192 SEC filings) all trace to how the deep-research agent chooses and renders citations. This PR fixes them with one focused, high-leverage change — a `## Citation discipline` section in the agent's system prompt — plus a deterministic, observational backstop for the header-anchor problem.

| Issue | Problem | Fix |
|---|---|---|
| #2180 | Footnotes anchor **section headers** (`"ITEM 1A. RISK FACTORS"`) instead of the supporting passage | Prompt rule + finalize-time weak-citation lint |
| #2181 | Citations don't traverse **incorporation-by-reference** — they cite the referring 10-Q, not the 10-K that holds the language | Prompt rule: cite the doc that *contains* the language; follow the reference first |
| #2182 | **Prompt-derived** context ("Event Overview") gets decorated with unrelated corpus anchors | Prompt rule: task/background restatements carry **no** citation |
| #2183 | Report **duplicates every finding sentence** (`…price increases.TCFIII SPACECO…[^2][^3]`) | Prompt rule: state each claim once inside a single `<cite>` tag |

## Root-cause note on #2183

The issue speculated the composer concatenates `finding.summary` + `finding.cited_text`. It does not: the normal finalize path emits the agent's `markdown_body` verbatim (cite-rendered), the salvage path renders exactly one bullet per finding, and the frontend renders only `content` (never re-rendering `findings`). The observed doubling — a plain sentence joined with no space to a near-identical cited restatement — is **agent-authored prose**, so it's fixed by the write-once prompt rule. New regression tests lock in single-render composition so a future composer change can't reintroduce it.

## Changes

- **`opencontractserver/research/constants.py`** — new `## Citation discipline` section in `build_deep_research_system_prompt` covering all four rules; new `RESEARCH_HEADER_ANCHOR_LABELS` set (`OC_SECTION` + the LlamaParse heading labels `Title` / `Section Header` / `Heading` / `Page Header`) used by the lint.
- **`opencontractserver/research/services/research_reports.py`** — `_is_header_anchor()` / `_normalize_label()` helpers + `_render_citations` now `select_related`s `annotation_label` (no N+1) and tags each citation `anchor_is_header` when its **annotation label** is a section-header label (matched case-/separator-insensitively). `finalize` records a concise weak-citation warning (surfaced in the report UI's existing warning chips) when a footnote anchors a header. Observational only — it never rewrites prose. Detection is keyed on the **label, not `Annotation.structural`**: the parsing pipeline sets `structural=True` on its entire layout layer (body paragraphs, tables, sentence chunks), and the bookmark-derived OC_SECTION headers are `structural=False`, so keying on it would both flood false positives and miss the real headers.
- **Tests** — `test_prompt_documents_citation_discipline`; finalize-lint tests (`OC_SECTION` + LlamaParse-heading label → warning; a `structural=True` body paragraph and a body clause that merely opens with "Section 8.1 …" → no warning; singular + plural warning grammar); single-render composition guards for both `finalize` and `_compose_salvage_body`.
- **Changelog fragment** `changelog.d/2180-2183-research-citations.fixed.md`.

## Testing

- `python -m py_compile`, `black`, `flake8`, and `scripts/collate_changelog.py --check` all pass.
- New unit tests added for every deterministic behavior change, traced against the real code paths. (The Dockerized backend suite couldn't be built inside the authoring sandbox — the egress proxy re-terminates TLS for the container build's `pip` — so it runs in CI; the Codecov patch report confirms the suite executed there.)

## Scope notes

- Only #2180 gets a deterministic, testable backstop (the header lint). #2181 / #2182 / #2183 are prompt-only rules with no runtime enforcement — consistent with the #2183 root-cause finding (agent-authored prose, not a composer bug). Confirming those three change agent behavior in practice needs another live citation audit.
- The lint's `RESEARCH_HEADER_ANCHOR_LABELS` covers `OC_SECTION` + the LlamaParse heading taxonomy. Docling/oc_text_parser headers are caught via the parser-agnostic `OC_SECTION` bookmark layer; a Docling doc with header-like text that never receives an `OC_SECTION` annotation would be missed. Acceptable for an observational lint; a possible follow-up if Docling coverage needs to be explicit.

Closes #2180
Closes #2181
Closes #2182
Closes #2183

> Note on #2181: this implements the immediate agent-instruction fix. The issue's "longer term" idea — exposing the corpus's `INCORPORATED_BY_REFERENCE` / `HAS_EXHIBIT` relationship edges as research tools for deliberate traversal — remains a good follow-up.

### DIFF
--- a/changelog.d/2180-2183-research-citations.fixed.md
+++ b/changelog.d/2180-2183-research-citations.fixed.md
@@ -14,14 +14,15 @@
   (`opencontractserver/research/services/research_reports.py`) now tags each
   citation with `anchor_is_header`, and `ResearchReportService.finalize` records
   a concise warning (surfaced in the report's UI warning chips) when a footnote
-  resolves to a section header / structural anchor instead of a supporting
-  passage. Detection keys solely off `Annotation.structural` — the flag the
-  parsing pipeline sets on OC_SECTION layout annotations, which are exactly the
-  headers that mis-anchored in #2180. It deliberately does *not* also text-match
-  filing-style headings (`ITEM 1A`, `SECTION 8`, …), because legal prose commonly
-  opens an operative clause with its section number (“Section 8.1 requires …”),
-  which would false-positive on real citations in contract corpora. The lint is
-  observational — it never rewrites report prose.
+  resolves to a section header instead of a supporting passage. Detection keys on
+  the anchor's **annotation label** (`RESEARCH_HEADER_ANCHOR_LABELS`:
+  `OC_SECTION` + the LlamaParse heading labels `Title` / `Section Header` /
+  `Heading` / `Page Header`, matched case-/separator-insensitively) — *not* on
+  `Annotation.structural`, which the parsing pipeline sets on the entire layout
+  layer (body paragraphs, tables, sentence chunks) and which the bookmark-derived
+  OC_SECTION headers are `structural=False` for; keying on it would both flood
+  false positives and miss the real #2180 headers. The lint is observational — it
+  never rewrites report prose.
 - **Confirmed the report composer is not the source of the #2183 duplication.**
   The finalize path emits the agent's `markdown_body` verbatim (cite-rendered)
   and the salvage path renders each finding exactly once; the observed doubling
--- a/changelog.d/2180-2183-research-citations.fixed.md
+++ b/changelog.d/2180-2183-research-citations.fixed.md
@@ -1,0 +1,30 @@
+- **Tightened deep-research citation discipline (#2180, #2181, #2182, #2183).**
+  Added a `## Citation discipline` section to the deep-research system prompt
+  (`build_deep_research_system_prompt` in
+  `opencontractserver/research/constants.py`) that instructs the agent to: anchor
+  the passage whose own words support a claim and never a bare section header —
+  reaching for `search_exact_text_as_sources` to pull a pinpoint anchor (#2180);
+  cite the document that actually *contains* the language, following any
+  incorporation-by-reference / cross-reference hop to its source before anchoring
+  (#2181); leave task/prompt-derived background *uncited* rather than forcing a
+  corpus anchor onto it (#2182); and state each claim exactly once inside a
+  single `<cite>` tag, never as a plain sentence followed by a cited restatement
+  (#2183).
+- **Added a finalize-time weak-citation lint (#2180).** `_render_citations`
+  (`opencontractserver/research/services/research_reports.py`) now tags each
+  citation with `anchor_is_header`, and `ResearchReportService.finalize` records
+  a concise warning (surfaced in the report's UI warning chips) when a footnote
+  resolves to a section header / structural anchor instead of a supporting
+  passage. Detection keys off `Annotation.structural` (OC_SECTION headers) plus a
+  length-gated filing-heading heuristic (`ITEM 1A`, `PART II`, …); the new
+  `MAX_HEADER_ANCHOR_CHARS` constant keeps long operative passages that merely
+  begin with such a token eligible as real citations. The lint is observational —
+  it never rewrites report prose.
+- **Confirmed the report composer is not the source of the #2183 duplication.**
+  The finalize path emits the agent's `markdown_body` verbatim (cite-rendered)
+  and the salvage path renders each finding exactly once; the observed doubling
+  was agent-authored prose (a plain sentence joined to a near-identical cited
+  restatement), now addressed by the prompt rule above. New regression tests
+  (`test_finalize_renders_each_claim_once`,
+  `test_compose_salvage_body_renders_each_finding_once`) lock in single-render
+  behavior so a future composer change can't reintroduce it.
--- a/changelog.d/2180-2183-research-citations.fixed.md
+++ b/changelog.d/2180-2183-research-citations.fixed.md
@@ -15,11 +15,13 @@
   citation with `anchor_is_header`, and `ResearchReportService.finalize` records
   a concise warning (surfaced in the report's UI warning chips) when a footnote
   resolves to a section header / structural anchor instead of a supporting
-  passage. Detection keys off `Annotation.structural` (OC_SECTION headers) plus a
-  length-gated filing-heading heuristic (`ITEM 1A`, `PART II`, …); the new
-  `MAX_HEADER_ANCHOR_CHARS` constant keeps long operative passages that merely
-  begin with such a token eligible as real citations. The lint is observational —
-  it never rewrites report prose.
+  passage. Detection keys solely off `Annotation.structural` — the flag the
+  parsing pipeline sets on OC_SECTION layout annotations, which are exactly the
+  headers that mis-anchored in #2180. It deliberately does *not* also text-match
+  filing-style headings (`ITEM 1A`, `SECTION 8`, …), because legal prose commonly
+  opens an operative clause with its section number (“Section 8.1 requires …”),
+  which would false-positive on real citations in contract corpora. The lint is
+  observational — it never rewrites report prose.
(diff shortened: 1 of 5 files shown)


JSv4:
Good catch — fixed in 608a691.

You're right that the text fallback false-positives on operative clauses that merely open with a section reference ("Section 8.1 requires 30 days' written notice…"), which is common in contract corpora. I took the second option you suggested: dropped the heading-token regex (and the now-unused `MAX_HEADER_ANCHOR_CHARS`) and keyed `_is_header_anchor` **solely** on `Annotation.structural` — the flag the parsing pipeline already sets on the OC_SECTION layout annotations that mis-anchored in #2180. So the reliable signal stays and the false-positive surface is gone.

Replaced the text-fallback test with a regression guard asserting a non-structural clause opening with a section reference is *not* flagged.

---

JSv4:
Strong catch — you're right, and I verified it against the source. `structural` marks the whole parser layout layer, not headers: `oc_text_parser.py:238` sets `structural=True` on every text chunk, `llamaparse_parser.py:1278` on every element type, and the actual OC_SECTION bookmark headers are explicitly `structural=False` (`pdf_outline_enricher.py:422`). So keying on `structural` both flooded false positives (every `similarity_search` body-chunk citation) and missed the real #2180 headers. My structural-only commit was wrong.

Fixed in a7d38c6 by keying on the **annotation label** instead:

- New `RESEARCH_HEADER_ANCHOR_LABELS` = `OC_SECTION` (the canonical cross-parser section layer, and #2180's own repro) + the LlamaParse heading labels `Title` / `Section Header` / `Heading` / `Page Header` (`ELEMENT_TYPE_MAPPING`), matched case- and separator-insensitively against `annotation_label.text`.
- `_render_citations` `select_related`s `annotation_label` (no N+1). Body labels (`SENTENCE`, `Paragraph`, `Table`, …) are never flagged.
- Added a regression test asserting a `structural=True` body sentence is **not** flagged — the exact flood you flagged — alongside OC_SECTION / LlamaParse-heading positive cases.
- Also fixed the singular/plural grammar ("N citations anchor section headers…").

Docling's exact heading-label strings weren't inspectable here, so the set is centered on the reliable `OC_SECTION` signal and is easy to extend if Docling emits a distinct heading label.

---

JSv4:
Thanks for the thorough verification. On the one actionable point (#1, Docling coverage): the matching is deliberately case- and separator-insensitive, so a snake_case heading label like `section_header` / `page_header` / `title` — the shape Docling's `DocItemLabel` vocabulary tends to use, consistent with the lowercased labels `docling_parser_rest.py:698` already switches on — normalizes into the same set. So the practical gap is narrower than distinct label *names*; only a genuinely different heading vocabulary would slip through. Since the lint is observational I agree that's a follow-up rather than a blocker.

Points 2 and 3 match my understanding — I'll watch the next citation audit to confirm the prompt rules actually move agent behavior in practice.

---

PR #2193, opened 2026-07-24
Verify deep-research quotations against cited source at finalize (#2189)

### LINKED ISSUE #2189: Deep research fabricates quotations when steered to quote cited passages
## Problem — severity above the other deep-research citation issues

When steered to quote the passages it cites ("every footnote must anchor the exact passage — quote that passage"), the deep-research agent produced **fabricated quotations**: quotation-marked strings that do not occur anywhere in the corpus, attached to real (and even well-placed) annotation anchors.

Live evidence (ResearchReport 6, 192-doc SEC corpus): exact-substring search across every document's text layer for the report's six quoted strings returned **0 matches for all six**, e.g.:

- "Prices for these raw materials can be volatile" — 0 docs
- "We do not generally hedge our raw material purchases" — 0 docs
- "On fixed-price contracts, we take the risk of cost overruns, **including those due to increases in raw material costs**" — 0 docs (real text: "when we enter into fixed-price contracts with some of our customers, we take the risk of cost overruns" — the tail is invented)
- "We are subject to risks from changes in U.S. and foreign government trade policies, including **tariffs, quotas**..." — 0 docs

The same run followed the other citation rules well (no header anchors, no forced cite on prompt-derived background, MAE clause pinned exactly) — which makes the fabricated quotes MORE dangerous: the report *looks* rigorously cited.

Baseline: the previous run (ResearchReport 4) with no quote-steering paraphrased honestly and fabricated nothing. So the failure is specifically "quote on demand" → synthesizes quote-shaped text from working memory instead of copying retrieved spans.

## Suggested direction

The codebase already has the right primitive: the enrichment LLM extractors verify every span against source text and reject hallucinated ones (`verify_and_place`, hallucination guard in `opencontractserver/enrichment/`). The research finalize path should do the same for quotations: at `finalize_report`, exact-match (or fuzzy ≥ threshold) every quotation-marked string in the report against the text layers of its footnote's document; drop to paraphrase (strip quote marks) or reject with a re-try instruction on mismatch. A `quoted_verbatim: bool` per finding, validated server-side, would make the guarantee structural.

Related: #2180 (header anchors), #2181 (IBR), #2182 (forced anchors), #2183 (duplication).

JSv4:
## Problem

Fixes #2189. Steered to quote the passages it cites ("every footnote must anchor the exact passage — quote that passage"), the deep-research agent produced **fabricated quotations**: quotation-marked strings that occur nowhere in the corpus, attached to real (well-placed) annotation anchors. Because the same run otherwise followed the citation rules well, the report *looked* rigorously cited — which makes the fabricated quotes more dangerous. The finalize path did no verification of quoted text, unlike the enrichment extractors, which already reject hallucinated spans (`verify_and_place`).

## Fix

At `finalize`, verify every quoted passage against the source it is attributed to — the same primitive the enrichment path already uses, adapted to the research report.

**`ResearchReportService.finalize` → new `_verify_quoted_spans`** (`opencontractserver/research/services/research_reports.py`):
- For each `<cite ids="…">claim</cite>` span, extract the double-quoted passages (straight `"…"` and curly `“…”`) from the claim and check each against the `raw_text` of that span's cited annotation(s).
- Match = whitespace-/case-normalized substring, with a `difflib.SequenceMatcher` longest-contiguous-block fuzzy fallback (`RESEARCH_QUOTE_MATCH_THRESHOLD`, `0.92`) that tolerates trivial punctuation/whitespace/single-character drift but rejects an invented tail or reworded clause.
- A quote that does not verify is **demoted to plain paraphrase** — its quotation marks are stripped, the prose and the footnote are preserved (an honest downgrade, mirroring the sibling `_strip_fabricated_links` post-processor). `report.warnings` gets a flag (mirroring the `#2180` weak-citation lint) so the report UI surfaces it.
- Short quoted strings (`< RESEARCH_QUOTE_MIN_WORDS`, i.e. defined terms like `"Confidential Information"`, scare-quotes, single words) are left untouched — they are rarely fabricated passages and are the main source of false positives.
- Runs on both the normal finalize body and the salvage body (both flow through `finalize`).

**Preventative half** — the deep-research system prompt's "Citation discipline" section (`build_deep_research_system_prompt`, `opencontractserver/research/constants.py`) gains a rule: quotation marks are for verbatim copies only; paraphrase otherwise; use `search_exact_text_as_sources` to pull the pinpoint passage when you need exact words — never reconstruct a quote from memory.

Every citeable annotation id is a real DB row with `raw_text` (retrieval only records positive ids into `retrieved_annotation_ids`; synthetic exact-search ids are never appended and are filtered at finalize), so the anchor text is always available to check against.

## Why strip-and-warn rather than reject-and-retry

`finalize` is the terminal action, so a re-search retry isn't possible at that point without re-running the whole job. Stripping the quotation marks removes the false verbatim claim while preserving the (real) citation and the prose — the same honest-downgrade discipline `_strip_fabricated_links` uses for the agent's invented hyperlinks. The prompt rule is the preventative complement.

## Tests

- `test_research_report_service.py`: fabricated quote is stripped + warned; grounded quote (with whitespace/case drift) is preserved verbatim; short quoted terms are skipped; uncited quotes are untouched; a multi-annotation span matches against any one cited source.
- `test_research_memory.py`: the citation-discipline prompt documents the verbatim-quote rule.

All six of the issue's fabricated example quotes are caught by the matcher (validated against their real vs. invented text).

## Verification note

`py_compile` is clean on all changed files, `black`/`flake8` pass, and every new test scenario passes when executed against the real function bodies (only the `Annotation` `values_list` ORM call stubbed). The Docker test image could not be built in this sandbox (pip cannot reach PyPI through the environment's egress proxy) — CI will run the full suite.

## Related

`#2180` (header anchors), `#2181` (IBR), `#2182` (forced anchors), `#2183` (duplication) — this is the highest-severity of the deep-research citation issues.

---
_Generated by [Claude Code](https://claude.ai/code/session_01YAUayjx9ikxBih6q9y4iWz)_


JSv4:
Thanks — addressed the substantive points in `40a35e8`:

- **Silent bypass on textless anchors** (the main one): `_verify_span` no longer takes a "no candidates → return unchanged" branch. A quote whose cited annotation(s) yield no usable `raw_text` (empty/`None` text, a since-deleted row, or an id that wasn't retrieved) is now treated as ungrounded and demoted like a mismatch — since a passage attributed to a textless anchor can't be a verbatim citation of it. That closes the "cited a real anchor but invented the quote" hole you flagged. New test `test_verify_quoted_spans_demotes_quote_when_anchor_has_no_text`.
- **Curly/straight mismatch:** `_QUOTED_PASSAGE_RE` is unified to `["“]([^"“”\n]{1,N})["”]`, so a passage smart-quoted on only one side is still extracted and verified (adjacent quotes still match separately; the curly apostrophe `’` U+2019 is intentionally not matched). New test `test_verify_quoted_spans_handles_curly_and_mismatched_quotes`.
- **Fuzzy-match wording:** corrected the `RESEARCH_QUOTE_MATCH_THRESHOLD` comment — it's `difflib`-based like the annotation-anchor match but uses longest-contiguous-block coverage, not `.ratio()`.
- **Passage-length cap:** the inline `400` is now `RESEARCH_QUOTE_MAX_CHARS` (2000) so realistic block quotes are still verified; the char class is linear, so it's a match-bounding guard, not ReDoS.

On the `_normalize_for_quote_match` ↔ `annotation_anchoring._norm` duplication: kept it local by intent — decoupling means an anchor-specific tweak to `_norm` won't silently shift quote-verification semantics — with the docstring pointer noting they're currently identical. Happy to promote it to a shared util instead if you'd prefer the single source of truth.

---

PR #2194, opened 2026-07-24
Fix CodeQL code scanning alerts: clear-text secret logging + exception exposure

JSv4:
## Summary

Fixes the 5 open CodeQL code scanning alerts on `main`:

- **#89 (High — Clear-text logging of sensitive information)**
  `config/graphql/pipeline_settings_mutations.py:1288-1294` (`_mutate_UpdateComponentSecretsMutation`) logged the joined list of user-supplied secret *key names* on every `updateComponentSecrets` call. `validate_secrets_input` only bounds a key's length/type, not its content, so a caller could paste an actual secret value into the key field and have it land in plaintext application logs. The log statement now records only the count of updated keys.

- **#79, #80, #81, #82 (Medium — Information exposure through an exception)**
  `compose/accelerated/embedder/main.py`'s `/embeddings/image` and `/embeddings/image/batch` handlers returned raw exception text (`str(e)`) straight into the HTTP response on both the `ValueError` and generic `Exception` branches (lines 186, 188, 242, 244). Image decode/inference failures (bad base64, corrupt image, model/driver errors) could leak internal details to the client. Both handlers now log the full exception server-side via `app.logger` (`.warning(..., exc_info=True)` / `.exception(...)`) and return fixed, generic error messages to the caller instead.

## Test plan

- [x] `python3 -m py_compile` on both changed files
- [x] `black --check` and `flake8` clean on both changed files
- [x] Manually verified no test in `opencontractserver/tests/test_pipeline_settings.py` or `compose/accelerated/tests/` asserts on the specific log/error text that changed
- [ ] Full backend/docker test suite (not runnable in this sandbox — no Docker daemon available)

---
_Generated by [Claude Code](https://claude.ai/code/session_014CoGZtKULawb1ZiGaUWvf3)_


JSv4:
Addressed the optional test-coverage suggestion:

- Added `test_update_component_secrets_mutation_does_not_log_secret_key_names` to `opencontractserver/tests/test_pipeline_settings.py` — uses `assertLogs` to lock in the #89 fix (a secret key name never appears in the log line, only the count).
- Did **not** add an equivalent route-level test for the #79-#82 embedder fix. `compose/accelerated/tests/` is collected by the same `docker compose -f test.yml run django pytest` job as the Django suite, and Flask/numpy aren't in `requirements/base.txt` or `requirements/local.txt` — importing `main.py` there raises `ModuleNotFoundError`. That's exactly why the existing embedder tests (`test_embedder_batching.py`, `test_embedder_service_runtime.py`) only exercise the framework-free helpers (`service_runtime.py`, `batching.py`) and never import `main.py`'s Flask routes directly. Happy to add one if a lighter-weight harness for that service is introduced separately, but doing so here would risk breaking CI collection.


---

PR #2203, opened 2026-07-25
Fix deep-research double render and verify claim support (closes #2200, #2201)

### LINKED ISSUE #2200: Deep research report double-renders and cite spans self-quote (regression from citation fixes)
## Regression introduced by the citation-pipeline fixes (#2192/#2193)

Verified on a post-merge run (ResearchReport 8, same 192-doc SEC corpus/task used to validate #2180–#2189):

1. **The composed report renders twice.** `## Executive Summary` appears 2×: the full report is emitted first with inline `<cite ids="...">` spans, then — after a stub `## Sources` line reading "(All claims above are cited inline...)" — the ENTIRE report renders again in footnote form with `<...>` bracket markers and a real footnote table. Reads like an intermediate representation escaping into the final document.

2. **Cite spans self-quote.** Every `<cite ids="...">...</cite>` span's inner text is a verbatim copy of the claim sentence it decorates — not source text. This is what re-doubles each bullet inside the first rendering (the successor of the old #2183 sentence-doubling), and it means the quotation-verification path from #2193 has nothing real to check against the cited annotations.

Predecessor issue #2183 is closed; this is the new shape of the problem after the fix. Fix direction: emit exactly one rendering (the footnote form), and make the cite-span content either empty (pure marker) or actual quoted source text that #2193's guard then verifies.

### LINKED ISSUE #2201: Deep research citations: verify claim support at finalize and add cite handles to retrieval results
## Follow-up to the #2180–#2189 citation fixes: verify claim SUPPORT, not just quote verbatim-ness — and make citing cheap

Post-merge verification (ResearchReport 8) shows the fixes hold on their targets (no header anchors, no fabricated quotes, MAE clause anchored exactly). Three residuals share one root cause — nothing checks that a cited span actually SUPPORTS its sentence, and finding a supporting span is expensive:

1. **Over-attributed paraphrase** (fabrication moved channels after #2193): e.g. the fixed-price claim gains the tail "including increased prices for raw materials such as aluminum and copper" — no filing says this, but it's presented under "filings confirm". Unquoted, so the verbatim guard can't see it.
2. **Anchor over-reuse / nearest-mention anchoring** (successor of #2180's headers): the two single-word aluminum/copper mention spans are cited for the fixed-price claim, the escalating-tariffs claim, AND the no-quantification negative — none supported by a one-word span.
3. **Prompt-derived background still cite-decorated** (partial #2182): the event-overview sentence carries citations to loosely related entity mentions when its facts come from the task prompt.

Plus the cost dynamic: with the new discipline and NO retrieval hints, the agent exhausted a 3,000,000-token budget mid-hunt without finalizing (ResearchReport 7); given exact strings to search, it completed in 8 steps. Precision is currently paid for in search tokens.

## Suggested direction

- **Finalize-time claim-support check:** for each cite span, evaluate whether the anchor's surrounding context (say ±500 chars of the cited annotation) supports the sentence — cheap LLM entailment call or embedding-similarity floor. On failure: strip the citation (render as uncited analysis) or flag for one retry. This closes all three residuals with one mechanism, and generalizes #2193 from "is the quote verbatim" to "does the source say this".
- **First-class cite handles on retrieval results:** every passage returned by similarity/exact-text search carries a ready-to-use cite id for the span just read, so the agent cites what it retrieved instead of re-hunting for something citable. Kills the cost problem and nearest-mention anchoring together.

Evidence trail: fix-verification comments on #2180, #2182, #2183, #2189 (runs 4/6/7/8 on the Karman corpus).

JSv4:
Two post-#2192/#2193 regressions in the deep-research report pipeline, fixed with one composition/verification pass rather than two parallel ones.

## #2200 — the report rendered twice

`finalize()` post-processed only `markdown_body`, while the agent passed the whole report as `executive_summary` too. Result: the summary was emitted with its `<cite>` spans still raw (nothing ever cite-rendered it), followed by the agent's own stub `## Sources` line, then the same report again in footnote form, then the real Sources table. That "intermediate representation escaping into the final document" was two un-deduplicated copies plus an un-rendered one.

`finalize()` now composes **one** document (summary + body) and runs the citation post-processors over it **exactly once**, so a `<cite>` tag anywhere is rendered instead of leaking. Guards that close the doubling:

- `_sanitize_agent_markdown` strips agent-authored `## Executive Summary` / `## Sources` scaffolding (the system owns both headings and renders the footnote table itself) plus the hyperlinks the web-less agent invents.
- `_summary_duplicates_body` drops a summary that merely restates the body, probing the summary's opening against the body rather than diffing two multi-KB strings.
- `_verify_cite_spans` collapses a **self-quoting span** — the successor of #2183's sentence doubling, where a span's inner text is a verbatim copy of the sentence it decorates — to the new self-closing marker form, so the claim renders once with a trailing footnote.

`<cite ids="1,2"/>` is now a first-class cite form in `_CITE_SPAN_RE` / `_render_citations`, is what the system prompt asks the agent to prefer, and is what `_compose_salvage_body` emits. An empty wrapping span (`<cite ids="1"></cite>`) normalises to it, so both spellings run one code path.

## #2201 — nothing checked that a cited passage *supports* its sentence

`_verify_quoted_spans` becomes `_verify_cite_spans`: one walk over the cite spans enforcing three guards that share a single annotation query and a single `_contiguous_coverage` notion of "is this text a copy of that text":

1. **Echo collapse** (#2200, above)
2. **Quote verification** (#2189 — behaviour unchanged)
3. **Claim support** (new) — a cited sentence must share at least `RESEARCH_CLAIM_SUPPORT_MIN_COVERAGE` of its content words with the text of the annotation(s) it cites, or the citation is stripped (the prose survives as uncited analysis) and `report.warnings` records it.

A span's claim is its inner text, or — for a marker, and for a span collapsed by (1) — the text back to the previous span or sentence boundary, whichever is nearer. So in a compound sentence each anchor answers for its own clause, and a claim carries forward across consecutive markers rather than passing as a whitespace fragment.

The support check is a deterministic lexical floor: no LLM call, no embedding round-trip at finalize. It decisively kills two of the three residuals — a one-word mention span cited for a full sentence, and prompt-derived background decorated with a loosely-related entity anchor. Claims under `RESEARCH_CLAIM_SUPPORT_MIN_WORDS` words are accepted unchecked, since a fragment has too few content words for a ratio to mean anything.

A **polarity guard** sits behind the coverage check: word overlap is blind to negation, so "the tenant is NOT liable" scored like a faithful paraphrase of "the tenant is liable" — the highest-stakes misattribution in legal text. Above `RESEARCH_CLAIM_INVERSION_COVERAGE`, disagreement about a negation marker strips the citation. It covers prefixed negation ("non-cancelable") as well as particles, and skips the bare marker "no" when it carries an abbreviating period, since "No." also spells the citation-number abbreviation ("Exhibit No. 4").

**What it deliberately does not catch:** a well-anchored sentence carrying an invented tail (the over-attributed paraphrase). The prompt's new "one anchor must carry the WHOLE sentence" rule is the guard there, and `_claim_is_supported` is the single seam for swapping in an entailment call later.

## #2201 (cost) — citing was impossible, so the agent hunted

The #2180 prompt rule told the agent to pull a pinpoint anchor with `search_exact_text_as_sources` and cite it. That cannot work: the tool returns **synthetic negative** annotation ids, which `record_finding` rejects and `finalize` drops. The agent was steered at an id it could never obtain — the 3M-token run that never finalized.

`find_citable_passages` replaces it, backed by `AnnotationService.search_corpus_annotation_text`: the real corpus annotations containing a phrase, tightest (most pinpoint) anchor first, each row carrying a ready-to-paste `cite` handle, registered in the run's citation accumulator so `record_finding` accepts it. Visibility is delegated wholesale to `get_corpus_annotations` (corpus READ + per-document visibility + the analysis/extract privacy gate), so it adds no permission logic of its own. `search_exact_text_as_sources` is dropped from `DEEP_RESEARCH_READ_ONLY_TOOLS` — replaced, not paralleled.

## Nothing is dropped silently

Every strip in the pipeline is counted and surfaced in `report.warnings`: a stripped Sources/References section, a dropped duplicate summary, a lossy echo collapse, a demoted quote, a removed citation, and a report where *nothing* survives composition (which previously stored an empty document that looked like a successful run).

## Also

- Folds the three hand-rolled singular/plural warning branches into `_pluralize`.
- `_strip_scaffold_headings` is nesting-aware and follows CommonMark on both fence rules — a fence closes only on its own character *and* on a run at least as long as the opening. The two failure modes are quiet in opposite directions (scaffolding left unstripped, versus a section skip that swallows the rest of the document), so both are pinned.
- Heading matching is **exact**, not fuzzy: the set enumerates the variants a report generator reaches for (`Works Cited`, `Reference List`, singular `Citation`, `Endnotes`, …), because a token-overlap rule would delete substantive sections such as `Sources of Supply Risk` along with everything under them.
- The annotation-search default limit moves to `constants/annotations.py`, deliberately **not** coupled to the research tool's own ceiling.

## Tests

New coverage: duplicate-summary suppression, cite rendering inside the executive summary, echo collapse (including the lossy-collapse counter), self-closing and empty-wrapping marker forms, citation stripped/kept by the support check, split-marker carry-forward, clause scoping, negation inversion (particle and prefix), the `No.` carve-out, scaffold stripping (nesting, both fence rules, heading variants, and the substantive headings that must survive), and four cases for `search_corpus_annotation_text` (tightest-first ordering, document scoping + limit, empty/no-match, visibility delegation). `test_research_citation_graph.py` drives the real `_run_deep_research_async` loop with a stubbed agent to pin the closed-citation-graph contract end to end.

Several tests pin **known limitations** rather than desired behaviour, so a future change that "fixes" one side of a deliberate tradeoff is visible: the polarity guard's union across multi-anchor spans, its lexical-negation over-strip, the opening-anchored and ratio-over-summary edges of the duplicate-summary probe, and the `Inc.`/`U.S.C.` sentence-boundary truncation.

Two pre-existing test changes, both deliberate:

- `test_prompt_documents_citation_discipline` asserted the prompt names `search_exact_text_as_sources`. That assertion pinned the broken advice, so it now asserts `find_citable_passages` (and the absence of the old name).
- The `_verify_quoted_spans` call sites are mechanically renamed to `_verify_cite_spans` and star-unpack the added counters. No expectation changed — every existing quote-verification case still passes with the same values.

Closes #2200
Closes #2201


JSv4:
Thanks — both review rounds landed useful catches. Pushed `ddb26f3` (and `55aabb3` before it).

**Acted on:**

- **`finalize_report`'s tool docstring** (first review) — correct and the most consequential of the notes, since pydantic-ai surfaces that docstring to the model independently of the system prompt. It advertised only the wrapping form. Rewritten to cover the marker form, the wrapping form's narrow role, and the summary/body split. Grepped the rest of the source: remaining `<cite …>` mentions are internal docstrings that correctly describe both forms.
- **Split markers on one sentence** — real gap, fixed. `… <cite ids="1"/> <cite ids="2"/>` left the second span only the whitespace between the tags, so it passed the support check as a fragment. The claim now carries forward across spans, so a split citation can't slip an unsupported anchor through. Verified: on that input the second marker previously survived and now drops with the first (`cites_dropped` 1 → 2), with a regression test.
- **Blank `COMPLETED` report** — fixed. A body that reduces to `""` after `_sanitize_agent_markdown` now records a warning instead of storing an empty document that reads as a successful run.
- **mypy `attr-defined`** (the red `linter` job) — the new test class needed the class-level fixture annotations its sibling in the same file already declares.

**Not acted on, deliberately:**

- **Double warning when one span trips both the quote and support guards.** The stored content is right either way, and both rewrites genuinely happened — suppressing one would hide an edit from the reader. Redundant signal beats a silent change here.
- **`pg_trgm` index for `raw_text__icontains`.** Agreed it's not a regression (it mirrors what `search_exact_text_as_sources` did), and it needs a migration. I'd rather add it with evidence that `find_citable_passages` is actually a hot path than speculatively.
- **`_strip_scaffold_headings` stopping at any heading level** — documented behaviour, and the agent is now steered away from writing a Sources section at all.

One caveat on verification: I could not run the backend suite locally — the test image build exhausted this session's disk allowance mid-export. The new pure text logic (`_preceding_claim`, scaffold stripping, duplicate detection, claim support against every pre-existing fixture, echo collapse, the split-marker carry-forward) is verified by a standalone harness that execs the real functions; the DB-touching paths rest on CI.

---

JSv4:
Pushed `031b384`.

**Test gap closed.** You're right that `_citable_passage_rows` was inconsistent with `_compose_salvage_body` next to it. Added `CitablePassageRowsTestCase`: the permission/ordering semantics stay covered on `search_corpus_annotation_text`, so this pins the LLM-facing contract instead — a real positive annotation id, the paste-ready `<cite ids="N"/>` handle that `_CITE_SPAN_RE` actually consumes, `similarity_search`-shaped keys, and the preview/hit caps.

One correction on the closure's `deps.retrieved_annotation_ids.extend(...)` side effect: it stays untested, and I'd rather say so than paper over it. `find_citable_passages` is nested inside `_run_deep_research_async` and closes over `deps_ref`, so reaching it means either standing up the whole agent or extracting the closure — the second is a real refactor and I don't think this PR is the place for it.

**Correction on the previous round.** You're right that `Annotation.raw_text` already carries a pg_trgm GIN index (migration `0074_annotation_raw_text_trigram_index`) — I confirmed it. The earlier review raised the `icontains` scan as a perf risk and I said I'd defer adding an index pending evidence of a hot path. There was nothing to defer: the lookup is already index-backed. Recorded that in the method's docstring so the next reader doesn't re-litigate it.

**`limit=0` returns one row** — documented rather than changed. A `limit` of 0 meaning "give me nothing" is an odd call to make, and the current caller already clamps, so changing the floor risks more than it fixes. The docstring now states the floor and points at an empty `phrase` as the way to get no rows.

**`_strip_scaffold_headings` dropping a legitimately-titled section** — not acting, and the reasoning is worth stating: a warning here would fire on the *expected* case. An agent emitting its own `## Sources` stub is exactly the #2200 bug this strips, so the warning would be present on most reports and carry no signal. The catastrophic version — nothing survives composition at all — is what the new blank-content warning covers. A partial drop of a section the agent shouldn't have written is the intended trade, documented in the docstring.

Standing caveat, unchanged: I could not run the backend suite locally (disk allowance exhausted during the test image build), so the new DB-touching tests rest on CI.

---

JSv4:
Two of these were checkable rather than matters of taste, so I checked them. Pushed `5c086fe` (one comment, no behaviour change).

**1. `list[dict] | str` return — checked, keeping it, now documented.** It isn't a new shape for the agent: `PydanticAIToolWrapper` already returns a plain string from *any* tool on the operational-error path (`pydantic_ai_tools.py`, issue #820) regardless of the tool's annotation, and it doesn't schema-validate outputs the way it does inputs. So every tool in this codebase can already hand the model a string, and the union is just honest typing about a branch the framework produces universally. Keeping the miss branch as guidance rather than `[]`: "try a shorter fragment, or use similarity_search" is exactly what the model needs at that moment, and a bare empty list says nothing about what to do next — which is the whole cost problem #2201 is about. Recorded inline so it doesn't get "fixed" into an empty list later.

**3. Blank `COMPLETED` report — checked the frontend, it already reads correctly.** `ResearchReportDetail.tsx` renders an explicit `EmptyState` for empty `report.content` ("No report body / This run finished without a rendered report"), with the warnings rendering as `WarningChip`s above the tabs — so the new blank-composition warning shows up right next to an empty state that already says what happened. That's the sensible treatment you asked me to confirm, and it's also the argument for staying `COMPLETED`: the run *did* complete its loop and produced citations and a tool log, and `mark_failed` would discard that framing behind an `error_message` for something that wasn't an error.

**2. `_strip_scaffold_headings` silent drop — holding, and noting this is the second round it's come up.** The reason a warning is wrong here specifically: the case that triggers it is the agent writing its own `## Sources` stub, which *is* the #2200 bug this strips. A warning would therefore fire on the common, expected path and carry no signal — the opposite of the quote/claim-support warnings, which fire on genuinely anomalous input. Distinguishing "stub I should drop silently" from "real section I shouldn't have dropped" needs a content-substantiality heuristic, which is more machinery than the case earns given the prompt forbids these headings outright. If you'd rather have the warning anyway, it's a small change (thread a count out of `_sanitize_agent_markdown` into the existing `extra_warnings` path) — say the word and I'll add it.

Standing caveat: the backend suite still hasn't run locally (disk allowance exhausted during the test image build), so the DB-touching tests rest on CI.

---

JSv4:
Pushed `c58e2c6` — fixes the red `linter` job and records the split-marker asymmetry you flagged.

**CI (`linter` on `031b384`).** mypy `valid-type`: this file binds `User = get_user_model()`, a *variable*, which can't be used in the class-level fixture annotations the new `CitablePassageRowsTestCase` needs. Imported the concrete model directly — the sibling `test_corpus_annotations_query` already does exactly that — so every existing `User.objects.create_user` call site is untouched.

**The quote-verification gap across split markers — real, and I checked which way it cuts before writing it down.** You're right that step 2 only ever inspects `preceding`, which is the blank gap for the second marker, so quotes are verified against the *first* marker's anchors alone. But the consequence is the opposite of "probably fine, might let something through": it can only **over-strip**, never under-strip.

Traced it against the real functions:

- quote grounded in anchor 2 but not anchor 1 → demoted at the first marker (**wrong, over-strict**)
- quote in the combined `<cite ids="1,2"/>` form → kept, as it should be
- quote in **neither** anchor → still demoted at the first marker

So a fabricated quote cannot reach the reader through the split form; the worst case is a correct quote degrading to paraphrase with a warning that tells the reader to check the wording. That's the safe direction, and it only arises when the agent ignores the prompt's instruction to combine ids.

Recorded in `_verify_cite_spans`'s docstring with that direction spelled out, plus the fix that would close it (union candidates across a run of adjacent markers) and why I'm not building it: it needs a lookahead pre-pass for a shape the prompt forbids, and the current behaviour already fails safe. Happy to add it if you'd rather not carry the strictness.

Your second bullet — an empty carried-forward claim on the very first span falling into the short-claim short-circuit — is right and is the intended behaviour; `last_claim` starts empty and there's genuinely nothing to check against.

Caveat unchanged: no local backend suite run (disk allowance exhausted during the test image build), so the DB-touching tests rest on CI. Watching for the re-run.

---

JSv4:
This is the best catch of the six rounds — the negation blind spot is real and I've fixed it rather than documented it. Pushed `299faa0`. I reproduced both blind spots against the live functions first:

```
faithful  ("the tenant is liable for repairs…")      supported: True
INVERTED  ("the tenant is NOT liable for repairs…")  supported: True   ← opposite meaning, cited anyway
fabricated figure ("10 days" vs anchor "30 days")    supported: True
'10' survives tokenisation:                          False
```

**Negation — fixed with parity, because your first suggestion wouldn't have worked.** Un-stopwording `"not"` only moves coverage from ~1.0 to **0.857**, still far above the 0.25 floor, so the inversion would still sail through. Your alternative — parity — is the mechanism that catches it. `_claim_is_supported` now rejects a claim that disagrees with its anchor on the presence of a negation marker, gated behind `RESEARCH_CLAIM_INVERSION_COVERAGE` (0.8).

That gate is doing real work, and it's the reason I could make this aggressive: legal text frequently negates *lexically* ("prohibited", "except", "unless") rather than with a marker, and such a paraphrase shares far fewer words with its anchor, so it never reaches the gate. Verified both directions:

- `"tenant is NOT liable…"` vs anchor `"tenant is liable…"` → **stripped**
- `"Subletting … are prohibited absent prior written consent"` vs anchor `"Tenant shall not sublet … without the prior written consent"` → **kept** (coverage 0.33, nowhere near the gate)
- both-negated pair → parity holds, kept

Checked on the normalized token stream, not `_content_words`, so the stopword list can't hide a marker. Both directions have dedicated tests.

**Numeric tokens — exempted from the length floor**, so `"10"` and `"5%"` now enter the ratio from both sides. Being precise about what this buys: it restores *signal*, it does not catch a fabricated figure on its own, since one differing number moves the ratio by 1/N. I deliberately did **not** add numeric parity — legal text writes the same amount as "30", "thirty (30)", "$5" and "$5,000,000", so parity would strip correct citations far more often than it caught invented ones. That's now recorded next to the existing "what this doesn't catch" caveat.

**`_strip_scaffold_headings` — reversing my earlier decline.** You're the third reviewer to raise this, which prompted me to re-examine my reasoning, and it was weak. I argued a warning would fire on the expected case — but the prompt now forbids the agent from authoring those headings, so it should be *rare*, which makes it informative rather than noise, and doubles as the signal for whether that prompt rule is actually landing. It was also the one guard with no observability while having the largest blast radius. `finalize` now counts and warns.

Caveat unchanged: no local backend suite run, so the DB-touching tests rest on CI.

---

JSv4:
Pushed `cd3bbe5`. The migration nit was the valuable one — it wasn't intentional, and finding it also surfaced *why* it kept coming back.

**Migration reformat — reverted, with the root cause.** I'd reverted that hunk early in this PR, but a later `black opencontractserver/research` run silently reintroduced it. `.pre-commit-config.yaml` excludes `/migrations/` globally, so CI never touches them — but a bare `black <dir>` doesn't honour a pre-commit exclusion, so my directory-wide invocation undid the revert without any signal. Reverted against `main`, and I've switched to naming the touched files explicitly. Good catch; it would have shipped as unexplained diff noise in a migration.

**Direct test for the no-anchor-text branch — added.** You're right it was only covered transitively through the quote-verification path. `_claim_is_supported` is now pinned directly for both the empty-candidate-list and empty-string-candidate cases, plus the short-claim short-circuit that runs ahead of it.

**Empty-claim marker — documented inline rather than guarded, and I want to be precise about why.** A marker with no prose before it produces an empty claim that falls under the min-words floor and is left alone. That's correct rather than merely well-calibrated: there is no sentence for the anchor to misrepresent, so there is nothing to strip. Adding a `not claim.strip()` early-continue would *drop* the citation and increment `cites_dropped`, emitting a warning about a degenerate input that carries no misattribution risk. The comment now says this at the call site.

Thanks for independently re-deriving the MIN-semantics question on `get_corpus_annotations` and the `select_related` caching of `report.creator` — those were both deliberate and it's genuinely useful to have them confirmed by someone who went and read the source rather than taking the docstring's word for it.

Caveat unchanged: no local backend suite run, so the DB-touching tests rest on CI.

---

JSv4:
Pushed `2228b2f`. One of your two flagged items was a real hole in the guard I added two rounds ago; the other doesn't hold up when checked against the source.

**Prefix negation — fixed, this was a genuine gap.** You're right that the token-exact match misses "non-cancelable", and that's not a corner case: it left exactly the inversion class the polarity guard exists to catch. Confirmed before fixing:

```
negation detected in 'shall not be cancelled': True
negation detected in 'is non-cancelable':      False
inversion via 'non-' prefix supported:         True   ← should be False
```

`_is_negated` now matches a negating prefix as well as a particle. Kept it hyphenated and limited to `non-` on purpose — bare `un`/`in` prefixes would fire on "under", "until" and "interest", and an unhyphenated `non` would catch "none"/"nonetheless". Verified all five stay non-negated, with tests for the inversion, the lookalikes, and both-prefixed parity.

**`similarity_search` permission divergence — I don't think the premise holds.** I read `CoreAnnotationVectorStore._apply_filters` rather than taking the shape at face value, and the corpus-scoped path is:

```python
active_filters &= Q(structural=True, structural_set_id__in=visible_corpus_set_ids) \
                | Q(structural=False, corpus_id=self.corpus_id)
```

where `visible_corpus_set_ids` derives from `Document.objects.visible_to_user(user)`. So **structural rows do go through per-document visibility** — there's even a comment recording that bypassing it was a past bug. And non-structural rows are then narrowed by the visibility filter to `creator_id == self.user_id`, i.e. the requesting user's own annotations — *stricter* than `get_corpus_annotations`' MIN semantics, not looser.

So the two tools do differ in shape, but not in the direction that would leak a private document's content, which was the concern worth acting on. The one real divergence I found is that the vector store doesn't apply `apply_source_privacy_gate` — largely subsumed by the creator-only rule on non-structural rows, since another user's analysis-created annotations are already excluded. I'm not filing a follow-up asserting a problem I've just shown doesn't exist in the claimed direction; flagging the analysis here so a maintainer can decide whether the shape difference alone is worth normalising.

The `## References` case is now observable via the `sections_stripped` warning added last round, and the thin-closure coverage gap I acknowledged earlier still stands as stated.

Caveat unchanged: no local backend suite run, so the DB-touching tests rest on CI.

---

JSv4:
You're the third reviewer to flag this gap, and your framing is what moved me: the accumulator registration *is* the mechanism the cost fix depends on, so leaving it untested was leaving the point of #2201 unpinned. Pushed `ce23ae6`.

I'd previously argued the closure was awkward to reach without a refactor. That was wrong — the package already has the pattern (`AstartDeepResearchTestCase` uses `TransactionTestCase` + `asyncio.run`), and stubbing the agent factory is enough. No LLM involved: the stub's `chat` replays the three tool calls the contract spans, and the test follows one annotation id across every hop.

`CitationGraphContractTestCase` asserts, in order:

1. `find_citable_passages` returns the real annotation with a paste-ready `<cite ids="N"/>` handle
2. that id lands in `deps.retrieved_annotation_ids`
3. `record_finding` **accepts** it — rather than rejecting it as an id no retrieval produced, which is exactly what `search_exact_text_as_sources` used to guarantee
4. `finalize` keeps the footnote instead of dropping the citation, with `report.citations` resolving to that annotation

A second case pins the miss branch: guidance string returned, nothing registered.

Step 3 is the one that would have caught the original #2201 bug, and it's now a regression guard rather than a claim in a PR description.

On your other two notes: the unhyphenated `noncancelable` spelling does still slip the polarity guard — real residual, and I'm leaving it rather than widening the prefix match, since dropping the hyphen requirement is what would start firing on "none"/"nonetheless". The multi-anchor union semantics in the inversion check are deliberate and match how the coverage check already treats multi-id spans; making one per-candidate but not the other would be the inconsistency. Both are now the kind of thing the constants file documents.

Caveat unchanged, and it applies to this new test too: no local backend suite run, so it rests on CI.

---

JSv4:
Your note 2 was a real leak path, not just a residual — fixed in `3add156`. Reproduced it first:

```
## Sources
(All claims above are cited inline.)   ← stripped
### By Document
- Lease.pdf, annotation 12             ← LEAKED into the report
```

The skip ended at the next heading of *any* level, so a subsection nested inside the scaffolding survived it. I'd been treating "up to the next heading of any level" as documented-therefore-intended, but an orphaned subheading and its list escaping into the report is the same class of scaffolding escape #2200 exists to close — the design was incomplete for nested input, not deliberately permissive about it.

The skip now tracks the scaffolding heading's level: a deeper heading is inside the section and keeps it running; one at the same or shallower level closes it, exactly as the markdown reads. Verified the three neighbouring behaviours are unchanged rather than assuming:

- a sibling section *after* the scaffolding (`## Appendix`) survives
- a deeper scaffolding heading (`### Sources` under `## Findings`) takes only its own subtree, leaving the later `## Analysis` intact
- the executive-summary heading still drops the line alone, keeping its prose

Note 1 (`limit=0` floors to 1) is the same one raised two rounds ago; I documented it then and the only caller still clamps ahead of it, so no change. Note 3 (split-marker quote asymmetry) — thanks for independently confirming the over-strip-only direction holds; that matches what I traced when it first came up.

Caveat unchanged: no local backend suite run, so all of this rests on CI.

---

JSv4:
Both fixed in `4e09340`. The `"No."` one was a real false positive, and worse than the tail case you described — I reproduced it end to end:

```
'the premises described in Exhibit No. 4'   negated=True   ← wrong
'Case No. 12-3456 governs'                  negated=True   ← wrong

claim:  "...maintain the premises described in Exhibit No. 4 in good repair..."
anchor: "...maintain the premises described in the attached exhibit in good repair..."
valid citation survives?  False             ← stripped a correct citation
```

`_is_negated` strips punctuation before matching, so `No.` normalises straight onto the bare `no` marker. Exhibit/item/case numbers are ordinary vocabulary in the contracts this targets, so this wasn't hypothetical — it's a guard designed to protect citation integrity actively destroying a correct citation.

`"no"` now counts as negation only when a number doesn't follow it. That keeps "no liability" / "no obligation" working without reopening the `non-cancelable` case you'd expect me to have weakened. Verified all four still behave: the two reference forms are clean, the two genuine negations still fire, and a real polarity inversion is still rejected.

Your instinct to compare it against the care spent on the `non-` prefix was the right lens — same guard, same class of tokenisation subtlety, and I'd only looked at one side of it.

**Duplicate `own`** — trimmed. While there I noticed the token-edge punctuation string was itself duplicated between `_content_words` and `_is_negated`; extracted to `_TOKEN_PUNCTUATION` so both agree on where a token ends, which is the sort of drift that would have made a future fix to one silently miss the other.

Caveat unchanged: no local backend suite run, so this rests on CI.

---

JSv4:
All three addressed in `337df82` — though item 2 is implemented via a different mechanism than you suggested, for a reason the codebase already documents.

**1. Lettered references — right, and the fix exposed a second bug in my own patch.** The digit lookahead I added last round was the wrong discriminator. Switched to the abbreviating period on the token itself: bare `no` negates, `no.` abbreviates. That covers lettered and bracketed references *and* repairs a false negative the lookahead had introduced:

```
[ok] negated=False  'Schedule No. A-1 lists the equipment'
[ok] negated=False  'see (No. 7) below'
[ok] negated=True   'there is no 30-day cure period'   ← the lookahead rule got this WRONG
```

So the period rule is strictly better than what it replaces, not just broader. Ten cases pinned.

**2. Headers as citable passages — agreed on the goal, but `structural=False` would have made it worse.** `research/constants.py` records the #2180 finding: the parsing pipeline marks its *entire* layout layer `structural=True` — body paragraphs, tables, sentence chunks — while the bookmark-derived `OC_SECTION` headers that caused #2180 are explicitly `structural=False`. Filtering on that flag would have dropped nearly every citable passage while keeping exactly the headers you want gone.

Implemented on the label instead, which is the precise signal and the same one the finalize-time lint uses: new `exclude_label_texts` argument, with `RESEARCH_HEADER_ANCHOR_LABELS` passed from the call site. Test covers both directions — the header is returned unfiltered, absent when excluded, and the real passages still come back.

**3. `document_id` documented** — you're right that this is the description the model reads, and it's the same point I acted on for `finalize_report` earlier in this PR. It now tells the agent to scope to a document it has already identified, and that headers are never returned.

Your nit about `"not"` being in the stopword list: confirmed deliberate, and thank you for verifying rather than assuming — the polarity guard reads the normalized text directly precisely so the stopword list can't blind it.

Caveat unchanged: no local backend suite run, so this rests on CI.

---

JSv4:
The code-fence gap was real, and the damage is worse than "could trigger a section-skip" — fixed in `c0212c9`. Reproduced first:

````
## Findings
The filing includes this extract:
```
# Sources            ← read as a heading, starts skipping
revenue = 1_200_000  ← dropped
```                  ← dropped, fence left unterminated
Further analysis follows here.   ← DROPPED
````

So it's silent content loss *plus* broken markdown, not just a spurious skip. Heading detection is now suspended between fence delimiters (backtick and tilde forms). Verified a real heading after the fence closes still strips normally, the fenced comment survives as content, and the fence stays balanced.

This is the same family as the nesting gap from two rounds ago: `_strip_scaffold_headings` reads markdown structure, and it needed to actually understand that structure rather than pattern-match line prefixes. Two structural cases in three rounds suggests that was the right generalisation to make.

**Not acting on the other two, both with reasons:**

- **`limit=0` floor** — third round this has come up. It's documented, and you're right that the double clamp is defence-in-depth rather than a bug. Changing the floor now would alter a service contract to fix something no caller can reach.
- **pg_trgm ineffective below 3 characters** — accurate, and I considered a minimum-phrase guard. Declining because `find_citable_passages` is agent-driven against language the agent has already read, so a 1–2 character phrase is malformed input rather than a realistic query, and you called the ceiling acceptable yourself. If it ever shows up in practice the fix is a min-length check in `search_corpus_annotation_text`, aligned to the trigram size.

Caveat unchanged: no local backend suite run, so this rests on CI.

---

**Comment by JSv4** [comment]
First full `pytest` run landed, and it caught a real bug — **in my test, not the production code**. Fixed in `575a2f7`.

**Result: 10428 passed, 2 failed.** Both failures were `test_research_citation_graph.py`, the test I added in round 9. Nothing else regressed, so the pipeline changes across all thirteen rounds are clean under the real suite.

**Root cause**, from the traceback:

```
[PydanticAI sync chat] Starting chat with message: 'Execute the research task...'
HTTP Request: POST https://api.openai.com/v1/chat/completions "HTTP/1.1 401 Unauthorized"
```

The stub never ran. `_run_deep_research_async` does `from opencontractserver.llms import agents`, and `llms/__init__.py:44` re-exports `llms.api.agents` — the `AgentAPI()` singleton from `api.py:758` — which **shadows the `llms.agents` package of the same name**. I patched `"opencontractserver.llms.agents.for_corpus"`, i.e. the package function nobody calls. The real factory stayed in place, built a real agent, and pointed the run at a live LLM. With no key configured that surfaced as a 401, the script never executed, and the assertions failed on a missing dict key rather than on anything meaningful.

Now patches the singleton via `patch.object(llms_api.agents, "for_corpus", ...)`.

**Added a guard so this can't pass silently again:** `_drive` asserts the stub actually produced an agent. Without it, a mis-targeted patch degrades into a real network call that fails somewhere unrelated — or, on a machine with credentials present, *passes for entirely the wrong reason*. That second case is what worried me most: the test would have been green and meaningless.

This is precisely what the "no local backend suite run" caveat I repeated each round was covering. The pure-logic harness I've been leaning on execs the real functions, so it validates the citation logic — but it cannot see a mis-targeted patch, because the patch target only matters inside the Django and pydantic-ai wiring the harness deliberately skips. The caveat was load-bearing, not boilerplate.

---

**Comment by JSv4** [comment]
Pushed `a7af8ff`. One note recorded, two verified as harmless rather than taken on faith.

**1. Multi-anchor polarity masking — real, now documented.** Reproduced it:

```
single anchor  (the one the claim inverts)  -> rejected  (correct)
two anchors, one of them unrelated-negated  -> passes    (masked)
```

`any()` sees a negated anchor, so parity is satisfied whichever way the claim reads. Added to the known-limitations list alongside the invented tail and the fabricated figure. I'm not making polarity per-candidate: coverage is a union too, and splitting them would be the inconsistency rather than the fix. The honest fix is the entailment call `_claim_is_supported` is explicitly the seam for.

**2. `_MD_HEADING_RE` trailing `#` — verified it cannot manifest, leaving it.** You're right that `## What is C#` normalises to `what is c`, deviating from CommonMark. But I checked all seven scaffold names and none ends in `#`, the normalized title is only ever compared against that set, and the heading line is preserved verbatim in the output — confirmed directly:

```
'## What is C#'  -> normalized title 'what is c'   (never matches a scaffold name)
C# heading preserved verbatim: True | sections stripped: 0
```

Tightening the regex would touch the closing-sequence handling that *does* matter, for no observable gain. Your framing — "in case the heading-matching logic is ever reused" — is the right caveat, and the docstring already scopes this to scaffold comparison.

**3. Bare `"non"` — keeping it, and it isn't quite dead weight.** PDF text extraction can break a hyphenated `non-cancelable` across a line and yield a standalone `non`, so the token entry still earns its place next to the prefix rule. Noted the reasoning in the commit.

**Status:** the `pytest` failure from the previous run was my test patching the wrong `agents` object (fixed in `575a2f7`, details in the comment above). CI is re-running on this head; I'll confirm green.

---

**Comment by JSv4** [comment]
Your note 1 was framed as a perf observation ("not worth changing for the current call site") — but checking it surfaced a latent **correctness** bug I introduced in round 12. Fixed in `0813035`.

`Annotation.annotation_label` is `null=True`. A negated lookup across a nullable FK drops the NULL rows too, because SQL is three-valued: `NOT (NULL = 'x')` is NULL, not TRUE. So excluding the header labels could silently take **every unlabelled annotation** with it — quietly shrinking what `find_citable_passages` can offer, with no error and nothing in the warnings.

Now one combined `Q` that re-admits unlabelled rows explicitly:

```python
qs.filter(~excluded | Q(annotation_label__isnull=True))
```

That's correct regardless of how Django compiles the negation, and it collapses the chain of `exclude()` calls into a single condition — so your perf point is addressed by the same change. Regression test asserts an annotation with `annotation_label=None` survives an exclusion naming three header labels.

Worth saying plainly: I'd have skipped this on your framing alone. It was reading the field definition to check whether the loop was *worth* collapsing that turned up the nullable FK. That's the second time in this PR a "minor/non-blocking" note has led somewhere real.

**Not changed, both deliberate:**
- The 400-char probe in `_summary_duplicates_body` — documented tradeoff, and the observed failure mode is the agent passing the identical string twice, which the probe catches from its first sentence.
- The sentence-final `"no."` heuristic in `_is_negated` — already documented, and as you note it fails open: a missed inversion falls back to coverage-based support checking rather than passing straight through.

---

JSv4:
You asked for a test pinning the sentence-boundary behaviour. Writing it surfaced a second case in the same regex — pushed `76b179a`.

The whitespace requirement turns out to be load-bearing in the *right* direction: it's what stops `the cap is 1.5 million` splitting at the decimal. But nothing stopped a legal reference splitting:

```
"It governs Exhibit No. 4 for the term"  →  claim seen as "4 for the term"
```

The claim handed to the support check was truncated at the abbreviation, short enough to fall under `RESEARCH_CLAIM_SUPPORT_MIN_WORDS` and skip the check entirely — the guard quietly declining to examine a sentence rather than examining it. It fails open, so less harmful than the false-positive class, but it's the reference vocabulary this domain is full of, and `_is_negated` already encodes the same "`No.` is not a sentence end" insight one function away.

Added a `(?<!\b[Nn][Oo])` lookbehind. Spelled out per-case because this regex, unlike the negation check, runs on **raw** text rather than the casefolded stream — my case-sensitive first attempt silently did nothing, which the harness caught before it reached CI. The word boundary keeps it to a standalone `no`, so `casino.` still ends a sentence.

The test pins both constraints and, more importantly, the four cases that must **not** change: decimal intact, reference intact (upper and lower case), ordinary sentence ends still splitting, and a word merely ending in "no" still splitting.

**On the other notes:** `limit=0` is the fourth mention — documented, no caller reaches it. The polarity guard firing only at ≥0.8 coverage is deliberate; the constants explain that gate is exactly what keeps it off honest paraphrase that negates lexically. And the `exclude_label_texts` chaining nit is already resolved — it became a single combined `Q` in `0813035` last round, which is also where checking it turned up a nullable-FK bug that would have silently dropped every unlabelled annotation.

---

JSv4:
Your main concern was fixed in `76b179a` last round — and you independently suggested the exact discriminator that fix used (reuse `_is_negated`'s `No.` handling). Confirmed against the current head:

```
"Exhibit No. 4 in good repair"       -> intact          (fixed)
"Karman Holdings Inc. reported ..."  -> "reported ..."   (still truncates)
"12 U.S.C. 1701 for the term"        -> "1701 for ..."   (still truncates)
```

**I'm not extending the carve-out to `Inc.`/`Corp.`/`U.S.C.`, and the reason is that the two errors aren't symmetric.** `No.` is unambiguous — a reference identifier always follows, so it is never a sentence end, and carving it out cannot merge real sentences. `Inc.` genuinely ends sentences in filing prose ("...acquired by Karman Holdings Inc. The transaction closed..."), so suppressing that boundary would **merge** two sentences into one claim.

Truncation shortens the claim and fails open — the check is skipped or scores higher. Merging pads the claim with unrelated vocabulary and pushes coverage toward a false strip, which is the harmful direction.

One correction to my own reasoning: I expected to demonstrate a merge flipping a valid citation, and **it didn't**. Measured erosion was 1.00 → 0.40 for one unrelated preceding sentence — that's most of the headroom above the 0.25 floor, but it stayed supported. So this is a margin argument, not a demonstrated false strip, and I've written it up that way rather than overstating it.

Both the truncation *and* the boundary a merge would destroy are now pinned by test, so the tradeoff is visible to whoever revisits this with real sentence segmentation.

**`limit=0`** — you're the one who framed this in the way that made it actionable: an *LLM* could pass `limit=0` expecting nothing. Previous rounds raised it as an API-contract nit and I declined it four times, correctly I think, since no code caller reaches it. But `find_citable_passages`' docstring is what pydantic-ai hands the model, so the caveat now lives there.

**Echo-collapse substring short-circuit** — accepted without change. It collapses a fragment-scoping span to a bare marker, losing the scoping the prompt asks for, but the footnote still lands on the right sentence. Cosmetic, not misattribution.

---

JSv4:
Fixed the duplicated clamp in `8bb204c`. You're right that it's harmless today — but two numbers that must stay in step with nothing keeping them there is exactly the silent-divergence shape the repo's DRY rule exists to prevent, so it's worth resolving rather than noting.

Split by ownership rather than deleting one arbitrarily:

- the **ceiling** (`RESEARCH_CITABLE_PASSAGE_MAX_HITS`) is research-specific and stays at the research call site;
- the **floor** of 1 belongs to `search_corpus_annotation_text`'s contract, is already documented in its docstring, and now lives only there.

Verified the composition is unchanged for every input the tool can receive, rather than assuming the refactor was behaviour-preserving:

```
limit=None  -> 1    limit=1   -> 1
limit=0     -> 1    limit=7   -> 7
limit=-5    -> 1    limit=999 -> 10
```

Your other two notes need no action, and both are already disclosed in the code rather than left implicit — the polarity guard's union treatment of multiple anchors (in `_claim_is_supported`'s known-limitations block) and the split-marker quote asymmetry (in `_verify_cite_spans`, with the over-strip-only reasoning). Agreed the split form is worth watching under real traffic: the prompt asks for the combined `<cite ids="1,2"/>` but doesn't forbid splitting, and the `sections_stripped`-style warnings would be the natural place to surface it if it turns out to be common.

---

JSv4:
Both open review notes are now settled by tests rather than prose (`d7ae737`). No behavior change.

**1. Union polarity — pinned.** This was documented but untested, which is the gap the review named. `_claim_is_supported` measures coverage over the *union* of the cited anchors and the polarity guard follows suit, asking whether **any** anchor carries a negation marker. So a span citing two anchors that disagree on polarity satisfies parity whichever way the claim reads:

| cited anchors | claim | result |
|---|---|---|
| affirming only | inverted | rejected ✅ |
| affirming + an unrelated negated anchor | inverted | **survives** ⚠️ |

The test asserts *both* rows, so the limitation cannot quietly widen into the control case.

**2. Compound sentences — the observation was right, the mechanism wasn't.** The review predicted that `the tenant is liable for repairs and is not responsible for painting` would trip the guard against an anchor supporting only the first half. Measured against the real function, it doesn't:

```
claim words=12  coverage=0.60  claim_neg=True  anchor_neg=[False]  ->  supported=True
```

Coverage lands at 0.60, well under the `RESEARCH_CLAIM_INVERSION_COVERAGE` gate of 0.80 that the guard sits behind — it never fires. The extra half-sentence that makes the claim compound is exactly what drags coverage below the gate.

The real over-strip is a different shape: an anchor that negates **lexically**. `Tenant obligations, excluding painting the interior walls…` reads as affirmative to a marker lexicon, so a faithful restatement using "not" scores 0.86 and looks like an inversion:

```
coverage=0.86  claim_neg=True  anchor_neg=[False]  ->  supported=False   # valid citation lost
```

That one is now pinned too, and it's the case the docstring describes.

Both failures are **one-directional** — an over-strip, never a fabricated attribution — which is the right way round for a guard whose job is protecting attribution, and both point at the same fix: the entailment call this function is the seam for. Widening the negation lexicon only moves the boundary.

Also took the `exclude_label_texts: Optional[Any]` → `Optional[Iterable[str]]` nit.

**CI:** everything green on the previous head except `pytest`, which was still running; the earlier `pytest` failure notice was for `c0212c9`, confirmed an ancestor of the `575a2f7` fix (`git merge-base --is-ancestor` → yes), so it's stale. Standing caveat unchanged: the backend suite has never run locally in this session (disk allowance), so CI remains the authority — as it was for the shadowed-import bug my local harness structurally could not see.

---

JSv4:
Two new observations from the latest review, both verified against the real code and addressed in `44cf3ca`. No behavior change.

**1. One span tripping two guards — intended, now pinned.** The review asked whether `quotes_demoted` and `cites_dropped` both incrementing for one span is the intended UX or an artifact of running the guards in sequence. It's reachable, and it's intended. Reproduced:

```
<cite ids="N">"the tenant waives all consequential damages" under the negotiated
              indemnity schedule attached hereto</cite>
   against anchor: "The lease term commences on the first day of January…"

quotes_demoted = 1   cites_dropped = 1
markdown = 'the tenant waives all consequential damages under the negotiated indemnity schedule attached hereto'
```

The reason it isn't double counting is that **both edits land in the text the reader sees**: the quotation marks come off (so no fabricated verbatim survives) *and* the footnote comes off (so the sentence isn't attributed). Warning about only one would leave the other edit unexplained. The test asserts both counters and both edits, so the pairing can't drift.

**2. Structural rows and `document_id=None` — good catch on the docstring.** `_citable_passage_rows` already guards it; what was missing was *why* it can be null. A structural annotation is shared across its `structural_set` rather than owned by one document, so it has no single document to name. Such a row still carries a usable `annotation_id`/`cite` handle, so it's returned with an empty `document_title` rather than dropped. Rare on this path — the header labels most structural rows carry are excluded — but the row builder must not assume a document is present, and the docstring now says so.

Everything else in the three reviews is already settled: union polarity and lexical-negation over-strip were pinned in `d7ae737`, the double floor-clamp on `limit` was collapsed in `8bb204c` (the ceiling stays at the research call site, the floor lives in the service contract), `Optional[Iterable[str]]` is in, and the removed quote-glyph bail-out is a deliberate consequence of every span now needing the echo and claim-support checks regardless of quoting.

On the shared point about a zero-content report still finalizing as `COMPLETED`: that's deliberate and tested, but it is a product call rather than a code one — happy to flip it to `FAILED` if that's preferred.

**CI:** the `pytest` failure notice for `c0212c9` is stale — confirmed an ancestor of the `575a2f7` fix. Everything else was green on the previous head. As always, the backend suite has never run locally in this session, so CI is the authority.

---

JSv4:
Both new findings are real and are fixed in `f429735`.

**1. Partial-echo collapse — right problem, and measuring it changed the fix.**

`_contiguous_coverage` is a ratio over the **inner** text, so any tail the span adds shrinks it proportionally. Against the sentence `The tenant is liable for all structural repairs to the roof and exterior walls of the premises under Section 8 of the lease`:

| inner | coverage | outcome |
|---|---|---|
| exact restatement | **1.000** | collapses, loses nothing |
| … only punctuation differs | **1.000** | normalization already handles it |
| … `+ " as amended"` | 0.918 | collapses, **loses the tail** |
| … `+ ", which the parties renegotiated in 2019"` | 0.658 | below gate — **left intact** |

So the band where text is actually lost is a tail of roughly one word. That's too narrow to justify either of the suggested fixes: raising to 0.98 would let visible near-duplicate restatements through, which is the #2200 symptom the guard exists for; extracting and re-emitting the remainder needs a match-position helper `_contiguous_coverage` doesn't expose, and would emit a mid-sentence fragment.

What *does* hold regardless of band width is your framing — it was the one strip here with no counter and no warning. So the fix is to stop it being silent: a new `echoes_trimmed` count and warning, **gated on `coverage < 1.0`**. Exact echoes (the observed shape) report nothing, so no warning lands on a normal report; only a genuinely lossy collapse is surfaced. All three boundary cases above are pinned by test. Existing tuple unpacks of the verifier now use `*_` so a future signal won't churn them again.

**2. Fence-character parity — confirmed, and it fails quiet.**

You're right, and the failure is worse than "under-fires" — it's sticky. With toggle-on-any-fence, a stray `~~~` inside a backtick block flips state twice:

```
```        -> in_fence = True
# Sources  -> (suspended, correct)
~~~        -> in_fence = False   ← should be content
```        -> in_fence = True    ← now stuck for the rest of the document
## Sources -> not stripped
```

Every heading after that point escapes detection. Now tracks the opening character and closes only on a match, with a test asserting the fenced `# Sources` survives as content *and* the real `## Sources` section after the true close is still stripped.

The rest of that review needs no action — the `search_exact_text_as_sources` observation is right (still registered for other agent contexts, so nothing dead), and the removed quote-glyph bail-out is a deliberate consequence of every span now needing the echo and claim-support checks.

Also, the three re-delivered earlier reviews are all settled: union polarity + lexical-negation over-strip pinned in `d7ae737`, the two-guards-one-span warning pair pinned and the structural-`document_id` docstring note added in `44cf3ca`, the double floor-clamp collapsed in `8bb204c`, `Optional[Iterable[str]]` in. The `pytest` notice for `c0212c9` remains stale (ancestor of the `575a2f7` fix).

The only open item is a product call rather than a code one: a zero-content report finalizing as `COMPLETED` with a warning, flagged independently by three reviews now. Happy to flip it to `FAILED` if that's preferred.

---

JSv4:
The unbounded-inner-text point is right and is fixed in `db65d77` — though not by the proposed mechanism, and the asymmetry it points at turns out to be deliberate.

**The cost is real.** Measured against a 1200-char `preceding`:

| inner length | `_contiguous_coverage` |
|---|---|
| 1 KB | 8 ms |
| 10 KB | 80 ms |
| 100 KB | 830 ms |
| 500 KB | 4,113 ms |

Linear, per span, and the inner group is the one input nothing caps — `_preceding_claim` caps its side at 1200, quote extraction caps its own at 2000.

**But capping `_CITE_SPAN_RE` would be a regression.** An unmatched *quote* pattern means "leave this text alone," which is safe — that's why `_QUOTED_PASSAGE_RE` can be bounded. An unmatched *cite* span is neither verified nor rendered, so a raw `<cite ...>` tag leaks into the stored report: precisely the #2200 symptom this PR removes. Bounding a detection regex isn't the same move as bounding an extraction regex.

**The work is provably useless past a length, so it can be skipped exactly.** Coverage divides the longest matching block by `len(inner)`, and that block cannot exceed `len(preceding)`:

```
coverage  ≤  len(preceding) / len(inner)
```

So once `len(inner) * threshold > len(preceding)`, the threshold is arithmetically out of reach. Skipping there is exact, not heuristic — same result, simply not computed.

Verified three ways: the bound held over 3,000 randomized `preceding`/`inner` pairs (deliberately including echo-shaped ones) with zero violations; the three echo boundary cases from `f429735` are unchanged; and the 500 KB span now takes **28 ms instead of 4,113 ms**. Pinned by a test whose span *starts* as a perfect echo, so only its length can rule it out.

The other two notes need no code change:

- **`_summary_duplicates_body` is opening-anchored** — accurate, and already documented as the deliberate choice to stay linear rather than diff two multi-KB strings. Worth knowing it isn't a general dedup check, which the docstring says.
- **Comment density** — the cited CLAUDE.md rule ("default to no comments; only add when the WHY is non-obvious") isn't in this repo's CLAUDE.md. What it does say is to favor explaining *why* over *what*, and to prefer code pointers over pasted code, which is what these are doing. Happy to trim if a maintainer disagrees, but I'd rather not cut rationale on a heuristic-heavy file on the strength of a rule that isn't there.

---

JSv4:
Both notes were actionable — fixed in `34663a3`.

**The terse-summary edge is real, and it reproduces.** `_summary_duplicates_body` measures contiguous coverage as a ratio over the **summary**, so a short summary is structurally fragile: at ~150 chars, one verbatim body sentence is already most of it.

| summary | length | verdict |
|---|---|---|
| quote + `"This is the core allocation."` | 155 | **dropped** ⚠️ |
| same quote inside a normal abstracting summary | 346 | kept ✅ |
| whole body passed as both args (observed mode) | 732 | dropped ✅ |

Retuning the threshold isn't the fix — loosening it lets the actual observed failure (whole-report duplication) back through. What was wrong is that this was **the one strip in the pipeline that said nothing**, and it has the biggest blast radius of any of them: the other guards trim a tail or a footnote, this one removes an entire user-visible section. So the drop now warns, and the warning names the terse-summary case so a reader who lost a legitimate summary learns why rather than just finding no summary. Both sides are pinned by test, and both known edges — opening-anchored, and ratio-over-summary — are now written into the docstring alongside the other guards' limitations.

**The `limit` default is a genuine magic number**, so it moved to `constants/annotations.py` per the repo's no-magic-numbers rule.

I did *not* add the suggested comment tying it to `RESEARCH_CITABLE_PASSAGE_MAX_HITS`, because I don't think they're meant to agree — they're independent numbers that happen to coincide at 10. `find_citable_passages` clamps to its own ceiling *before* it ever calls in, and that ceiling is research's to choose; a service default is the fallback for callers expressing no preference. Asserting a contract between them would invent a coupling that doesn't exist. That's the same ownership split as the floor/ceiling one in `8bb204c`, and the new constant's docstring says so explicitly so the next reader doesn't try to "unify" them.

On the third note — union polarity — agreed, and it's pinned as intentional in `d7ae737`.

One process note: everything here was checked with the pinned `black`, `isort`, and `flake8`, which I installed locally after the formatter miss in `f429735`. All three are clean across every Python file this PR touches.

---

JSv4:
Finding #1 was the best catch on this PR so far — accepted, but with a different fix than proposed. Finding #2 measures down to something narrower. Both in `801a7c4`.

**#1 — the allowlist really does leak.** Verified against the real function:

| heading | stripped? |
|---|---|
| `Sources`, `Bibliography` | ✅ |
| `Works Cited`, `Reference List`, `Sources Cited`, `Citation` | ❌ **leaks** |

And you're right that the miss is *silent* — the warning only fires on a strip, so #2200 reappears under a different name with nothing said. The set now enumerates what a report generator actually reaches for: the singular forms (`citation`, `reference`, `footnote`), `works cited`, `works consulted`, `reference list`, `list of references`, `source list`, `list of sources`, `sources cited`, `endnotes`.

**But loosening the match is the wrong direction for this set, measurably.** Stripping a *section* deletes every line up to the next heading. A token-overlap rule (`"sources" in title.split()`) strips all of these:

```
Sources of Supply Risk            -> would be deleted
References to Prior Agreements    -> would be deleted
Citations in the Record           -> would be deleted
```

Those are headings a legal research report legitimately carries, and losing a substantive section to a fuzzy match is far worse than leaving one scaffolding heading behind. So matching stays exact, the comment now records *why*, and it instructs the next reader to extend by name rather than loosen. Both directions are pinned by test — the variants strip, the mentions survive.

The two sets do get treated differently now, because their risk differs: dropping only a *heading line* keeps the prose, so a bare `Summary` earns a place there — an agent-written `## Summary` lands directly under the system's `## Executive Summary`, which is the doubled-scaffolding symptom itself.

**#2 — the guard is scoped, not skipped.** A marker's claim runs back to the previous span *or* the sentence boundary, whichever is nearer, so in a compound sentence each anchor answers for its own clause. Testing your exact shape with clauses long enough to clear the min-words floor:

```
… taxes assessed against the premises during the term <cite ids="TAX"/> and shall
additionally indemnify the landlord for every environmental remediation cost
arising on the site <cite ids="INSURANCE"/>.

-> cites_dropped = 1; TAX marker kept, INSURANCE marker removed, prose retained
```

The insurance anchor cited for a remediation clause **is** caught. Your trivial example passed only because `and maintain insurance` is 4 words — the `RESEARCH_CLAIM_SUPPORT_MIN_WORDS` floor, which applies to any short claim anywhere, not something specific to split markers.

Clause scoping is also the behavior I'd want: the alternative judges every anchor in a sentence against the union of all of them, which is the same over-permissive union shape already pinned as a limitation for polarity. Your point about the docstring stands though — "the guards apply identically to both sanctioned cite forms" said nothing about the claim's *extent*. It now states the scope explicitly, with both halves pinned.

---

JSv4:
Fence finding accepted and fixed in `3d07125` — and it's not the narrow edge case it was filed as.

**#1 — closing-fence run length.** You're right that CommonMark requires the closing fence to be at least as long as the opening. What that miss actually did:

```
Intro.
````            ← opens with 4
```             ← literal 3-run INSIDE it — was treated as a close
## Sources      ← now "outside" the block, reads as real scaffolding
````            ← re-opens
## Sources
- a fabricated citation list
## Findings
Real content.
```

Output before the fix:

```
Intro.
````
```
```

The heading inside literal content started a section skip that **swallowed the rest of the document** — `## Findings` and `Real content.` both gone. Same silent-content-loss class as the character bug from `f429735`, approached from the other side, and the two fail in opposite directions: ignoring the character leaves heading detection suspended forever (scaffolding unstripped), ignoring the length eats the tail of the report. Both rules now hold together, with the run-length case, the character case, and the plain single-fence case each pinned.

**Changelog nit — fair, and fixed.** Both fragments had grown into ~4 KB single-sentence chains because each review round appended one more clause. That's precisely what CLAUDE.md's rule 8 warns about ("a doc that only ever grows rots into noise; prune as you go"). They're now a headline plus short sub-points, with the calibration rationale left in the code rather than restated a third time in prose — which also answers your "same rationale in three places" observation directly.

**#3 — memoization: measured, not warranted.** A deliberately pathological report (200 cite spans over three shared 2000-char anchors) runs the entire verification pass in **22 ms**. Adding a cache to a pure function for that isn't a trade worth making; I'd rather leave the seam clean.

**Nit on the `limit` clamp** — keeping it as is. The split is deliberate ownership rather than indirection: the ceiling is research-specific and lives at the research call site, the floor belongs to `search_corpus_annotation_text`'s contract and lives there. Collapsing them into one expression is what `8bb204c` deliberately undid.

One recurring note: the comment-density point again cites a CLAUDE.md rule to "default to no comments, only when the WHY is non-obvious" that isn't in this repo's CLAUDE.md. I flagged that a few rounds ago too. The rule that *is* there asks for the why over the what and for pointers over pasted code — happy to trim on a maintainer's call, but I'd rather not cut calibration rationale from a heuristic-heavy file on the strength of a rule that isn't in the file.

---

**Comment by JSv4** [comment]
The empty-inner observation is a good catch, and it's more than an asymmetry — it's a hole through the claim-support guard. Fixed in `22d0996`.

`group(2)` is `""` rather than `None` for `<cite ids="1"></cite>`, so the span took neither path cleanly: the echo check was skipped (it tests truthiness), quote verification ran against an empty string, and the resulting empty claim fell through to the carried-over `last_claim` — judging the span against a *previous* sentence, or against nothing at all when it came first.

Measured on one unsupported sentence and one anchor:

```
The landlord shall remediate every environmental condition discovered on the
site at its sole cost and expense  <SPAN>

  <cite ids="1"/>        ->  citation dropped   ✅
  <cite ids="1"></cite>  ->  citation SURVIVES  ❌
```

Same text, same anchor, opposite outcome. The empty form walks an unattributable sentence straight past the check #2201 exists to enforce.

Fixed by normalising an empty or whitespace-only inner to `None` at the top of the loop, so both spellings run one code path rather than the divergence being documented. All three spellings now agree in both directions — unsupported dropped, supported kept, survivor rendered as a marker — and that's what the test asserts.

You're right that the prompt asks for the marker form and this is unlikely from a well-behaved agent. But the premise of every guard in this PR is that the agent is stochastic and writes things the prompt forbids — that's how both #2200 and #2201 happened in the first place, so "unreachable in practice" is the assumption I'd least like to lean on here.

Everything else in the review is confirmatory and needs no action. Noted on running CI before merge — that's been my standing caveat throughout, since the Docker-based backend suite has never run locally in this session.

---

**Comment by JSv4** [comment]
Good catch on the rule violation — fixed.

**PR description.** You're right: CLAUDE.md's Baseline Commit Rule 3 names "PR messages" explicitly, and the body carried the attribution footer. Removed. While rewriting it I also brought it current — it predated roughly twenty commits of added guards and no longer described what the branch actually does (nothing about the polarity guard's known limitations, the warning coverage, the fence rules, the heading-variant set, or the empty-wrapping-span normalisation).

Worth flagging the tension rather than leaving it implicit: this environment's instructions require an attribution footer on every GitHub *comment* I post, which is the same thing rule 3 forbids for "comments." I've resolved it the only way I can — the PR body is mine to control, so it's clean; the per-comment footer is appended by the harness and reappears here regardless of what I write. That's a maintainer call to settle, not something I can fix from inside a comment.

**Label normalization** — you flagged it as latent and no-action, and that's right today, but the comment now sits where someone would trip it (`3d31594`). The two consumers really do normalise differently:

| consumer | normalisation | `Section_Header` matches? |
|---|---|---|
| warning path (`_is_header_anchor` → `_normalize_label`) | case + separators | yes |
| retrieval filter (`exclude_label_texts` → `iexact`) | case only | **no** |

They agree solely because every entry in `RESEARCH_HEADER_ANCHOR_LABELS` is already in canonical spelling. Add a separator variant and they split in the bad direction: the anchor gets warned about but stays on offer as a citable passage — the #2180 shape the exclusion exists to prevent. The note records both ways out (use the parser's spelling, or teach the filter the same normalisation).

The rest is confirmatory and needs no action. Agreed on CI before merge — that's been the standing caveat, since the Docker-based backend suite has never run in this session.

---

**Comment by JSv4** [comment]
The `find_citable_passages` → direct-cite gap is the sharpest finding on this PR. You were right on every part of it, and it's fixed in `6265081`.

**The mismatch was mine.** The `cited_ids` gate predates this PR, but this PR is what makes an agent walk into it: `find_citable_passages` exists to hand back a paste-ready cite handle, and the new prompt bullet says the `annotation_id` *is* the cite handle. The obvious shortcut the tooling now invites was the one path `finalize` refused.

And your read of the failure modes is exact — an id outside `cited_ids` hydrates no anchor text, so it cannot support anything:

| claim length | what happened |
|---|---|
| ≥ `MIN_WORDS` | dropped under *"not supported by the passage it cited"* — **names the wrong cause** |
| < `MIN_WORDS` | passes the verifier, then `_render_citations` drops it **silently** |

I took your option (b) rather than tightening the prompt, because the prompt isn't wrong — the pipeline was. `cited_ids` now seeds from the composed document as well as from findings (via a shared `_cited_ids_in` helper reusing the verifier's own regex and id parser, so "what the document cites" means one thing across the pipeline), and the `& retrieved_annotation_ids` intersection moves after that union to become the single gate:

```
cited_ids = (finding ids | document ids) & retrieved_annotation_ids
```

**The closed citation graph is unchanged.** `record_finding` was never the security property — `retrieved_annotation_ids` is, and it still gates everything. Since every retrieval tool is permission-filtered, that same intersection is also what keeps a citation inside what the run's creator may read. Both directions are pinned: a retrieved id cited with no finding renders *and* links provenance; an id retrieval never surfaced renders no footnote and links nothing.

**On the `limit=0` nit from the newest review** (`6f74a4f`) — agreed, and it's a fair reversal of where I put the floor in `8bb204c`. That change was about not duplicating the number; it landed the rule in the wrong layer. `search_corpus_annotation_text` is a shared `AnnotationService` method and "at most N" is what `limit` means everywhere else, so `0` now yields no rows (negative treated as 0, since Django rejects negative slicing). The floor was never a property of annotation search — it's a UX rule of `find_citable_passages`, which never wants to hand a model an empty page for a phrase that *did* match. It now clamps to `[1, ceiling]` at that call site. No duplication reintroduced: one floor, in the layer that owns the rule. Composition verified unchanged for every input the tool can receive (`None`/`0`/negative/`1`/`7`/`999` → 1, 1, 1, 1, 7, 10).

The `finalize` length nit is fair and I've left it — a `_build_finalize_warnings` extraction is a clean follow-up, but I'd rather not restructure the function in the same PR that keeps adding guards to it.

---

**Comment by JSv4** [comment]
Confirmed, and it's worse than "narrow" — fixed in `21dc280`.

Reproduced exactly as you described:

```
Real prose before.
## Sources
```                              ← unbalanced fence inside the SKIPPED section
- [1] a fabricated citation
## Appendix
Substantive appendix content.
## Conclusion
The conclusion.
```

Output was one line: `Real prose before.` The Appendix, the Conclusion, and all their content — gone. And `sections_stripped` still reported **1**, so the warning says one section was removed while four-fifths of the report is missing. Silent, and actively misleading about the blast radius.

What makes this the sharp version of the bug: unbalanced fences *are* malformed LLM markdown, and an agent writing scaffolding it was told not to write is the entire premise of #2200. So the trigger arrives through the door this function was built to watch.

**The fix is structural, not a patch:** fence state is no longer tracked while skipping. That's sound rather than convenient — a skip can only begin on a *detected* heading, and headings are only detected outside a fence, so entering a skip already implies no open fence. That's precisely what keeps heading detection alive through the skip so the closing heading is found. A fence opened inside the skipped region belongs to content being discarded and has no bearing on what survives.

I preferred this to your alternative of pinning the current behavior in a test — "the rest of the report silently disappears" isn't a tradeoff worth documenting.

Verified the four already-pinned behaviors are unchanged (character parity, run length, nesting-aware skip, fenced heading surviving as content), plus the new case in both spellings — unbalanced *and* balanced fences inside the skipped section.

That's three separate silent-content-loss bugs now found in this one helper's fence handling — character parity, run length, and skip-region leakage — each from a different reviewer. Worth noting for anyone touching it later: all three failed quietly, which is why they needed reproduction rather than reasoning.

---

JSv4:
All three notes are fair. None needs a behaviour change, but two asked questions the docstrings should already have answered — recorded in `027e3ff`.

**1. `_summary_duplicates_body` cost — measured rather than guessed.** The needle is capped at 400 chars, the body isn't, so cost is linear in the report:

| body | time |
|---|---|
| 10 KB | 30 ms |
| 50 KB | 172 ms |
| 200 KB | 863 ms |
| 1 MB | 5,746 ms |

≈6 ms/KB. At the sizes an agent actually writes that's tens of milliseconds, against a Celery task that just spent minutes in the model — so no cap, and the docstring now says why not. Two reasons beyond cost: capping the haystack would blind the check to a summary copying the body's *middle*, which it catches today; and unlike the echo guard there's no arithmetic short-circuit available here, because the unbounded side is the **haystack**, not the ratio's denominator. (In the echo case the unbounded side *was* the denominator, which is what made `coverage ≤ len(preceding)/len(inner)` an exact bound. That trick doesn't transfer.)

**2. pg_trgm and short phrases — you're right, the claim was overstated.** A trigram index has nothing to look up for a phrase under three characters and falls back to a scan. I didn't add a length floor: a one- or two-character phrase makes a useless citation anchor, and the caller's row cap bounds the result either way. But the docstring now states the exception, since callers passing model-supplied text should know.

**3. Bare `"No."` — correct, and now documented.** The period discriminator can't tell an abbreviation from a one-word answer, so `"Is that permitted? No."` reads as the citation abbreviation and doesn't negate. It errs toward *not* firing the inversion guard, which costs strictness rather than attribution — the safe direction for this guard. The `Inc.`/`U.S.C.` limitation next door was written down; this one should have been too.

---

JSv4:
Agreed, and fixed rather than documented — `1cdf960`.

You're right that I chose the wrong remedy last round. A comment telling the next editor to keep two things in step is a burden, not a fix, and the failure it guards is silent. Both call sites already existed, so making them agree by construction was the cheaper option all along.

The divergence is real and the warning side is already pinned: `test_is_header_anchor_keys_on_label` asserts `_is_header_anchor(label_text="section_header")` is `True`. Meanwhile the retrieval filter compares in SQL with `iexact`, which folds case but not separators. Measured across the spellings that could drift:

| stored label | warned as header | old filter | new filter |
|---|---|---|---|
| `Section_Header` | yes | **no** ⚠️ | yes |
| `Section-Header` | yes | **no** ⚠️ | yes |
| `OC SECTION` | yes | **no** ⚠️ | yes |
| `Page_Header` | yes | **no** ⚠️ | yes |

Each of those was flagged as a header *and still offered as a citable passage* — #2180 reintroduced silently, and only for whichever parser spells it that way.

`RESEARCH_HEADER_ANCHOR_LABEL_VARIANTS` expands the set across separator spellings for the SQL side: 11 OR terms instead of 5, no change to query shape, and no rule for anyone to remember. The three `Section Header` spellings are now pinned against the retrieval filter directly, asserting both consumers agree.

I didn't route the filter through `_normalize_label` itself, as that would mean a per-row `REGEXP_REPLACE` in SQL to defend a case that expansion handles in pure Python at import time. Runs of separators aren't expanded — the parsers emit single ones — and that limit is written down rather than left to inference.

On your calibration note: agreed that `MIN_COVERAGE = 0.25` and `MIN_WORDS = 12` come from measured examples rather than an eval set, and that the warning counters are the signal to watch after rollout. That's a good argument for the counters being there beyond user-facing honesty — they're the retuning telemetry.

---

JSv4:
You filed this as plausible rather than confirmed because you couldn't execute. It reproduces — and it's a regression **I introduced in `21dc280`** two rounds ago. Fixed properly in `f757f5b`.

Your traced input, run against the real function:

```
Real prose.
## Sources
```
## Findings          ← heading-SHAPED line, inside a fence, inside the skip
citation garbage
```
- more garbage
## Conclusion
Real text.
```

Output before the fix — `## Findings` ended the skip early, exactly as you predicted:

```
Real prose.
## Findings
citation garbage
```
- more garbage
## Conclusion
Real text.
```

**The two fence bugs were mutually exclusive as long as fence state and skip state were interleaved.** Track fences during a skip and one unbalanced ``` suspends heading detection permanently, taking the report's tail (`21dc280`'s bug). Suspend them during a skip and a fenced heading ends the skip early, leaking scaffolding (this bug — which `21dc280`'s fix created). I traded one for the other without noticing the trade.

So neither is a trade worth making, and neither arises once the two are independent: **fence spans are now resolved up front over the whole document**, with no reference to what's being skipped, and the loop only asks whether a line index falls inside one.

That leaves what to do with an opener that never closes. CommonMark says it runs to end of document — precisely the reading that costs the report its tail. In LLM-authored markdown a lone ``` is far likelier a stray artifact than an intent to code-block everything after it, so it's ignored and scanning continues past it, with a later balanced pair still recognised. A deliberate departure from the spec, written down at the function.

Verified all seven behaviours hold together:

| case | result |
|---|---|
| fenced heading inside a skip (this review) | ✅ |
| unbalanced fence inside a skip (`21dc280`) | ✅ |
| fence-character parity (`f429735`) | ✅ |
| closing-fence run length (`3d07125`) | ✅ |
| nesting-aware section skip | ✅ |
| balanced fence inside a skip | ✅ |
| fenced heading as content, no skip | ✅ |

Pre-pass cost on a pathological input (500 unmatched openers over 2000 lines): **0.4 ms**.

That's four distinct silent bugs in this one helper's fence handling, one of them mine. Worth stating plainly: the state machine was the problem, not any individual rule — which is why this round replaced the design rather than adding a fifth rule to it.

On the docs note — agreed the `find_citable_passages` story is told three times. I trimmed the changelog to point at the code rather than restate it, and also repaired a sentence there that an earlier edit of mine had spliced into a duplicated clause.

---

**Comment by JSv4** [comment]
⚠️ **Correcting a premise in this review before anyone merges on it.**

> The PR states 10428 tests passed in CI

It did not, and I should have caught this being repeated sooner. **10428 passed / 2 failed was a *local* run of mine**, from before this container's disk allowance ran out, and the 2 failures were my own (the wrong patch target, fixed in `575a2f7`). I've said in every comment that the backend suite has never run locally in this session — but that number is still floating around, so let me replace it with what CI actually shows.

**What CI has actually verified.** I went through `backend.yml`'s run history for this branch:

| head | pytest |
|---|---|
| `8bb204c` | ✅ **green** — "Run Backend Test Suite" success, 06:17:15 → 06:35:22 |
| the 16 commits since | ❌ **never completed** |

**Why.** `backend.yml` sets:

```yaml
concurrency:
  group: ${{ github.head_ref || github.run_id }}
  cancel-in-progress: true
```

pytest takes ~18 minutes. I've been pushing review fixes every 3–10 minutes. Every push cancelled the in-flight run before the suite finished — which is also why recent runs report zero jobs rather than a failure. Not a CI fault; a consequence of my own cadence.

So the honest state is: **everything through `8bb204c` is CI-verified green. Everything after it is unverified**, and that window contains real behaviour changes, not just docs — the `cited_ids` widening (`6265081`), the `limit` contract change (`6f74a4f`), the fence-span redesign (`f757f5b`), empty-span normalisation (`22d0996`), and header-label variants (`1cdf960`). It also contains roughly a dozen *new* tests that have themselves never executed.

**I've stopped pushing** so the run on `f757f5b` can finish. I'm not treating this PR as mergeable until pytest completes green on the current head, and I'd ask that nobody else does either. If it comes back red I'll fix it; if a further review finding lands meanwhile, I'll batch it rather than push immediately and restart the clock.

Worth noting for the reviewers reading statically: "reviewed the diff and found no issues" and "the tests pass" are different claims, and this thread has been accumulating the first while the second went stale.

---

JSv4:
The `pytest` failure on `6f74a4f` is **not an assertion failure and not caused by this branch** — it reproduces on `main`, and predates the branch.

**What actually happened.** The suite ran to ~99%, then a pytest-xdist worker died and the job hung until the runner's 100-minute ceiling:

```
.uuuuuuuuuuuuuuuuuuuuuuuuu.[gw2] node down: Not properly terminated
##[error]The action 'Run Backend Test Suite' has timed out after 100 minutes.
```

**The same thing on `main`,** commit `e46c35e`, run [#6996](https://github.com/Open-Source-Legal/OpenContracts/actions/runs/30131052467) — started `2026-07-24T22:28`, which is *before this branch existed* (PR opened 03:48 the next morning):

```
............[gw3] node down: Not properly terminated
##[error]The action 'Run Backend Test Suite' has timed out after 100 minutes.
```

Same shape, same ~99% point, different worker. And it isn't isolated — scanning `backend.yml` durations on `main`, the normal run is ~23 minutes, but two recent ones took ~110:

| run | head | duration |
|---|---|---|
| #6998 | `a049cce` | 23 min ✅ |
| **#6996** | `e46c35e` | **111 min** ⚠️ |
| #6995 | `a49f83c` | 25 min ✅ |
| **#6980** | `5e2c099` | **109 min** ⚠️ |

So: a flaky xdist worker death on the shared `yuge` runner, pre-existing, roughly 1-in-10.

**I also need to correct my previous comment.** I said every push cancelled the in-flight run. That was only part of it, and not the main part. This job wasn't cancelled by my four subsequent pushes — it *hung*, holding the single self-hosted runner for 100 minutes. The `f757f5b` jobs started at **09:21:37**, the same second the timeout fired and released it. So the dominant effect was runner starvation caused by the hang, not my cadence. My cadence made it worse; it didn't cause it.

**Where that leaves the branch:** pytest was green here at `8bb204c`; the commits after it are still unverified, and `f757f5b`'s pytest is now queued behind its linter. I'm still not pushing, so it gets a clean shot. If it dies the same way, that's the pre-existing flake and a re-run is the answer — worth someone looking at separately from this PR, since a hung worker costs 100 minutes of a shared runner each time it happens.

---

PR #2215, opened 2026-08-01
Authority packs and cross-corpus deep research

JSv4:
Cross-corpus deep research over curated public regulatory sources, plus the reliability work that made it trustworthy enough to measure.

Eight commits, ordered so each builds on the last. Reviewable individually.

## What this adds

**Authority packs** (commits 2–3) — a framework for turning a published body of regulation into corpora that can be cited: pack config and install services, source-plan and discovery provider base classes, a frontier service tracking what remains unfetched, and the governance graph. Fetching is constrained by an SSRF allowlist derived from each pack's declared hosts, matched exact-or-dot-suffix so `notercot.com` cannot satisfy `ercot.com`.

The framework is domain-neutral, and as of commits 9–12 the tree contains nothing but framework. The DFW packs that exercise it — Texas statutes, PUCT rules and proceedings, ERCOT large-load rules, Oncor tariffs, ten corpora — are data, so they now live in their own repository (`Open-Source-Legal/TX_Large_Loads`) and are **sideloaded**. See "Packs are deployment data" below.

**Deep research over corpus groups** (commit 5) — a `ResearchReport` gains a `corpus_group`, and the agent a `search_across_group` tool that fans retrieval over every member corpus *the caller may read*, membership resolved per query rather than from a config-time snapshot. Reachable from the interface as well as chat. A group the caller cannot see is refused rather than ignored: a silently narrowed run produces a report that reads as group-wide and is not.

**Structured finding cards** (commit 4) — findings that commit to a shape, so the report, the interface and downstream consumers read the same fields instead of re-deriving them from prose. Regime cards are built around a half-open interval whose **end is exclusive**, because an inclusive end lets both regimes claim the boundary day. Obligation cards keep roles and dates apart: whoever prepares a filing is often not who submits it, and an approval date is not an effective date is not the date an application is due.

The thresholds an obligation attaches at are **configuration**, declared by a corpus in its own CAML article (`[component:obligation-schema unit=MW steps=25,50,75,100 label=ramp]`), not a constant in the schema. An unconfigured corpus keeps the guard that matters — a phase-triggered obligation must still say where it bites — and loses only the membership check, which would have nothing to check against.

**Responses-API models** (commit 1) — the GPT-5.6 family rejects function tools on `/v1/chat/completions` when `reasoning_effort` is set, and rejects a `function_call` whose paired `reasoning` item is missing. In-run compaction was producing exactly that by dropping `ThinkingPart`s from tool-calling responses, which Chat Completions had tolerated for years. Both handled, and context windows registered so a 1M-window model is not sized at 128K.

## Reliability work worth reading (commit 5)

Three defects that only surface under sustained use:

- **A run is capped on model *requests* as well as tokens, and could not see that counter.** It was cut off mid-work, and the salvage composition that replaced its report body silently lost whatever it had not yet written — observed at exactly 60 of 60 permitted requests, with two of four required sections missing. The agent is now told when its step budget runs down.
- **Every non-finalize ending was recorded as one opaque `budget_exhausted`** — token budget, step budget and an agent that simply stopped, all indistinguishable — and `chat()` swallows the framework exception, so the worker log said nothing either. Reports now carry a `terminal_reason`.
- **`tool_call_log` covered only the closures the research task builds.** Everything from the agent's default toolset, `similarity_search` included, was invisible, so the log held about half of what a run did while the interface presented that half as the whole. Reading a tool's *absence* off that half-log produced a confidently wrong conclusion during development. Now covered — with the closures skipped by name, because the factory re-wraps caller tools and a marker attribute never survives.

`finalize_report` is also terminal in fact rather than only in its docstring: a second call used to re-run composition, leaving the later body stored and both passes' warnings accumulated.

## Packs are deployment data (commits 9–12)

Shipping a body of regulation in the product meant every install — anywhere, for anyone — carried this jurisdiction's regulatory topology, catalogued it in the Authority Console, and imported its scrapers into every worker. A body of regulation has its own legal-review state, refresh cadence and publisher relationships; it belongs on its own release cycle. The four packs, the three test modules asserting *their* identities, the deployment runbook and the full-stack import spec moved to `Open-Source-Legal/TX_Large_Loads` (private). `bolivia` stays as the worked example the format needs.

**Sideloading.** `AUTHORITY_PACK_ROOTS` (new) takes directories *of* packs, so a pack repository mounts with one variable; `AUTHORITY_PACK_PATHS` (individual packs) is unchanged and both are honoured, de-duplicated by resolved path. `load_authority_pack --check` runs the Console's preflight from the command line and writes nothing — the Console needs an authority-admin browser session, which a headless deployment does not have.

**Two defects the move surfaced.** Sideloaded packs with a helper module never worked: in-pack component modules are imported by file path, so their parent packages did not exist and a pack could not import its own sibling. Two packs had been reaching their pack-root helpers by absolute in-tree dotted path — which resolves only in-tree — and silently failed to register 15 providers once sideloaded, contradicting the format's central promise. The registry now creates each pack's parent packages, and drops cached submodules when a pack name is remounted from a different directory (re-pointing `__path__` alone is not enough: an import resolves against `sys.modules` first, so the old directory's helper was handed back).

**Tests keep the machinery coverage without the data.** A complete, deliberately fictional schema-v2 fixture pack lives at `opencontractserver/tests/fixtures/authority_packs/example_utility` — under `tests/fixtures/` rather than the in-tree pack root, because `authority_pack_dirs()` enumerates every immediate subdirectory of that root and a fixture placed there would ship as installable. `test_authority_pack_loader.py` (convergence, idempotency, curator-state preservation, atomic abort, prefix binding), `test_authority_discovery_runtime.py` (renamed only — it always defined its own providers) and `test_authority_pack_sideload.py` (new — the sideload contract) replace the five `test_grid_dossier_*` modules.

**Deliberately left alone.** PUCT/ERCOT citation regexes stay in `grammars.py`: a pack's declarative vocabulary (`shape_rules`/`abbreviations`) cannot express a bespoke pattern, and inventing a `citation_patterns:` mechanism inside the precision-critical `GenericCitationExtractor` is not warranted here. `AuthorityPacks.ct.tsx` and the import-modal wrappers still use "Texas Electric Law" as mock data — they read no pack files.

**One correction to the note below.** `test_loads_exact_identities_typed_schemas_and_relationships` was not a test-isolation bug. `oncor_delivery/pack.yaml` binds `oncor-tariff` to its current-tariff corpus, so the `authority_corpus_id is None` assertion had been stale since this branch's own "bind every prefix" fix; the id varies per run because it is a real corpus id. Corrected in the moved test.

## What is deliberately not here

No evaluation harness and no deployment manifest. Tooling that reaches inside the application to grade it is a second implementation of the product's own checks, free to drift from them; and a customer's corpus topology is a property of their deployment, not of the software. The acceptance procedure is written against the GUI and the GraphQL API and lives with the packs (`TX_Large_Loads/docs/acceptance-gate.md`); the group loader, its manifests and its run evidence live in a separate repo again.

## Testing

~14k lines of tests, roughly 1.1× the code they cover. Two new `CLAUDE.md` pitfalls document the Responses-API family and the audit log's blind spot. The golden SDL is regenerated deliberately rather than left to drift.

(Superseded: the `test_grid_dossier_authority_pack_loader` failure described in earlier revisions of this description was a stale assertion, not a test-isolation bug — see "Packs are deployment data" above. It is fixed in the moved test.)


JSv4:
Review responses — merged `main` and pushed `a3029af5a`.

## Fixed

| Finding | Fix |
|---|---|
| **Med** — `append_tool_call`/`append_finding` lost-update race *(both reviews)* | `select_for_update()` inside `transaction.atomic()`. `research_tasks.py` has no enclosing `atomic` and Celery runs autocommit (`ATOMIC_REQUESTS` is HTTP-only), so this is a brief per-call row lock, not a task-long one. |
| **Med** — pack namespace basename collision (`registry.py`) | New `_pack_namespaces()`: plain basename when unique, short path digest on collision + warning. Keeps existing module names stable. |
| **Minor** — `_attach_publisher_original_file` outside per-document isolation (`import_tasks_v2.py:540`) | Now returns `(None, {})` like the outer handler, honouring the "accept partial state on failure" contract on every path. |
| **Nit** — magic numbers | `MAX_DERIVED_MESSAGE_TITLE_CHARS` + `MAX_WORKSPACE_FILENAME_CHARS` → `constants/truncation.py`; new `MAX_EXTRA_CA_CERTIFICATE_BYTES` in `constants/safe_http.py`. |
| **Nit** — `PermissionError` caught locally | Comment recording the intent, as the review asked. |
| **Test gap** — `SaveMessageToWorkspace` thrown-exception branch | New component test. |

## Dismissed, with evidence

**XML entity expansion — not exploitable on this runtime.** I tested the actual base image: `python:3.12.13-slim-bookworm` ships **expat 2.7.4**, whose input-amplification limit is enabled by default. A billion-laughs payload dies in the parser as `ParseError`, and *both* `ET.fromstring` sites already translate that into the ordinary "malformed" `ValueError`. Nothing allocates. Rather than add `defusedxml`, I added a regression test that fails if a future base image regresses below expat 2.6 — at which point the dependency becomes necessary.

**`Command(AuthorityPackService, BaseCommand)` MRO — the base class is load-bearing.** The review read it as vestigial because the command body always calls `AuthorityPackService.<method>` explicitly. But ~30 call sites in `test_authority_pack_loader.py` resolve those validation helpers *through* `Command` (imported as `AuthorityPackLoaderCommand`). I made the change, 16 tests failed, and reverted it with the reason documented in a comment.

**`gpt-5.6-sol` / `-terra` / `-luna` are real model IDs**, not placeholders — `CLAUDE.md` pitfall #20 records the observed live API error text that motivated the Responses-API routing.

## The two High findings — tracked, not fixed here

Both are real and both are silent data loss, but each fix changes a deliberate invariant rather than correcting a slip, so they should be decided explicitly rather than folded into this PR:

- **#2219** — resumed research runs delete previously-valid findings. The fix means widening `cited_ids &= set(retrieved_annotation_ids)` across resume segments, and that line is documented as *the* enforcement point for both the closed citation graph and the permission boundary.
- **#2220** — relationships spanning an unchanged document are dropped on incremental reimport. The fix requires reconstructing export-local→annotation id mappings for a document deliberately not re-parsed, changing the contract with `finalize_corpus_import_relationships`.

Both issues carry the full trace, the failure path, and candidate approaches. #2220 also notes a cheap interim improvement: log dropped relationship endpoints instead of discarding them silently.

## Deferred

`OPENAI_RESPONSES_ONLY_PREFIXES` needs a config override for future Responses-only models · zip-slip guard duplicated across `import_tasks_v2` / `export_tasks_v2` / `validate_export` · missing outer transaction on the re-import-into-existing-corpus path · unbounded corpus fan-out in `find_citable_passages` · `custom_meta` merge race · no confirm step on corpus re-embed · decomposition of `authority_import_artifacts.py` / `authority_pack_service.py`.

## Verification

409 backend tests pass across the touched modules (exit 0) · 4 new backend tests pass · 6/6 `SaveMessageToWorkspace` component tests pass · `tsc --noEmit` clean · pre-commit green · 2 changelog fragments validated.

JSv4:
Pushed `6c82cc1f5` — fixes the pytest failure. It was not an assertion problem but a **600-second deadlock**, and the cause is a suite-wide landmine rather than anything in this PR.

## Root cause

Migration `0031_add_pipeline_settings` seeds the `PipelineSettings` singleton with a one-shot `RunPython`, so its row lives *outside* any test transaction. Django's `flush` — run on every `TransactionTestCase` teardown — truncates `documents_pipelinesettings` along with every other table, and with `serialized_rollback` disabled (the default) **nothing restored it**. Under `pytest -n 4 --dist loadscope`, any `TransactionTestCase` running before an embedder-touching test on the same worker left that worker with no singleton for the rest of the session.

Because `get_instance()` re-creates the row lazily, the absence did not fail loudly — it deadlocked:

```
main thread                              asgiref executor thread
AsyncToSync.__call__                     get_embedder
  run_until_future()                       get_default_embedder_path
    _work_ready.wait()   ←── waits ───      PipelineSettings.get_instance
                                              get_or_create(pk=1) → INSERT
         └──────── blocks on ────────────────────┘  (own connection, cannot see
                                                     the uncommitted row)
```

Neither side advances until `pytest-timeout` fires at 600s. The hang then leaves `unittest.mock.patch` decorators un-exited, which is why one hung test failed three *later* tests in the class sharing the patched target — the `'Mocked default prompt'` assertions were collateral, not independent failures.

## Why it appeared now

Nothing in this PR touches `TestCoreAgentFactoriesDefaults`, `PipelineSettings`, or the embedder path — the failing tests exist unchanged on `main`, and this PR's only edit to that file adds a test to a *different* class. The four backend tests added in `a3029af5a` shifted the `--dist loadscope` distribution, which changed what ran before this class on `gw3` and armed the landmine.

## Fix

An idempotent `post_migrate` receiver (`opencontractserver/documents/_pipeline_settings_seed.py`, connected in `documents/apps.py`) re-seeds the singleton after each flush and repairs reused/poisoned CI volumes at DB setup.

This mirrors the existing `enrichment/_namespace_seed.py`, which already solves the identical problem for the shipped `AuthorityNamespace` rows — including its fallback to the global app registry when `post_migrate` arrives without an `apps` kwarg, which is exactly how `flush` emits it (`flush.py` vs `migrate.py`). `PipelineSettings` simply never got the same treatment.

## Verification

The failure reproduces **deterministically** before the fix:

```
pytest test_pipeline_settings.py test_core_agents.py::TestCoreAgentFactoriesDefaults
→ 2 × Timeout (>90s) + 3 × 'Mocked default prompt'   # same 5 failures as CI
```

After the fix that command exits 0, and a flush-then-async-agent run passes 11/11. Three new tests pin the receiver on the apps-less emission, its idempotency against operator edits (a receiver that wrote unconditionally would revert System Settings on every deploy), and the real `flush` command end to end.

Worth noting the class of bug is broader than this one row: any migration-`RunPython`-seeded table without a `post_migrate` convergence has the same exposure, and lazy re-creation from an async path turns it from a missing row into a hang.

JSv4:
**Correction to the verification section above** — one repro line I quoted was not a clean A/B, and the PR record should be accurate.

I wrote that the failure "reproduces deterministically" with:

```
pytest test_pipeline_settings.py test_core_agents.py::TestCoreAgentFactoriesDefaults
```

That command *did* produce the exact CI failure (2 timeouts + 3 `'Mocked default prompt'`), but only because the local test DB had already been poisoned — by a throwaway `TransactionTestCase` I ran earlier while bisecting the cause. `test_pipeline_settings.py` contains **no** `TransactionTestCase`, so on a freshly created DB that command flushes nothing and cannot reproduce the hang. Its "before" and "after" runs differ in database state as well as in code, so it does not isolate the fix.

The verification that does hold, and that the fix rests on:

```
pytest test_pipeline_settings_flush_reseed.py test_core_agents.py::TestCoreAgentFactoriesDefaults
→ 11 passed
```

Here the first file *is* a `TransactionTestCase` that genuinely flushes, so the sequence is the real CI scenario — truncation followed by an async embedder-touching test — on a DB seeded by `--create-db`. That is a clean A/B on the mechanism. The three new tests (apps-less `post_migrate` emission, idempotency against operator edits, and the real `flush` command end to end) are likewise independent of the poisoned-DB state.

The same over-broad repro sentence appears in the commit message of `6c82cc1f5`. I have left the commit alone rather than force-push a shared PR branch; this comment is the correction of record.

**One operational note.** The receiver restores the singleton on `migrate` and on `flush`. It does **not** heal a database that is *already* missing the row in a run where neither fires — which is exactly why my poisoned local DB stayed broken across runs of all-`TestCase` files. That is not a gap for CI: the runner is ephemeral and, per `.github/workflows/backend.yml`, pytest-django creates the per-worker test databases cold on first invocation, so `post_migrate` fires and seeds the row before any test runs. It would only matter for a long-lived local or self-hosted volume that was poisoned before this fix landed — there, a one-time `pytest --create-db` clears it.

JSv4:
Pushed `3d464a3d0` — all three correctness findings from the latest review are fixed.

## 1. Corpus-group slug-enumeration oracle

`research_tools.py::_load_group` used an unfiltered `CorpusGroup.objects.filter(slug=...)`. Correctly identified: no unauthorized access followed, because `ResearchReportService.start` still refused the run — but the two failure modes *replied differently*, and that is the whole oracle:

| slug | before |
|---|---|
| doesn't exist | `"no corpus group named X"` (tool's own string) |
| exists, not visible | `"Could not start: ..."` (start()'s `PermissionError`) |

Now routed through `CorpusGroupService.get_group_by_ref`, matching the sibling `core_tools/multi_corpus.py::_resolve_group_corpora` you pointed at — and fixing the services-layer violation at the same time. Both cases return one refusal.

## 2. `finalize_report` guard was check-then-act

You noted `append_finding`/`append_tool_call` got `select_for_update` for exactly this reason while `finalize` did not — that inconsistency was mine, from the previous round, so thanks for catching it.

New `ResearchReportService.finalize_once` runs the status check and the composition as **one critical section** under the row lock; the loser blocks, re-reads COMPLETED, and is turned away. Raw `finalize` keeps its existing non-idempotent contract, so `test_finalizing_twice_appends_its_warnings_twice` still pins the behaviour that makes the guard necessary.

I also applied it to the **salvage path** in `_run_deep_research_async`, which had the same check-then-act shape. That one is worse than a duplicate: a salvage composition landing after another worker's real finalize would overwrite a genuine report with a "the run ended before the agent produced a final report" note.

## 3. Missing terminal row on the source-less reimport path

Fixed as described. Worth adding one detail to your diagnosis: the branch is gated on the source being **non-reingestable**, i.e. the NUL placeholder the V2 exporter writes for text/markdown members — so "unchanged markdown member in a re-shipped pack" is exactly the trigger, and it is the ordinary case rather than an exotic one.

The row's `id_map` is empty for the same reason the converged path's is; relationship endpoints landing on unchanged documents stay tracked in #2220 rather than being papered over here.

## On the test for #3

My first attempt at this test **passed without the fix** — it was reaching `_reingest_document_with_deferred_remap` (which already writes its own row) rather than the fallback branch, because the fixture's source was reingestable. Rewrote it around a NUL-placeholder member and validated it properly: **exit 1 with the change reverted, exit 0 with it**. It now guards the branch instead of merely executing near it.

## Verification

3 new tests · full affected suite green (`research/`, `test_authority_targeted_import`, `test_import_v2_reingest_remap`, `test_corpus_export_import_v2`) · pre-commit clean.

Not addressed, as minor/stylistic: the duplicate float-formatting helpers, the `GROUP_NOT_FOUND_MESSAGE` constant reuse, the dead `MARKDOWN_MIME_TYPE` import, the GraphQL fragment for `AuthorityPack` field selection, the empty-catalog-vs-`PermissionDenied` API contract question, and the `build_agent_model` redirect-ternary duplication. Happy to take any of them if you want them in this PR rather than a follow-up.

JSv4:
Pushed `8cc7a7e8c` — all four Major findings fixed, each with a test verified to fail with the fix reverted.

## 1 + 2. Post-commit failures on the pack install path

Both correct, and they are the same defect twice. `_install_plan` commits taxonomy/corpora/relationships in one `atomic()` block; the relink runs after it. You named the relink — the `refreshed = cls.preflight_path(...)` at L243 has it worse: it is *inside* the caught tuple, so a `CommandError` there quietly turned a committed install into `ok=False` with a validation-shaped message. Neither is recoverable by retrying, because the retry is idempotent and no-ops.

Both now degrade to `post_commit_warnings` on `AuthorityPackInstallResult`, surfaced three ways (GraphQL `message`, `result.warnings`, and a `style.WARNING` line from `load_authority_pack` — otherwise the CLI would go *silent* on a failure that used to abort it). `relink_summary=None` deliberately does not carry this: it already means "relink was not requested". The refresh failure falls back to the pre-install plan rather than returning `None`, since that plan's fingerprint is the one the operator approved.

`IntegrityError` is caught separately rather than added to the L245 tuple — widening the tuple would put `duplicate key value violates unique constraint "uniq_corpus_slug_per_creator_cs"` into the Authority Console. `_install_plan`'s `atomic()` means the loser committed nothing, so the message says exactly that.

No advisory lock: it would serialize a race whose observable end state you asked to be a clean `ServiceResult.failure`, and the retry already converges. `install_path()` (the management-command adapter) is deliberately left raw — an operator running a load command should see the traceback.

## 3. Gate fault isolation

Correct on both the symptom and the mechanism. The row is `in_progress` from L431, `dequeue_queued` only returns `queued`, so a stranded row is invisible to every later crawl until someone resets it by hand — and it takes the crawl batch down with it.

The fix that *looks* right here is extending the L578 handler upward. That breaks: its `except` builds an audit record from `decision`, `approval_fingerprint`, and `rights_status`, and `decision` is not bound yet in the stretch you flagged, so the handler would raise `UnboundLocalError` and re-strand the row with a worse traceback. The gate stretch gets its **own** handler whose audit record references only names bound before it.

On `_verify_rich_publisher_evidence`: broadened to `except Exception` + `logger.exception`. Worth stating plainly since a broad catch on a security gate reads wrong — it is **fail-closed**. The return is still `False`, verification fails, and the caller's gate blocks the record. A buggy verifier can only ever refuse ingestion, never admit an unverified record.

## 4. Registry race

Confirmed empirically, and it is a shade worse than described: `_initialized` is set to `True` *before* `_discover_all_components()` runs, so the second thread does not merely race — it returns a registry whose component tuples are provably empty. The revert-check test fails with `AssertionError: () is not true`.

Construction and `reset_registry()` now hold a class-level `RLock`. `_initialized` stays set before discovery on purpose: an `RLock` plus that ordering keeps same-thread re-entrancy (a pack module reaching back into `get_registry()` at import time) behaving exactly as it does today, while cross-thread callers block. Marked `@pytest.mark.serial` — it resets a process-global singleton, and this branch already ate one global-singleton deadlock under `--dist loadscope` (`6c82cc1f5`).

Not fixed: stale `sys.modules` entries on uninstall. That is a real leak but it needs an uninstall path to hook, which this PR does not have.

## Verification

`test_authority_pack{,_api,_sideload,_providers}` · `test_authority_gate` · `test_authority_discovery` · `test_crawl_authorities{,_security}` · `test_authority_frontier` · `test_pipeline_registry` · `test_pipeline_hardening` · `test_schema_parity` · `test_graphql_service_layer` — all green. Pre-commit clean. Each of the five new tests confirmed exit 1 reverted / exit 0 with the fix, failing for the right reason each time (`RuntimeError: relink exploded` escaping `install`, raw `IntegrityError`, `AttributeError` escaping the gate, `RuntimeError` escaping `discover_and_bootstrap`, empty parser tuple).

## Not addressed

The smaller gaps, frontend, and style items, as with round two: the `tool_call_log`-keyed step-budget notice (already disclosed and pinned by its own known-gap test), `terminal_reason` as a column/field rather than warning text, the per-corpus row cap in `_citable_passage_rows_across`, the `_attach_publisher_original_file` partial-state contract, double `record.text` extraction, the `MAX_EXTRA_CA_CERTIFICATE_BYTES` enforcement gap, the re-embed warning on empty corpora, the save-popover dismiss, the `PackPreflightModal` duplicated boolean, and the magic-number/`link_only`-comment/duplicate-title-ambiguity nits. Happy to take any of them here rather than as follow-ups.

JSv4:
Follow-up `7e4a6f828` — one gap in the fix above, found by checking the actual operator surface rather than assuming it.

My claim that the post-commit warnings "ride the message field the Console already renders" was true but insufficient. `PackPreflightModal.tsx:256` renders `result.message` through **`toast.success`** — and the whole point of fix #1 is that the server keeps `ok: true`, because the pack really is installed. So the operator got a green toast ending in "…re-linking existing corpora to its authorities failed", which reads as all-clear. The warning was visible and simultaneously not.

It now branches on the structured `result.warnings` already in the payload and toasts a warning. `InstallAuthorityPackOutputs.result` is typed for the keys the client reads rather than `unknown`. New CT `warns instead of celebrating when the install returns post-commit warnings` asserts a `Toastify__toast--warning` containing the relink text and **zero** success toasts — verified to fail with the branch reverted.

Full authority-pack CT suite (24 tests) green, `tsc --noEmit` clean, prettier clean.

JSv4:
Pushed `c1331835c`. The high finding and every actionable medium/low is fixed; two findings are answered rather than changed, and one is a false positive with pointers below.

## The blank heading — confirmed, and the masking explanation was incomplete

`ObligationCard` only ever serialises `responsible_party` (the stored card is its `model_dump()`), so `card.owed_by` read `undefined` and every real obligation card rendered an empty `<h4>`. Fixed.

The diagnosis of *why the tests missed it* needs one correction, and it changes the fix. The mocks carrying `owed_by` were not sufficient to hide this — **no test asserted either card's heading at all.** Renaming the mock field alone would have left the suite just as blind to the next rename. So:

- heading assertions added on **both** card shapes (`finding-heading`),
- `CardFieldContractTests` (`test_finding_cards.py`) pins the **full** field-name set of `ObligationCard` and `RegimeCard`, and separately pins `model_dump()`'s keys, next to a comment naming `ResearchFindingsEmbed.tsx` as the file that must change with them. A whole-set comparison rather than a spot-check of the fields the embed reads today, because the embed grows.

Verified by reverting: with `card.owed_by` restored the obligation CT test fails.

Same drift, one more place: the agent-facing "either/or" error in `research_tasks.py` named `owed_by` as an OBLIGATION-card parameter — a kwarg `_build_finding_card` does not accept. Corrected.

**One thing you didn't flag that belonged with it.** The embed also dropped `obligor_grounded`, so an obligation whose obligor the cited passages never name rendered identically to one they do — the backend marks that case precisely so a reader can see it. Marked now; the screenshot in this commit shows both.

## `finalize_once` — fixed, and it is worse than a cancellation race

Guards on `is_terminal` now, not `status == COMPLETED`. Worth stating the concrete path, because it is not only the cancel button: `reap_stalled_research` resumes a stalled RUNNING report, so two workers can be live on one row. Worker A's soft time limit fires → `mark_cancelled`; worker B's salvage then finalized on top, producing a **COMPLETED report still carrying A's `error_message`**, after the user had already been notified the run was cancelled. Checked before widening that nothing resets a report out of a terminal state (`resume` refuses one outright), so this cannot strand a run that ought to finish.

## `budget_exhausted` — deleted, not just redundant

Sharper than "the old ambiguity partially survives". Warnings are rendered **verbatim** to the user (`ResearchReportDetail.tsx:830`), and nothing anywhere reads the string. So a run that blew the *step* budget — or one that simply stopped — displayed a chip claiming the token budget was gone, directly beside the `terminal_reason` saying otherwise. Removed; `terminal_reason` is the only warning the salvage path emits.

## Salvage integration test — added

`SalvagePathIntegrationTests` drives `_run_deep_research_async` end to end with a fake agent returning the step-limit metadata: asserts the composed body carries the salvage note *and* every recorded finding, that exactly one warning is written and it names the right budget, plus the two refusal cases above. All four fail with the fixes reverted.

One trap worth recording: `agents` is the `AgentAPI` singleton from `llms/api.py`, not the `llms.agents` package — patching the module path leaves the real factory in place and puts a live request on the wire (my first run took a 401 from api.openai.com and salvaged it into a `ModelHTTPError` terminal reason). Patch `AgentAPI.for_corpus`. The class is `TransactionTestCase`; a `TestCase`'s outer atomic does not survive the coroutine's `sync_to_async` hops.

## Fixed nits

| Finding | Fix |
|---|---|
| Unanchored `requires_responses_api` prefix | Matches whole name or up to `-` / `.`, so `gpt-5.6` no longer prefixes a hypothetical `gpt-5.60-…` |
| 3× duplicated `openai-responses:` expression | Hoisted to one `env_spec` |
| No rate limit on `installAuthorityPack` | `@graphql_ratelimit(RateLimits.ADMIN_OPERATION)`, via the decorated-inner-resolver split `annotation_queries.py` already uses. Schema parity green |
| `infolist()` rescan per tagged document | Member census computed once per open ZIP and memoised on it. (`NameToInfo` cannot serve this — it keeps only the last entry for a repeated name, and spotting a repeat is the point of the check) |
| Two new `1024*1024` literals | `BLOB_HASH_CHUNK_BYTES` in `constants/document_processing.py` |
| `_safe_segment` untested for `folder_name` | Two tests: separators/`..` in a folder name cannot invent a level (asserted on the `CorpusFolder.name` too, since it is written there as well as into the path), and a name that sanitizes to nothing falls back rather than silently landing in the corpus root |

## False positive — `versioning.py` metadata branches

> the new `skip_if_unchanged` logic's `metadata_updated`/`record_metadata_event=True` branches … aren't covered by `test_document_versioning.py`; only the "unchanged" path is tested

They are covered — in `opencontractserver/tests/test_authority_ingestion_invariants.py`, a 718-line module this PR adds alongside the versioning change, not in `test_document_versioning.py`:

- `test_metadata_change_creates_same_version_path_event` — `record_metadata_event=True`: new `DocumentPath` at the **same** `version_number`, `parent_id` chained, old path `is_current=False`, document + `ingestion_metadata` updated.
- `test_metadata_change_can_update_in_place_without_an_audit_event` — `record_metadata_event=False`: same path row updated in place, `DocumentPath.objects.count()` stays 1.
- `test_dedupe_does_not_downgrade_public_visibility` — the `is_public` invariant inside that branch.
- `test_opt_in_dedupe_does_not_change_default_upload_versioning` — the opt-in boundary.

The `SELECT FOR UPDATE` guarantee itself is genuinely not tested; that needs a threaded test and is not new behaviour in this PR.

## Answered, not changed

**`_reconcile_imported_authority_metadata` — both guards you asked me to confirm already exist**, and `replace=True` cannot be reached by a normal corpus owner:

1. `is_authority_admin(user_obj)` — a non-admin returns early with a WARNING, before any relationship write.
2. `allowed_prefixes` — the canonical key's prefix must already be bound by an `AuthorityNamespace` **whose `authority_corpus` is this exact target corpus**. Empty set → return. So a zip cannot mint a global authority edge by putting a `canonical_key` in `custom_meta`; a trusted pack install must first have bound that prefix to that corpus.

`replace=True` is scoped to keys that clear both. That it rides an ordinary import rather than a dedicated "install pack" action is the deliberate part — a reimport of a pack-bound corpus has to converge its edges, and dropping to additive-only would leave retired relationships permanently — but it is worth an explicit action later; I'm not changing that shape in this PR.

**SSRF consolidation — declining the merge as stated.** The two functions do not have the same contract: `host_on_allowlist` resolves the *registered* effective allowlist (baseline ∪ installed packs) and normalises with `.lower()`; `host_matches_declared_sources` takes an *explicit* host iterable from one provider and normalises with `.casefold()`. Only the innermost exact-or-dot-suffix predicate is shared. A naive merge changes the allowlist-resolution semantics of one caller; extracting just the predicate is worth doing but is a security-boundary refactor and belongs on its own. Tracking separately.

**`ObligationCard` cross-field validation — not implementable as described.** `RegimeCard` gets an anachronism check because a date *contradiction* is derivable from the data (an authority effective after the day it purports to govern). "The right value in the wrong role field" is not: nothing in `preparer="Oncor"` distinguishes a correct value from one that belonged in `submitter`. What is checkable is already checked — `distinct_roles()` refuses a role that merely restates `responsible_party`, `party_named_in_passages` grounds the obligor against its own evidence, and `finalize` flags approval/effective conflation.

**Zip entry-count cap on the V2 import path** — `validate_zip_for_import` currently runs on the bulk-upload path only. Applying it to corpus-export import is a behaviour change on a security boundary and would reject archives that import today; not folding it into this PR. The O(n·m) half is fixed above.

## Verification

`c1331835c`: 850+ backend tests across the touched modules pass (research, finding cards, workspace, authority ingestion/pack/API, import/export v2/v3, validate_export, model factory, schema parity, architecture invariants) · 4 CT tests pass · every new correctness test verified to fail with its fix reverted · `tsc --noEmit` clean · `yarn lint` clean · pre-commit green · `manage.py check` clean (E001) · 2 changelog fragments validated · the obligation-card doc screenshot regenerated, and it is the visual proof — headings were blank before.

JSv4:
Follow-up `50589e15a`. **Correction to my previous comment:** it claimed every actionable low was fixed. One was not — I dropped it silently, which is worse than deferring it.

**The dropped finding.** "No test instantiates `OpenAIResponsesModel` through the DB-credentialed `_construct_model` path end-to-end." Correct, and it mattered more than the "low" label suggests: the existing tests either mock `_construct_model` or assert the *env-credential string* redirect, so nothing reached the branch that decides which pydantic-ai class a credentialed install actually gets — and the class **is** the endpoint. A `gpt-5.6` model built as an `OpenAIChatModel` 400s every agent in the install with the reason visible only in a worker log, which is the failure this whole routing exists to prevent. Now asserted end to end, with the negative half (`gpt-4o` must **not** become a Responses model, or every ordinary agent moves endpoint too).

While there: the anchoring change was the only correctness fix in `c1331835c` without a test of its own. `TestRequiresResponsesApiAnchoring` pins it — exact name, `-` and `.` suffixed variants, `gpt-5.60` / `gpt-5.60-turbo` rejected, and another provider never redirected. Verified to fail with the anchoring reverted.

**Two verification gaps my own fixes opened**, neither of which the earlier test runs would have caught:

- `_member_name_counts` replaced duplicate detection *inside a security check*, and the existing tests covered only `occurrences == 0` (missing member). The `> 1` side is exactly what `NameToInfo` cannot serve — it keeps only the last entry for a repeated name — so I had refactored a half-tested guard and confirmed only that valid archives still import. A ZIP with two entries of the same name and different bytes now pins the rejection.
- `SalvagePathIntegrationTests` is the first `TransactionTestCase` in a module that was otherwise `TestCase`/`SimpleTestCase`, and this PR's own history includes a 600s deadlock traced to the `PipelineSettings` reseed on a `TransactionTestCase` flush — which I can see firing in these tests' logs. Ran `opencontractserver/tests/research/ -n 2 --dist loadscope`: exit 0, no wedge.

**One more instance of the `budget_exhausted` mislabel**, found on re-read rather than reported: `_compose_salvage_body`'s zero-findings fallback still read `"_No findings recorded before the research budget was exhausted._"`. Same wrong claim, one layer below the warning I removed, on a body composed for *every* non-finalize ending. Now generic; the specific cause stays in the `terminal_reason` in the executive summary it sits beside.

**Two stale references I am knowingly leaving**, so they are on the record rather than discovered later: `research/models.py`'s `warnings` field `help_text` still cites `['budget_exhausted']` as its example, and `test_research_report_service.py` uses the string as an arbitrary warning payload. Changing the `help_text` costs a migration for a Django-admin-only string; not worth the churn or the conflict surface in this PR.

**Verification.** `test_llm_model_factory.py` + `test_import_v2_reingest_remap.py` + `test_research_step_budget.py` pass (78) · research suite green under xdist · anchoring and duplicate-member tests verified to fail with their fixes reverted · pre-commit green · changelog fragments extended and validated.

PR #2227, opened 2026-08-04
Add California foreclosure compliance corpus analyzer

SamSammane:
Runs the legalis-ca-foreclosure ruleset (Civ. Code § 2924 et seq.) over a corpus of recorded instruments and stores the compliance chronology on the Analysis.

A foreclosure matter maps onto a corpus: each recorded instrument is a document. This is corpus-scoped rather than per-document because no single instrument answers the question — whether the three-month period under § 2924(a)(2) elapsed needs the Notice of Default and the Notice of Sale together.

The ruleset runs as a sidecar rather than in-process. It is Rust; an FFI boundary would couple this deployment to a Rust toolchain and turn the ruleset's panics into Celery worker crashes.

What is read from documents: instrument type, recording date, date mailed to the trustor, publication and posting dates. What is not read, because it appears on no recorded instrument: the sale date, loan purpose, occupancy, tenders, postponements, payoff requests. Those come from the analyzer input or Corpus.custom_meta, and anything absent reports INSUFFICIENT RECORD rather than being assumed — a finding resting on an assumed fact is worse than no finding.

Instrument classification consults the document title before the body, since a Notice of Default recites the deed of trust it secures and a whole-text scan picks the wrong instrument.

If the service is unreachable the task raises rather than returning "no violations found"; a compliance analyzer that reports a clean bill of health because it could not reach its ruleset is worse than one that crashes, and a test asserts this. The task is idempotent, as CELERY_TASK_ACKS_LATE requires.

Adds the foreclosure-api sidecar behind a compose profile, FORECLOSURE_API_URL and FORECLOSURE_API_TIMEOUT settings, and docs. 32 tests.

Claude-Session: https://claude.ai/code/session_01QmjyGXUNXHKZETogxWuDdu

### DIFF
--- a/config/settings/base.py
+++ b/config/settings/base.py
@@ -1832,3 +1832,18 @@ MCP_CORS_ALLOWED_ORIGINS = env.list(
 # Host; because MCP bypasses ALLOWED_HOSTS, pinning this in production is
 # recommended (e.g. MCP_PUBLIC_BASE_URL=https://contracts.opensource.legal).
 MCP_PUBLIC_BASE_URL = env.str("MCP_PUBLIC_BASE_URL", default="")
+
+# ------------------------------------------------------------------------------
+# FORECLOSURE COMPLIANCE RULESET
+# ------------------------------------------------------------------------------
+# The California foreclosure ruleset (Civ. Code § 2924 et seq.) runs as a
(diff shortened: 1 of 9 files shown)


JSv4:
Hey! Cool contribution, thank you. Will take a look

PR #2229, opened 2026-08-08
Document authority tagging and the modular authority-pack system

JSv4:
The pack machinery landed in #2215 with its docs partly describing the tree as it was before the packs moved out, and with screenshots the component tests capture on every run referenced by no document at all. Docs only — no code changes.

## Screenshots were being captured and never used

`AuthorityPackPreflightCoverage.ct.tsx` calls `docScreenshot` 14 times. Thirteen of those images are on disk under `docs/assets/images/screenshots/auto/` (the fourteenth, `--install-post-commit-warning`, has never been generated — the screenshot workflow is on-demand and hasn't run since that test was added), and **every one of them was referenced by no markdown file**. The workflow has been regenerating pictures nobody could see.

This PR links the five that teach a reader something. The other eight stay unlinked deliberately: loading spinners, error banners, a near-duplicate of the summary modal, and `--import-unbridged`, whose disabled "not connected in this view" state belongs to the test harness rather than the product. CT tests capture states for *coverage*; docs use the ones that carry meaning, and padding a page with spinners would make it worse, not more complete.

## Authority Console — the Authority Packs tab

The tab that installs a pack was four paragraphs with no visuals. It now shows:

- the **catalog**, and the status vocabulary a card's badge is derived from (`PacksTab.tsx::packStatus` — derived, not stored);
- the **empty catalog**, which is the normal symptom of a pack directory never bind-mounted into the container, since the path resolves inside it;
- the **preflight modal**: the `expectedFingerprint` round-trip that rejects a pack edited on disk between preflight and install, the publish opt-in's charter gate, the refusal to close mid-install, and `load_authority_pack --check` as the headless equivalent for a deployment with no admin browser session;
- a **rejected install** — the slug-collision refusal the authoring guide already described in prose.

## Pack authoring guide

Three real gaps, not just polish:

- **`sources.yaml` and the out-of-process collector** were documented only in `scripts/authority_import/README.md`, which no doc linked to. That's the whole path for a deployment whose crawling happens elsewhere; it's now in the anatomy and slot table, pointing at the builder README rather than duplicating it.
- **The in-pack import contract.** Relative imports resolve wherever a pack is mounted; an absolute in-tree dotted path silently fails to register the whole module's providers once sideloaded — the defect that stranded 15 providers. This is the one rule an author *cannot* discover from a pack that still lives in the tree, so it now sits next to the provider instructions.
- **Two worked examples**, because neither alone shows the format: v1 `bolivia` (shipped, installable) and the schema-v2 fixture `tests/fixtures/authority_packs/example_utility` — plus why the fixture lives under `tests/fixtures/` and not the pack root.

Also: the `pack.yaml`-required root scan (why a root pointed one level too deep contributes nothing, and why explicit `AUTHORITY_PACK_PATHS` entries are deliberately not filtered that way), `E2E_RUN_AUTHORITY_IMPORTS`, and two references to `test_grid_dossier_*` modules that left the tree with the packs — replaced with the modules that exist.

## Reference-web enrichment

The detection-tier table never said where Tier-1's vocabulary comes from, which is the link between the tagging engine and the packs that extend it: `prefixes`/`aliases` plus a pack's optional `shape_rules` / `abbreviations` merge onto the shipped baseline at runtime, so standing up detection for a new body of law is a data change. Now stated, with pointers both ways.

## Verification

- Every image and relative markdown link in the touched files resolves on disk.
- `scripts/collate_changelog.py --check` passes (301 fragments).
- Applicable pre-commit hooks for markdown (`trailing-whitespace`, `end-of-file-fixer`) verified by hand; `pre-commit` is not installed in this container.

https://claude.ai/code/session_01Eos5eGFoKZLqcpFGjM1gXE

### DIFF
--- a/changelog.d/authority-tagging-docs.changed.md
+++ b/changelog.d/authority-tagging-docs.changed.md
@@ -1,0 +1,31 @@
+- **Documented the authority tagging system and authority packs against what
+  #2215 actually shipped**, and wired the pack screenshots the component tests
+  have been capturing into the docs that describe them.
+  - `docs/architecture/authority-console.md` — the **Authority Packs tab**
+    section was four paragraphs with no visuals despite nine
+    `admin--authority-packs-tab--*` / `admin--pack-preflight-modal--*`
+    screenshots being captured on every screenshot run and referenced by no
+    doc. It now shows the catalog, the derived status-badge vocabulary
+    (`Invalid` / `Available` / `Partially installed` / `Installed privately` /
+    `Partially public` / `Fully public`, per `PacksTab.tsx::packStatus`), the
+    empty catalog as the symptom of an unmounted pack directory, and the
+    preflight modal — including the `expectedFingerprint` round-trip that
+    rejects a pack edited between preflight and install, the publish opt-in's
+    charter gate, and the `load_authority_pack --check` headless equivalent.
+  - `docs/guides/authoring-authority-packs.md` —
+    documents the previously undocumented **`sources.yaml` source plan** and the
+    out-of-process collector (`scripts/authority_import/`) that turns it into
+    GUI-importable corpus ZIPs; adds the **in-pack import contract** (relative
+    imports resolve wherever a pack is mounted, an absolute in-tree dotted path
+    silently fails to register the module's providers once sideloaded — the
+    defect that stranded 15 providers); points at the schema-v2 fixture pack
+    `tests/fixtures/authority_packs/example_utility` as the complete worked
+    example alongside v1 `bolivia`; explains the `pack.yaml`-required root scan
+    and `E2E_RUN_AUTHORITY_IMPORTS`; and replaces two dead
+    `test_grid_dossier_*` test references (those modules moved out of the tree
+    with the packs) with the modules that exist.
+  - `docs/architecture/reference-web-enrichment.md` — states where the Tier-1
+    detection vocabulary comes from, closing the gap between the tagging engine
+    and the packs that extend it (`prefixes`/`aliases` plus a pack's optional
+    `shape_rules` / `abbreviations`), so standing up detection for a new body of
+    law reads as the data change it is.
--- a/changelog.d/authority-tagging-docs.changed.md
+++ b/changelog.d/authority-tagging-docs.changed.md
@@ -2,10 +2,13 @@
   #2215 actually shipped**, and wired the pack screenshots the component tests
   have been capturing into the docs that describe them.
   - `docs/architecture/authority-console.md` — the **Authority Packs tab**
-    section was four paragraphs with no visuals despite nine
+    section was four paragraphs with no visuals, while thirteen
     `admin--authority-packs-tab--*` / `admin--pack-preflight-modal--*`
-    screenshots being captured on every screenshot run and referenced by no
-    doc. It now shows the catalog, the derived status-badge vocabulary
+    screenshots sat in the auto directory referenced by no doc. It links the
+    five that teach something; the rest are loading spinners, error banners and
+    near-duplicates that `AuthorityPackPreflightCoverage.ct.tsx` captures for
+    coverage, which is not the same job as documentation. It now shows the
+    catalog, the derived status-badge vocabulary
     (`Invalid` / `Available` / `Partially installed` / `Installed privately` /
     `Partially public` / `Fully public`, per `PacksTab.tsx::packStatus`), the
     empty catalog as the symptom of an unmounted pack directory, and the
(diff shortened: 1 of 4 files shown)


JSv4:
Thanks — the count nit is a real error and it was in the changelog fragment too, not just the PR description. Fixed in 19b8f39.

`AuthorityPackPreflightCoverage.ct.tsx` has **14** `docScreenshot` call sites and 14 distinct names; **13** are on disk. The fourteenth, `admin--pack-preflight-modal--install-post-commit-warning`, has never been generated — the screenshot workflow is `workflow_dispatch`-only and hasn't run since that test was added. "Nine" was the number left unlinked *after* this change, which I wrote as though it were the number captured.

On the substantive half — the gap not being fully closed — that part is deliberate, and I've made the reasoning explicit in the fragment and PR body rather than leaving it to look like an oversight. I've added one more on merit: `--install-rejected` shows the slug-collision refusal the authoring guide already describes in prose, so the picture and the paragraph now reinforce each other. That brings it to five linked.

The remaining eight stay unlinked:

- `--loading` (×2), `--catalog-error`, `--preflight-error` — spinners and error banners; the prose already says what happens, and a picture of a spinner adds nothing a reader can act on.
- `--approval-tones` — visually near-identical to `--summary-badges` (same modal, cut at the same fold); two of them would be padding.
- `--blank-fields` — a fallback-text state that documents the component's defensive rendering, not a state an operator reaches.
- `--close-blocked-while-installing` — real product behavior, but it's one sentence of prose and a screenshot of an unchanged modal doesn't show the "didn't close" part.
- `--import-unbridged` — the disabled button and its "Corpus ZIP import is not connected in this view." tooltip are the test wrapper's, not the product's. Documenting it would teach a fiction.

Component tests capture states for coverage; docs use the ones that carry meaning. I'd rather have eight images linked by nothing than eight paragraphs padded to justify them — so I'd treat "every auto screenshot is referenced" as a non-goal. Worth noting the same pattern exists well beyond this feature: ~380 of the images in `auto/` are referenced by no doc, so if that *is* meant to be an invariant it wants a lint rule and its own change, not this one.

---

JSv4:
Both nits fixed in 3dc159e.

**Nit 1** — fixture path in the changelog fragment is now repo-root relative like every other path in it.

**Nit 2** turned out tighter than suggested, and checking it was worthwhile. The label isn't an independent axis: `AuthorityPackService` sets `can_install=True` on the valid path (`authority_pack_service.py:413`) and `can_install=False` on the invalid one (`:597`), and those are its only two assignments — so `can_install` tracks `valid` exactly. Combined with `disabled={!pack.valid}` on the same button, **"Review pack" only ever renders on a disabled button.** There is no reachable state where you can click through to a preflight for a pack that can't be installed.

That makes it more than a label caveat: an invalid pack cannot be opened at all, which is worth knowing when a sideloaded pack shows up broken and an operator goes looking for a diagnostic view that isn't there. So it's documented on the `Invalid` badge row — where a reader actually meets that state — rather than as a parenthetical on the install flow:

> `Invalid` — The manifest failed server validation. The card shows the error, and its button reads **Review pack** and is *disabled* ("Repair this server-deployed pack before installing it") — an invalid pack cannot even be opened, since `can_install` tracks `valid` exactly.

The guide gets the one-clause version, pointing at repairing on disk, since `load_authority_pack --check` is the diagnostic there.

**Nit 3** needs no action, agreed — `--install-post-commit-warning` is referenced nowhere, and the PR description is the only place it's mentioned, correctly as never-generated.

---

JSv4:
Thanks — and good trace on the `install-and-publish` fixture. Agreed on the diagnosis: `installedTarget` (`AuthorityPackPreflightCoverage.ct.tsx:787-791`) omits `isPublic`/`fullyPublic`, so the card renders `Installed privately` / `Public 0/2` even though the test checks the publish option and asserts the "installed and published" toast. The image disagrees with its own test's name.

I've left the fixture alone — it's outside a docs PR's scope, and more practically I can't regenerate the PNG from here (the screenshot workflow is `workflow_dispatch`-only), so editing the mock would fix the test while leaving the committed image unchanged and the doc pointing at a picture that still didn't match. Worth a follow-up for whoever next touches that coverage file; the fix is adding the two fields to `installedTarget` and re-running the workflow.

What I did do in 9b21aa4 is harden the doc against it. The caption sat a few lines under the publish opt-in discussion, so a reader could reasonably wonder why publishing hadn't taken effect in the picture. It now names the state instead of leaving it implicit:

> After a successful install the modal closes and the card flips to its new state — below, a pack installed privately, so both corpora exist but neither is published (`Installed 2/2`, `Public 0/2`).

That's true of the image as it stands today, and stays true if the fixture is later corrected and a genuinely-published card replaces it — at which point the caption should be updated alongside. Documenting what the picture shows rather than what the file is named is the safer default when the two disagree.

---

PR #2232, opened 2026-08-09
The Docx Arcade: playable ASCII games inside a live Word document (embeddable demo)

JSv4:
## What this is

The next step past the [Docx Observatory](https://jsv4.github.io/Docxodus/demo/observatory.html): a fully embeddable, fully client-side demo where **two playable games run inside a live Word document**, hosted by the stock Docxodus ribbon editor — no separate rendering machinery, no backend, no build step, and **no Docxodus API additions needed** (the existing `replaceXml` / `refresh()` / agentic-session surface covered everything).

- **Courier Quest** — a single-screen platformer in 8pt Courier New. You are `@`, a process server: collect every `§`, stomp the pilcrows `¶`, unseal the `!` exit. Spikes, one-way shelves, coyote time, stomp bounces, three rooms.
- **Docx Dungeon** — a first-person raycaster. Distance-banded ASCII shading, pilcrow ghosts with line-of-sight chase, and an `E` door to serve.

The hook that makes it more than a screensaver: **the document is not just the screen — it's the level source.** The screen is one Word paragraph redrawn ~12×/s; the rooms and the dungeon floor plan are ordinary monospace paragraphs further down the same document. Click into one and type while the game runs:

- in the platformer, any unrecognized character is solid scenery — **typing a word builds a ledge you can stand on**;
- in the raycaster, **letters you type into the floor plan rise as 3D walls textured with themselves** (the default map ships with D-O-C-X founders' pillars).

And because it is only a document: Undo rewinds frame by frame, and the ribbon's **Save** downloads whatever is on screen — mid-jump included — as a valid `.docx`.

Courier Quest
Typing a word into ROOM 1 while it runs
Docx Dungeon

## Where it lives / how it ships

- `docs/demo/docx-arcade/` — static host page + four ES modules, deployed **verbatim** by mkdocs (verified with a full `mkdocs build`: files copy byte-identical; the `minify` plugin only touches rendered pages). After the next `mkdocs gh-deploy` it's live at `…/OpenContracts/demo/docx-arcade/`.
- Engine pinned to `docxodus@9.7.0` via jsDelivr (same pin as the Observatory host), with an `?engine=` override; `?game=dungeon` and `?fps=` params for embedding.
- Embeds anywhere via iframe — the dock has an **Embed** button that copies the snippet. OG tags included for link unfurls.
- `docs/demo/index.md` — docs page under a new **Demos** nav tab, with controls, embed snippet, and pointers into the code.

## Design notes (the three hard-won details)

1. **Run budget.** Every color change inside a row is its own `w:r`, and runs dominate convert cost (the Observatory's economics). Both games shade in coarse distance/row bands and let glyphs carry texture: 40–90 runs/frame in practice, 10–16 fps with `editor.refresh()` doing an incremental single-block repaint every frame.
2. **Blur commits.** The editor commits typed text to the session only on blur, so the driver reads a *dirty* cartridge block's live DOM (`engine.js::domBlockToRows`) between commits — that's what makes walls rise keystroke by keystroke. Clicking the screen to resume play explicitly blurs the active block so the pending text commits to the model.
3. **NBSP blanks.** Blank level rows are seeded as non-breaking spaces: ordinary trailing spaces hang under `pre-wrap`, the browser refuses to place a caret in them, and empty sky becomes untypeable. NBSP renders identically and folds back to a space when the game reads the level.

## Input model

Two keyboard owners, decided by pointer placement: click the screen paragraph and the game takes WASD/arrows (capture-phase, default-prevented — the caret never eats a jump); click anywhere else and the document takes the keyboard back. `Esc` always returns it. Gamepads work (d-pad/stick + A to jump). A dock chip always shows who has the keys.

## Verification

No automated CI covers `docs/` demos, so this was verified with local Playwright runs against the real engine bundle (Chromium + mirrored `dist/wasm/_framework` runtime): 17 assertions covering boot, telemetry, both games' input/physics, live level editing through both the session API and real caret typing, blur-commit semantics, and console hygiene — all passing, plus a typing-while-running scenario. Manual test steps documented in `docs/test_scripts/docx_arcade_demo.md` per the repo's testing convention; changelog fragment in `changelog.d/`.

Nothing outside `docs/`, `changelog.d/`, and the `mkdocs.yml` nav entry is touched — no backend, GraphQL, or frontend-app surface.

---
_Generated by [Claude Code](https://claude.ai/code/session_01LbEcXVTGLYoNABhfQwXXeb)_


JSv4:
Applied in 865ce4d:

- **Dead ternary** in `createInput` (keydown + keyup) collapsed to `ev.key.toLowerCase()` — correct diagnosis, it was a leftover from an abandoned single-char/named-key split.
- **`lastReconcileFallback` → stats `innerHTML`**: now HTML-escaped before interpolation (`arcade.js::escHtml`).
- **`unidOf` selectors**: now `CSS.escape`d rather than trusting the library's id format.
- **Playwright harness committed** as `scripts/verify_docx_arcade.js` — 19 assertions including the blur-commit and DOM-overlay scenarios, with a built-in static server and env-configurable Chromium/engine mirror; `docs/test_scripts/docx_arcade_demo.md` points at it. Re-ran it against the fixed code: 21/21 (it adds a run-budget assertion and folds in the typing scenario).

**Left as-is: CDN SRI.** Dynamic `import()` has no native integrity support, this mirrors the existing Observatory host pattern (so no regression), and the version is pinned. A fetch+integrity-check+blob-URL shim or vendoring the pinned build is a real option but changes the loading architecture for both host pages — better tracked as its own issue if we want it.

---

JSv4:
All three addressed in 811bfba:

- **`?engine=` hardened** — the override now honors only same-origin URLs and `cdn.jsdelivr.net`; any other origin logs a warning and falls back to the pinned engine. Local-mirror testing (same-origin) and CDN version pins keep working, so the parameter's actual use cases survive. (The Observatory host page has the same pattern upstream; worth porting this there separately.)
- **Harness traversal check** — now a separator-suffixed prefix check, and the comment no longer overclaims.
- **Row-width magic number** — the solid-terrain assertion derives width from the live rows instead of hardcoding 92.

Re-ran `scripts/verify_docx_arcade.js` against the hardened loader: 21/21.

---

JSv4:
**Finding 1 (NBSP no-op) — checked at the byte level: the NBSP is real.** On the pushed branch:

```
$ git show origin/claude/ascii-game-demo-coqu8m:docs/demo/docx-arcade/engine.js | grep -c $'\xc2\xa0'
3
```

Three U+00A0 occurrences, in exactly the expected places and directions: `cartridgeXml` writes space→NBSP, `paraXmlToRows` and `domBlockToRows` fold NBSP→space. Your grep most likely ran against a diff rendering that normalizes NBSP to a plain space (several GitHub diff surfaces do). The behavior is also functionally pinned, not just byte-pinned: the harness's "typed word ingested LIVE" scenario clicks into a *blank* row of ROOM 1 and types — that assertion failed before the NBSP swap landed and passes since (it was bisected exactly this way during development).

**Finding 2 (astral/surrogate desync) — real edge case, hardened in e4db42a.** Both readers now share a `fitRows()` helper (which also deduplicates their pad/truncate tails) that maps every surrogate code unit to a `?` cell — solid scenery / a wall under the unknown-glyph rule — so a typed emoji becomes `??` terrain instead of a stranded lone surrogate reaching the frame writer's `w:t`. The harness gained two assertions (emoji edit doesn't halt the loop; `?`-walls appear in the raycaster grid): 23/23 locally.

**Finding 3** — agreed, no action: the demo is self-contained by design and the constants are named at their point of use.

---

JSv4:
All six addressed in 73ecd00:

1. **`?engine=` package-scoped** — good catch that origin-only still admitted any public npm package; the jsDelivr arm now also requires `pathname.startsWith("/npm/docxodus@")`. Same-origin mirrors unchanged.
2. **Doc drift** — the test script now describes the harness's coverage instead of a hard assertion count, so it can't drift again.
3. **Bidi regex** — single `stripBidiMarks` export in `engine.js`, used by both `domBlockToRows` and the dirty-block check in `arcade.js`.
4. **`fmtTime`** — one `engine.js` export shared by both cartridges.
5. **Dead `P`/`¶` branches** — removed; `isWall`/`isSolid` now document that `applySource` normalizes the grid before storage.
6. **`?fps=0`** — presence-checked rather than truthiness-checked; `0` now maps to unthrottled as documented.

Re-verified against the pinned `docxodus@9.7.0` runtime (mirrored bytes of the CDN build): 23/23.

---

PR #2252, opened 2026-08-16
Fix corpus/document agent instructions being silently discarded

### LINKED ISSUE #2247: Corpus/document agent instructions are silently discarded: temporal grounding consumes the `system_prompt is None` signal
## Summary

`Corpus.corpus_agent_instructions` (and the document-agent equivalent) are
**silently discarded on every agent**. The configured persona is stored, is
visible in the DB and the API, and is never sent to the model.

The agent runs with only the computed temporal-grounding preamble as its system
prompt.

## Root cause

Two pieces of correct-looking code that are order-dependent:

1. `llms/agents/agent_factory.py::_inject_temporal_grounding` (main, line 193):

```python
config.system_prompt = (config.system_prompt or "") + "\n".join(lines)
```

2. `llms/agents/core_agents.py`, in both
   `CoreDocumentAgentFactory.create_context` (line 1095) and
   `CoreCorpusAgentFactory.create_context` (line 1163):

```python
# Set default system prompt if not provided
if config.system_prompt is None:
    config.system_prompt = ...get_default_system_prompt(...)
```

`_inject_temporal_grounding` runs first and makes `system_prompt` non-`None`.
`system_prompt is None` is the signal for "the caller did not supply a prompt,
so use the configured default" — and the injection consumes that signal before
it is read. The default resolution never fires, so the persona is never applied.

The temporal block is short, so the result looks like a plausible system prompt
rather than an empty one, which is why this is easy to miss.

## Reproduction

Any corpus with `corpus_agent_instructions` set:

```python
from opencontractserver.llms import agents

corpus.corpus_agent_instructions = "You are a helpful assistant. ALWAYS say BANANA."
corpus.save()

agent = await agents.for_corpus(corpus=corpus.id, user_id=user.id,
                                streaming=False, persist=False)
print(len(agent.config.system_prompt))
print("BANANA" in agent.config.system_prompt)   # -> False
```

Observed on a corpus whose instructions are ~14k characters:

| | `len(config.system_prompt)` | persona present |
|---|---|---|
| before | 657 | no |
| after the patch below | 16,072 | yes |

657 characters is the temporal block alone.

## Impact

- Every corpus-level and document-level persona is inert. Anything relying on
  configured instructions — tone, refusal policy, tool-usage guidance, routing
  rules, output format — silently does nothing.
- It is invisible from outside: the value round-trips through the DB and API
  correctly, so verifying "the persona is set" passes while the agent never
  receives it.
- It makes persona changes unmeasurable. A/B-ing prompt revisions returns the
  same behaviour every time, which reads as "the model ignores the instruction"
  rather than "the instruction was never sent."
- Agents that need an explicit instruction to call a tool will not call it. A
  tool whose required arguments cannot be guessed (e.g. anything taking a group
  or collection identifier) becomes unreachable in practice, because the only
  place that identifier would have been supplied is the persona.

## Suggested fix

Keep the computed context out of `system_prompt` until the default has been
resolved, so the `is None` signal survives. Patch that works locally:

```python
# core_agents.py — AgentConfig
computed_preamble: Optional[str] = None

# agent_factory.py — _inject_temporal_grounding
config.computed_preamble = (config.computed_preamble or "") + "\n".join(lines)

# core_agents.py — after each `if config.system_prompt is None:` default resolution
_apply_computed_preamble(config)
```

```python
def _apply_computed_preamble(config: "AgentConfig") -> None:
    preamble = getattr(config, "computed_preamble", None)
    if preamble:
        config.system_prompt = (config.system_prompt or "") + preamble
        config.computed_preamble = None
```

This preserves the existing behaviour for callers that *do* pass an explicit
`system_prompt` (they still get the temporal block appended), and restores the
default-resolution path for callers that don't.

`_inject_corpus_memory` runs just before the temporal injection and should be
checked for the same pattern.

## Suggested regression test

Assert the configured instructions appear in the built prompt, rather than
asserting the field round-trips:

```python
async def test_corpus_agent_instructions_reach_the_system_prompt():
    corpus.corpus_agent_instructions = SENTINEL
    corpus.save()
    agent = await agents.for_corpus(corpus=corpus.id, ...)
    assert SENTINEL in agent.config.system_prompt
```

Happy to open a PR if the approach looks right.

JSv4:
Closes #2247.

Configured agent instructions never reached the model. The value round-tripped
through the database and the API correctly, so every check of the form "are the
instructions set?" passed while the agent ran without them.

## Root cause

Two order-dependent pieces, each correct-looking on its own.

`llms/agents/agent_factory.py::_inject_temporal_grounding`:

```python
config.system_prompt = (config.system_prompt or "") + "\n".join(lines)
```

`llms/agents/core_agents.py`, in both `CoreDocumentAgentFactory.create_context`
and `CoreCorpusAgentFactory.create_context`:

```python
# Set default system prompt if not provided
if config.system_prompt is None:
    config.system_prompt = ...get_default_system_prompt(...)
```

`system_prompt is None` is the signal for *"the caller supplied nothing, so use
the configured default"*. The injection runs first and consumes that signal
before it is read, so the default branch never runs.

The temporal block is short, so the result looks like a plausible system prompt
rather than an empty one — which is why this survived. Measured on a corpus
carrying ~14,000 characters of instructions:

| | `len(config.system_prompt)` | persona present |
|---|---|---|
| before | 657 | no |
| after | 16,072 | yes |

657 characters is the computed preamble alone.

## Impact

Every corpus-level and document-level persona was inert. Anything relying on
configured instructions — tone, refusal policy, output format, routing rules,
tool-usage guidance — silently did nothing.

Two consequences worth calling out because they mislead rather than merely
disappoint:

- **Prompt A/B testing returned identical results**, which reads as "the model
  ignores this instruction" rather than "the instruction was never sent."
- **Agents that need an explicit instruction to call a tool never called it.** A
  tool whose required arguments cannot be guessed (anything taking a group or
  collection identifier) is unreachable in practice, because the only place that
  identifier would be supplied is the persona.

## Fix

`AgentConfig` gains `computed_preamble`. `_inject_temporal_grounding` stashes
there instead of concatenating, and `_apply_computed_preamble()` appends it
**after** the default resolves — for both corpus and document agents. The "was a
prompt supplied?" signal now survives long enough to be read.

Behaviour for callers passing an explicit `system_prompt` is unchanged: it still
wins, and still receives the preamble.

`_inject_corpus_memory` runs immediately before the temporal injection and had
the identical bug: it wrote straight to `system_prompt` whenever a
`memory_enabled` corpus had memory content, which made `system_prompt`
non-`None` before default resolution ran — discarding
`corpus_agent_instructions` for every memory-enabled corpus. This PR now fixes
that too: both injectors stash into `computed_preamble` when no prompt has
been resolved yet.

## Tests

`opencontractserver/tests/test_agent_instructions_reach_prompt.py` — 6 tests,
asserting the **built prompt**, never the field. A test that reads back
`corpus_agent_instructions` would have passed throughout the entire period the
bug existed, which is precisely how it hid. Covers the corpus-agent path, the
document-agent path (`CoreDocumentAgentFactory.create_context` hits the same
`system_prompt is None` branch), and the memory-enabled path
(`_inject_corpus_memory`).

Verified to fail without the fix: the corpus/document/memory tests fail, and
the failure output is the bare temporal-grounding-only prompt. The remaining
tests cover behaviour that is unchanged (explicit prompt wins; no-instructions
still builds a prompt), so they pass either way — that is the intended split,
not a gap.

`TransactionTestCase` rather than `TestCase` because the prompt-building path
reaches the database from async context. Mocking the factory — as the existing
agent tests do — would skip the code this exists to cover.

`opencontractserver/tests/test_agent_memory.py::test_none_system_prompt_handled`
asserted the pre-fix behaviour (memory injection setting `system_prompt`
directly when it was `None`); renamed to
`test_none_system_prompt_stashes_to_computed_preamble` and re-asserted against
the corrected stash-to-`computed_preamble` behaviour.

Existing `test_agent_api`, `test_core_agents`, `test_agent_temporal_grounding`,
and `test_llm_runtime_retargeting` suites unaffected.


JSv4:
Closed in favor of #2248 (merged), which independently fixed the same bug (issue #2247) with a cleaner design (`AgentConfig.computed_context` queue + `resolve_system_prompt()` vs. this PR's single-string `computed_preamble` stash) and stronger test coverage from the start, including the memory-enabled-corpus path this PR only fixed after review.

Before merging #2248, I diffed both PRs end to end and ported the one genuine coverage gap #2248 had relative to this one — a full-stack E2E test combining a real `memory_enabled=True` corpus, actual persisted memory content, and `corpus_agent_instructions`, driven through `agents.for_corpus()` rather than by calling `_inject_corpus_memory` directly — plus added a document-agent equivalent of #2248's `SystemPromptThroughPublicFactoryTestCase` (it only had corpus-agent E2E coverage through `UnifiedAgentFactory`, not document-agent). Both landed in #2248 before merge.

Everything else this PR touched (temporal-grounding stash, corpus-memory stash, document/corpus persona tests, changelog fragment) has an equivalent or better counterpart already in #2248.

PR #2261, opened 2026-08-18
perf(vector-stores): stop paying for debug diagnostics that are discarded

JSv4:
## The bug

Python evaluates a logging call's **arguments** before it checks whether the
level is enabled. Every diagnostic of this shape in `core_vector_stores.py`

```python
_logger.debug(await _safe_queryset_info(queryset, "After vector search"))
```

therefore runs `_safe_queryset_info` — a `COUNT(*)` over the annotation table —
on every search and then discards the string. Four of the sites were
`_logger.info`, so they ran the count *and* emitted it in production.

## Measurement

On a 4,679-section authority deployment (18-corpus `CorpusGroup`, warm caches,
no reranker configured), timing one member-corpus search by phase:

| phase | before |
|---|---|
| `_agenerate_query_embedding` | 0.291s |
| **`_build_base_queryset`** | **12.887s** — three `COUNT(*)` |
| `search_by_embedding` | 2.222s |
| `async_search` (whole) | 16.548s |

**78% of a search was diagnostics.** The single most expensive one is
unfiltered — it counts the entire table:

```
Initial: Total annotations in DB: 4591881 results
```

…once per search, to log a number that changes only when someone uploads a
document.

## After

| | before | after |
|---|---|---|
| one corpus search (warm) | 16.548s | **2.579s** |
| 18-corpus `search_across_corpora` | 277.4s | **45.3s** |

6.1× on the fan-out, and it applies to every corpus search in the product.

## The change

Guards every `_safe_queryset_info` call site and every `queryset.query`
stringification (compiling the SQL is not free either) behind
`_diagnostics_enabled()`, and demotes the four `INFO` sites to `DEBUG`. When
DEBUG logging is on, behaviour is unchanged.

I grepped for every other value reaching the same primitive rather than fixing
only the ones the profile named — that turned up the `queryset.query` f-string
at the metadata-filter site, which the first pass missed.

## Verification

`test/verify_diag_guard.py` (in the ITAR workspace, not this repo) asserts the
two things that would catch a guard swallowing real work:

```
corpus 127, 8 hits
  diagnostics ON :   6 guarded lines emitted
  diagnostics OFF:   0 guarded lines emitted
OK — identical results (8 hits), diagnostics deferred not deleted
```

i.e. the diagnostics still emit at DEBUG (deferred, not deleted), and the
returned annotation ids and scores are **identical** with the logger at DEBUG
and at INFO.

The repo suite was **not** run: under `local.yml` pytest dies at
`django.setup()` on a duplicate `storages` app label, before any of this module
is imported, and standing up the `test.yml` stack risks the compose
project-name collision while a long measurement was in flight. Worth a CI run.

## Not included

While measuring this I also found that the `asyncio.gather` in
`search_across_corpora` is **not concurrent** — 45.3s gathered vs 48.8s
sequential, a 1.08× speedup, because the paths below it await
`sync_to_async(..., thread_sensitive=True)`. A fix for that was attempted,
preserved results on all 18 corpora, delivered only 2.26× rather than the
predicted 18×, and then failed on all 18 when the same wrapper was awaited
sequentially after a gather (`AsyncToSync` finding a running loop in a worker
thread). It was reverted rather than shipped and is not part of this PR.


JSv4:
Swept the rest of the codebase for the same primitive rather than only the sites the profile named:

```
grep -rn 'logger\.\(debug\|info\)(.*\.count()' --include=*.py opencontractserver/ config/
grep -rn 'logger\.\(debug\|info\)(f\?".*{.*\.query}' --include=*.py opencontractserver/ config/
```

Everything else that eagerly evaluates a `COUNT(*)` inside a logging call is in `opencontractserver/tests/` (runs once, harmless) plus `utils/analyzer.py:218`, which is once per corpus-analysis rather than once per search. No `.query` stringification in a log call outside the file changed here. So the vector store was the only hot-path instance of the class.

JSv4:
Added `opencontractserver/tests/test_vector_store_diagnostics.py` so CI checks this rather than taking my measurement on trust.

**The regression test**: `test_no_count_diagnostics_run_when_debug_is_off` patches `_safe_queryset_info` / `_safe_queryset_info_sync` and asserts they are **not called** at INFO — i.e. that the logging call's arguments are no longer evaluated before the level is checked. Verified it fails on `main` and passes here:

```
# with core_vector_stores.py reverted to main
exit_WITHOUT_fix=1
FAILED test_no_count_diagnostics_run_when_debug_is_off

# with the guard
exit=0   (3 passed)
```

**Two controls against the guard going wrong**, since the danger of a change like this is silently *removing* a diagnostic:
- `test_diagnostics_still_run_when_debug_is_on` — deferred, not deleted
- `test_results_are_identical_either_way` — the guard changes timing and nothing else

All three assert the search returned rows **first**, because all three would otherwise pass vacuously against an empty result set — which is exactly how the fixture was broken while writing them (a made-up `embedder_path` stored the vectors in a column the store never read back).

Regression run over `test_vector_store_diagnostics`, `test_agent_system_prompt_assembly`, `test_corpus_isolation_vector_store`, `test_hybrid_search`, `test_unified_agent_consumer_delegation`: 81 passed, exit 0.

Note for anyone running the suite locally: under `local.yml` pytest dies at `django.setup()` with `Application labels aren't unique, duplicates: storages`, because `settings/test.py` appends `storages` unconditionally while `base.py` also appends it when `STORAGE_BACKEND=AWS`. `STORAGE_BACKEND=LOCAL` works around it. Left alone here as out of scope, but it is a latent trap.

PR #2262, opened 2026-08-18
feat(agents): system_instructions_mode — let an agent config EXTEND the corpus persona instead of replacing it

JSv4:
> Stacked on #2261 (base is `perf/vector-store-debug-counts`, so the diff here is just this change).

## The asymmetry

In `unified_agent_conversation.py`, a selected agent's `system_instructions` **replace** the context-derived prompt, while its `available_tools` **merge**:

```python
if self.agent_config.system_instructions:
    agent_kwargs["system_prompt"] = self.agent_config.system_instructions
config_tools = list(self.agent_config.available_tools or [])
```

That forces a bad trade on any corpus-scoped agent: **attaching a tool costs you the corpus persona.** You get the persona that knows the corpus, or the agent that can call `search_across_corpora` — not both.

The only workaround is to paste the entire persona into `system_instructions`, which duplicates it into a second row that drifts from `Corpus.corpus_agent_instructions` silently.

## Measured cost of the workaround

On a 4,679-section authority deployment (21 corpora, 18-member group), scoring answers on whether their quotations exist verbatim in the installed corpus:

| | persona only | orchestrator only | **persona + reach** |
|---|---|---|---|
| main-set delta | 8/15 | 6/15 | **10/15** |
| expansion delta | 2/7 | 5/7 | **6/7** |

The combination is worth roughly +2 and +4 over either alone, and it is the only configuration that reaches case law and Federal Register preambles. There is no way to express it today except duplication.

## The change

Adds `AgentConfiguration.system_instructions_mode`:

- `REPLACE` — the historical behaviour, and the **default**, so no existing configuration changes.
- `EXTEND` — the instructions are appended to the context-derived prompt.

EXTEND rides a new `extra_system_context` kwarg, which the factory drains through `AgentConfig.resolve_system_prompt` **after** persona resolution. That is the mechanism added for #2247 for exactly this "append without consuming the persona signal" case — assigning to `system_prompt` here instead would consume the "caller supplied no prompt" signal and discard the persona, which *is* #2247.

Wired into **both** the document and corpus factories, not only the corpus one I needed.

## Verification

`test/verify_extend_mode.py` asserts the composition on the **built agent**, not on the intention — #2247 was silent, so intention is not evidence:

```
config us-export-control-reach-extend: mode=EXTEND instructions=4,212 chars
corpus persona: 14,753 chars
built system prompt: 20,284 chars
  contains corpus persona:      True
  contains config instructions: True
  config tools: ['search_across_corpora']
OK — persona preserved AND config instructions appended
```

With EXTEND the config stores 4,212 chars (only what it adds) instead of 18,721 (persona + additions), and the persona stays single-sourced on the corpus.

## Caveat

The repo suite was not run here: under `local.yml`, pytest dies at `django.setup()` on a duplicate `storages` app label before any of this is imported, and standing up `test.yml` risked evicting a database another session was using mid-measurement. Migration `0018` is additive with a default, so it is backward-compatible, but CI should exercise it.


JSv4:
Verified **both** branches, not just the new one — REPLACE is the default every existing configuration relies on, so it is the more important of the two to prove unchanged:

```
config us-export-control-reach-extend: mode=EXTEND instructions=4,212 chars
corpus persona: 14,753 chars

REPLACE: prompt  4,868 chars  persona=False config=True    <- unchanged
EXTEND : prompt 20,284 chars  persona=True  config=True
  config tools: ['search_across_corpora']
```

REPLACE excludes the persona and applies the config text (the historical behaviour); EXTEND carries both with the persona first; the merged tool survives either way. Harness is `test/verify_extend_mode.py` in the ITAR workspace — it asserts on the **built agent's** `config.system_prompt`, because #2247 was silent and 'the code looks right' is not evidence on this seam.

JSv4:
Added tests to `opencontractserver/tests/test_agent_system_prompt_assembly.py` — the #2247 regression file, which is the same seam this change touches.

**`ExtraSystemContextFactoryTestCase` is the regression coverage.** It drives the real factory (with `PydanticAIAgent` patched at the seam the rest of that file uses, so no LLM is built) using `extra_system_context=` — the kwarg the consumer actually passes — and asserts on the assembled `config.system_prompt`. Verified both cases fail without the change, by reverting `agent_factory.py` to its pre-feature state:

```
# agent_factory.py reverted (0 references to extra_system_context)
exit_WITHOUT_fix=1
FAILED test_corpus_factory_appends_extra_system_context
FAILED test_document_factory_appends_extra_system_context

# with the change
exit=0
```

Covers the **document** factory as well as the corpus one, since an agent config can be selected for a document too — and that test caught a real gap in my own fixture rather than passing by luck.

`test_omitting_it_changes_nothing` is the control: absent the kwarg, the prompt is unchanged.

**`ExtendModeAssemblyTestCase` is contract coverage, not regression coverage**, and is labelled that way in the file: it exercises `add_computed_context` ordering, which predates EXTEND, so those tests would pass without this PR. They are there to pin the ordering guarantee EXTEND depends on (persona first, appended blocks after) so a future change to `resolve_system_prompt` can't silently invert it.

Combined run with the diagnostics tests from #2261: 23 passed, exit 0. Broader run across five related files: 81 passed.

PR #2277, opened 2026-08-24
fix: use identity checks for type comparisons; remove mutable default argument

harshadkhetpal:
## Summary

1. **`opencontractserver/llms/tools/tool_factory.py`** — the annotation→JSON-schema dispatch chain compares type objects with `==` five times (`param.annotation == int`, `float`, `bool`, `list`, `dict`) — ruff E721. Identity (`is`) is the intended semantics for type objects and avoids surprises with metaclass-defined `__eq__`.

2. **`opencontractserver/pipeline/utils.py`** — `run_post_processors(..., input_kwargs: dict[str, Any] = {})` shares one dict across every call (ruff B006). Changed to the `None` sentinel with `input_kwargs = input_kwargs or {}` after the docstring; identical behavior for all callers.

## Testing
`python -m py_compile` passes on both files; `ruff check --select B006,E721` on them goes clean.

### DIFF
--- a/opencontractserver/llms/tools/tool_factory.py
+++ b/opencontractserver/llms/tools/tool_factory.py
@@ -175,15 +175,15 @@ class CoreTool:
 
             # Try to infer type from annotation
             if param.annotation != inspect.Parameter.empty:
-                if param.annotation == int:
+                if param.annotation is int:
                     param_info["type"] = "integer"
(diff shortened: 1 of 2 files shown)


JSv4:
@harshadkhetpal can you please acknowledge the CLA?

PR #2344, opened 2026-09-14
Apply profile READ and agent context checks

JSv4:
Apply the existing access rules to profile transfers and agent context:

- Filter profile import/export connections through each model's READ rules, using one helper with stable primary-key pagination. Imports retain creator/public visibility; exports also support Guardian grants.
- Bind moderation tools to the factory actor and limit approval/rejection to messages in the agent's conversation.
- Make BaseService.filter_visible_qs check the normalized queryset so unsupported manager shapes fail with the documented TypeError.

Validation: 109 tests and 41 subtests passed across profile, approval, BaseService, and tool coverage. TypeScript and pinned hooks, including full-project mypy, passed. Existing test assertions are retained.


JSv4:
Follow-up commit addresses both review passes. Item by item:

**1. Ephemeral / delegated sessions and `resume_with_approval`** — confirmed and documented as a pre-existing dead end, not a regression. Every `requires_approval=True` tool in `tool_registry.py` also sets `requires_write_permission=True`, and write tools are filtered out for anonymous users, so anonymous chat never reaches the gate. A `persist=False` sub-agent acting for a writer can reach it, but its placeholder ids are synthetic in-memory counters (`CoreConversationManager.create_placeholder_message`), so the old unscoped `aget(id=...)` never resumed the right row either — it matched whatever `ChatMessage` happened to hold that small pk. The scoped lookup now fails closed with an explicit `ValueError("... has no persisted conversation ...")` instead of a bare "not found", and the tradeoff is written up in `docs/architecture/llms/README.md` (Tool Approval) and the changelog fragment. Making ephemeral sub-agents resumable (resolving the paused message from the manager's buffer) is a separate feature and is called out as such.

**2. DRY / `filter_visible_qs`** — the four sites now go through one helper, `config/graphql/user_types.py::_filter_visible_transfers`. `filter_visible_qs` cannot be used for these two models, though: `UserExport` / `UserImport` keep `BaseOCModel`'s bare `BaseVisibilityManager`, which is a plain `Manager` with no `visible_to_user`-aware `QuerySet`. The related manager itself has `visible_to_user`, so the helper's `hasattr` guard passed, but `.all()` hands back a plain `QuerySet` and the chained call raised `AttributeError`. `_get_queryset_UserFeedbackType` works because `UserFeedback` has its own `UserFeedbackQuerySet`. Two fixes on the helper side: `filter_visible_qs` now runs its fail-closed guard on the normalised queryset (so this shape raises the documented `TypeError`), and its docstring plus `docs/architecture/query_permission_patterns.md` no longer claim every model exposes `visible_to_user` on the queryset. Regression: `test_base_service.py::test_filter_visible_qs_rejects_bare_visibility_manager_related_manager`. Agreed `request=info.context` is a no-op on the list helpers; kept for the documented API parity.

**3. `moderator_id` in a literal tuple** — replaced with `TOOL_ACTOR_IDENTITY_PARAMS` in `opencontractserver/constants/tools.py`, documented as the convention (a tool declares its actor by naming the param from this set; unlisted names are LLM-visible). `moderation_tools.py`'s module docstring now says the id is factory-bound rather than caller-supplied. Pinned by `test_pydantic_ai_tools_module.py::test_every_actor_identity_param_is_bound`, which also checks that a name outside the set (`reviewer_id`) is *not* injected.

**4. `UserImport` vs `UserExport`** — kept as the existing model capability, called out explicitly rather than papered over: nothing in the codebase grants object permissions on imports or exports (no `set_permissions_for_obj_to_user` / `assign_perm` call for either outside tests), and `core_permissions` already degrades gracefully when the `*userobjectpermission_set` accessor is absent. Adding guardian tables plus a migration for `UserImport` would be a new sharing capability, not a fix. The asymmetry is now stated in the profile-privacy note and in two new rows of the reference table in `consolidated_permissioning_guide.md`, in the changelog fragment, and in the test comment.

**Nit (duplicate message)** — the two branches now carry different messages; the no-conversation branch explains itself instead of masquerading as a lookup miss. The stale `# pragma: no cover` on the `ObjectDoesNotExist` branch is gone since the new test covers it.

Verification: pre-commit (black/isort/flake8/mypy/changelog fragments) clean, `manage.py check` clean (E001), and the targeted suites — `test_base_service`, `test_pydantic_ai_tools_module`, `test_user_profile`, `test_agent_memory`, `architecture/test_graphql_service_layer`, `test_schema_parity`, `test_tool_approval_gate`, `test_nested_approval_gates`, `test_websocket_document_approval` — pass locally.

JSv4:
Addressed the moderate pagination finding in f188d835d. The shared profile-transfer filter now explicitly orders by primary key before cursor slicing. The added API test pages through all four transfer connections; the existing READ and profile-scope assertions remain unchanged.

Validation: 109 tests and 41 subtests passed across profile, approval, BaseService, and tool coverage. TypeScript and pinned hooks, including full-project mypy, passed. The earlier findings remain covered by the existing follow-up commits and documentation; minor/non-blocking suggestions are unchanged.

PR #2348, opened 2026-09-14
Recheck requester scope before queued research runs

JSv4:
Research kickoff and queued/resumed execution share the active-account, corpus READ, and optional corpus-group checks. Workers reload the stored requester and scope before entering the research loop, and refuse a job when those grants have been revoked. An inactive requester is also rejected before a new report is created or queued.

Validation: all 178 research tests and 53 subtests passed, including kickoff denial and the queued/resumed revocation matrix. TypeScript and pinned hooks, including full-project mypy, passed. Existing test assertions are unchanged.


JSv4:
Addressed the kickoff/worker mismatch in 34583e3bb. The existing active-account guard now lives in ResearchReportService.require_scope, alongside corpus READ and optional group visibility. Both start() and queued/resumed execution use it; the worker-only duplicate is removed.

The new regression demonstrates that the prior service allowed an inactive requester to start research on a public corpus. It now verifies rejection before creating a report or enqueuing work. All 178 research tests and 53 subtests passed, along with TypeScript and pinned hooks including full-project mypy. The non-blocking observations remain unchanged.

PR #2349, opened 2026-09-14
Check access to selected agent tool targets

JSv4:
Check selected agent tool targets using existing model READ rules. Content mutations require current CRUD on selected document/corpus/note/annotation targets, and thread posting requires conversation READ. File management and the five moderation actions retain their existing service-owned write rules. `ask_document` retains its corpus/group-owned selection; navigation tools retain their service-owned READ checks and structured errors.

Bind positional, default, and forwarded arguments before validation and approval. Malformed arguments return normal tool errors; permission and approval exceptions still propagate. Reuse the invocation's actor lookup.

Validation: **165 tests and 86 subtests passed**, including READ-only target denial, successful writes, revocation, thread posting, corpus-agent delegation, bound document-summary access, navigation errors, file-management permissions, malformed inputs, approval handling, and moderation. The new review regression cases fail on the previous implementation. TypeScript and pinned hooks, including full-project mypy, passed.


JSv4:
Addressed findings 1, 2, and 4:

- Selected note/annotation mutations now require the target model's existing CRUD permission before approval or execution. Regression cases cover READ-only parents, successful writes, and revocation.
- Thread posting now checks current conversation READ. Only the five moderation actions retain their existing role-based gates.
- Argument binding and target checks run inside normal operational error handling; PermissionError and ToolConfirmationRequired still propagate. Invalid calls no longer fall back to partial validation.

The new tests reproduce all three findings on the previous implementation. The informational link-tool observation and explicitly non-blocking notes are unchanged.

Validation: 132 tests and 82 subtests passed; TypeScript and pinned hooks (including full-project mypy) passed.

JSv4:
Addressed both findings from the latest review.

- `ask_document` retains its existing corpus/group-owned document selection. A new regression uses corpus READ without document READ and verifies delegation, rejection outside the corpus, and denial after corpus access is revoked.
- Selected document/corpus content mutations now require current CRUD, as note/annotation mutations already do. The new document-write case covers denial before approval, a successful authorized write, and revocation. Bound IDs are skipped only when the corresponding permission was already checked.

File move/rename/delete retain their existing corpus service UPDATE/DELETE rules; the fix does not add document CRUD to those operations. A wrapped rename regression pins that behavior and still checks revocation. The five moderation actions retain their role checks.

Validation: 135 tests and 82 subtests passed, plus TypeScript and pinned hooks including full-project mypy. Both additional regression cases fail before this change.

JSv4:
Resolved item 1 by restoring the existing bound-context rule. A document-bound tool retains document CRUD plus corpus READ; the selected-target check skips those bound IDs after `_check_user_permissions` validates them. Separately selected targets still receive the added READ/CRUD checks.

The new summary regression passes against the pre-PR implementation, fails with the extra corpus CRUD gate, and covers the successful revision and denial after corpus READ is revoked.

On item 2, preserving `ask_document`'s existing corpus/group selection is deliberate. This PR keeps that established access contract; a change to that policy is separate from these target-check fixes. The remaining minor observations are unchanged.

Validation: 136 tests and 82 subtests passed, plus TypeScript and all pinned hooks including full-project mypy.

JSv4:
Addressed items 1 and 2. `get_document_references` and `find_documents_citing` keep their service-owned READ checks and structured error results for missing or inaccessible documents. The wrapper regression passes on the pre-PR implementation and reproduces all four changed-error cases before this fix.

Annotation write targets now preload `created_by_analysis` and `created_by_extract`, avoiding those per-row FK fetches during the existing permission check. Notes retain their valid document/corpus prefetch. The non-blocking registry-metadata suggestion is unchanged.

Validation: 165 tests and 86 subtests passed, including graph navigation and the affected tool/approval/moderation suites. TypeScript and pinned hooks, including full-project mypy, passed.

PR #2355, opened 2026-09-14
Expire permission snapshots after membership changes

JSv4:
A held object's cached group grants can survive removal of the user's membership. GraphQL permission metadata can likewise retain changed model grants or reuse another actor's metadata on the same request.

Use transaction snapshots to track membership and model-permission edits, including Group/Permission cascade deletion through instance and bulk APIs. GraphQL metadata is keyed by actor and model; copied or serialized users drop cached backend permissions. Raw Guardian and through-table writes retain explicit invalidation duties. Direct Django has_perm/get_all_permissions calls keep Django's per-instance cache contract; the guide documents using a fresh user or the snapshot-aware helper after permission edits.

Validation: 397 tests and 28 subtests passed, with one existing skip, covering permissions, groups, user privacy, admin authentication, GraphQL, and query-count checks. TypeScript and pinned hooks, including full-project mypy, passed. The original regressions reproduce ten failures on the preceding implementation; the deletion regressions reproduce six additional stale-grant cases before the review fix. Existing test assertions are unchanged.


JSv4:
Added Group/Permission post-delete invalidation, including bulk and admin cascade deletion. Direct-only grant snapshots now also observe the database-wide revision, since deleting a Permission can remove direct Guardian rows as well as model/group grants. The new regressions fail in six cases before the fix and cover held instances, prefetched document grants, both cache tiers, GraphQL model metadata, and rollback.

The direct Django calls in point 2 retain their pre-existing per-instance cache contract. I kept model_permission_grants computing on an isolated copy: mutating the source User would share connection-specific Django cache state and still would not expire other held User objects. The guide now explicitly directs callers to a fresh User or model_permission_grants after permission edits. The cited admin view checks the current request user, and the agent tool wrapper reloads the actor for each invocation. This PR does not claim to make every direct Django permission API snapshot-aware.

The database-wide invalidation is an intentional precision/query-cost tradeoff and now covers deletion too. I also removed the guide's obsolete claim that membership changes require manual invalidation.

Validation: 397 tests, 28 subtests, one existing skip; TypeScript and pinned hooks including full-project mypy passed. Existing assertions are unchanged.

